PR #55, opened 2016-11-17
Fix: Used parallel processing concept to upload samples by mzroll file

Giridhari013:
-Only one processor work at a time while uploading samples by mzroll file
-It is taking lot of time in uploading samples


sahil21:
@Giridhari013 This code sometimes crash in Windows. Ubuntu it is fine. I have undo the changes from the develop branch.

PR #62, opened 2016-11-25
Ftr:To select samples as blank samples

Giridhari013:
-Add a button in Samples Widget to select samples as blank samples
-After selecting some or whole samples and then click on that button
 can select those samples as blank samples


chubukov:
Can we please make this a toggle instead of just a set (kind of like how check/uncheck works)? It would be annoying to set the wrong one and not have a way to go back without reloading the files.

PR #189, opened 2017-03-24
Fix: A option for adding isotopes in bookmark & barplot in options.

rkdahmiwal:
- A different checkbox option is added in isotope detection tab
     of options dialog.

   - This option when checked enables the addition of isotopes in
     bookmark table and isotopic barplot.


mzroll/settingsform.cpp
```diff
@@ -72,9 +74,15 @@ void SettingsForm::setSettingsIonizationMode(QString ionMode) {
     else                                    ionizationMode->setCurrentIndex(0);
 }
 
+void SettingsForm::showSetIsotopeAtom() {
+    setIsotopeAtom();
+    cerr << mainwindow->mavenParameters->isotopeAtom["IsotopeWidget"] << " hello \n\n";
```
sahil21: Remove this print statement

PR #198, opened 2017-04-05
fixed int divide error in quantile calculation

chubukov:


sp-eldata:
@sahil21: Why does this PR have unmerged commits? Are we going to merge them? Just checking.

sahil21:
@sp-eldata Yes, we will merge them.

PR #201, opened 2017-04-08
Fix baseline

chubukov:


chubukov:
Pulling the full EIC and doing peak finding on the whole range really does create a major performance problem. Probably need to limit the peak finding to the right RT range as originally suggested.

libmaven/EIC.cpp
```diff
@@ -503,9 +519,12 @@ void EIC::getPeakStatistics() {
         findPeakBounds(peaks[i]);
         getPeakDetails(peaks[i]);
 
-        if (peaks[i].peakAreaTop > maxAreaTopIntensity)
+        if (peaks[i].peakAreaTopCorrected > maxAreaTopIntensity)
             maxAreaTopIntensity = peaks[i].peakAreaTop;
```
sahil21: maxAreaTopIntensity = peaks[i].**peakAreaTopCorrected**;

PR #224, opened 2017-04-28
Ref peakdetector cli

sahil21:


chubukov:
@sahil21 Great. I'm noticing that the number of options is sort of spiraling out of control. Would it be better to instead have a configuration file (json or xml format) that gets passed in?

sahil21:
@chubukov Yes, I was also thinking about introducing a configuration file before I add more arguments. XML is better in this case. Readability is better in XML.

Issue #136, opened 2017-02-09
feature: export groups from scatterplot


chubukov:
Something Gina tried to do the other day, and I was surprised it was not possible.

- Do peak picking (targeted or untargeted, but makes more sense on untargeted)
- Assign samples to sets 
- Make a scatter plot or volcano plot from your peak table, and filter groups based on log fold change or p-value
- Export just the remaining (filtered) groups.

I think best would be if you could copy the currently filtered groups to a new peak table, then you could export them to CSV from there?

Not particularly high priority, but if it's easy enough, it would be a nice feature. In terms of philosophy: I don't necessarily want to put a lot of effort into making Maven into a super intricate statistical analysis package, but I think the rudimentary features it has are very useful and should continue to be supported.

shubhra-agrawal:
@chubukov this feature has been merged to develop. Let us know if this is what you had in mind. We will also add a "How-to" in the Wiki section for other users.

chubukov:
@shubhra-agrawal holy cow this was from over a year ago, i'd basically forgotten about it. Anyway, it's a nice feature, thanks. I'll take a look.

shubhra-agrawal:
@chubukov Closing this issue. [Here's a link](https://github.com/ElucidataInc/ElMaven/wiki/Introduction-to-ElMaven-UI#statistics-module) to the Statistics module intro in the Wiki.

PR #225, opened 2017-04-28
Ftr: Peak table is added for visualizing groups present in scatterplot.

rkdahmiwal:
- A peak table button is added in toolbar of scatterplot to hide
     and show the peak table in UI.

   - This peak table will contain all the groups that are present in
     scatterplot graph.

   - This peak table is directly attached and get refreshed whenever
     two or more sample sets are compared with some log2 fold change
     cutoff, p-value cutoff etc.

   - This peak table works same as other peak tables.

### DIFF
--- a/src/gui/mzroll/comparesamplesdialog.cpp
+++ b/src/gui/mzroll/comparesamplesdialog.cpp
@@ -147,6 +147,7 @@ void CompareSamplesDialog::compareSamples() {
 	if (sset1.size() == 0 || sset2.size() == 0)
 		return;
 	compareSets(sset1, sset2);
+	if (parentWidget()) ((ScatterPlot*) parentWidget())->showPeakTable();
 }
 
 void CompareSamplesDialog::compareSets(vector<mzSample*> sset1,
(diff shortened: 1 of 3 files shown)


chubukov:
So is the peak table supposed to show the fold change and p-value associated with each group? If so, I think it's a great feature. If not, then I'm not sure I follow -- how will it be different from the peak table that was used to generate the scatter plot in the first place?

rkdahmiwal:
This pull request is for github issue #136.

1) Main idea of implementing this peak table is get all the information about the groups that are currently present in scatterplot. Also it will show fold change and p-value associated with each group.

2) If user filter groups in scatterplot on basis of min2 log fold diff, p value or any other parameter present compareSampleDialog. Then this peak table will contain only those filtered groups and user can export the required groups.

chubukov:
@rkdahmiwal Ah, now I understand, sorry. Yes, that definitely has utility.

If there's a way to add the log2fc and pvalue columns, it could also be super useful. Can be separate issue (not high priority) if non-trivial.

suresh-surampudi:
Hi Victor!

Should we add as a feature (not high priority) to github?

"If there's a way to add the log2fc and pvalue columns, it could also be
super useful."

Regards
vs
=====================
Venkata suresh kumar, PhD
Project Manager, Elucidata

PR #227, opened 2017-05-01
Fix: Expanding of Toolbar in TableDockWidget

sahil21:
- Implementation of class QWidgetAction for custom widgets in the QToolBar
- Added a new class TableToolBarWidgetAction for implementing QWidgetAction
- Merge Table Button is not included in this Class because of Dynamic Drop Menu

### DIFF
--- a/mzroll/tabledockwidget.cpp
+++ b/mzroll/tabledockwidget.cpp
@@ -37,93 +37,13 @@ TableDockWidget::TableDockWidget(MainWindow* mw, QString title, int numColms, in
     connect(clusterDialog->clusterButton,SIGNAL(clicked(bool)),SLOT(clusterGroups()));
     connect(clusterDialog->clearButton,SIGNAL(clicked(bool)),SLOT(clearClusters()));
 
-    QToolBar *toolBar = new QToolBar(this);
-    toolBar->setFloatable(false);
-    toolBar->setMovable(false);
-
-    QToolButton *btnSwitchView = new QToolButton(toolBar);
-    btnSwitchView->setIcon(QIcon(rsrcPath + "/flip.png"));
-    btnSwitchView->setToolTip("Switch between Group and Peak Views");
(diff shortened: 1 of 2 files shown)


mzroll/tabledockwidget.cpp
```diff
@@ -2389,3 +2282,155 @@ MatrixXf TableDockWidget::getGroupMatrix(vector<mzSample*>& samples, PeakGroup::
     return X;
 }
 
+QWidget* TableToolBarWidgetAction::createWidget(QWidget *parent) {
```
sahil21: @GeorgeSabu This is a virtual function of class QWidgetAction. Get's called when an instance of this class or class inheriting this class is made.

sahil21: @GeorgeSabu Done. It's in the tableDockWidget.h file.

mzroll/tabledockwidget.h
```diff
@@ -183,4 +183,20 @@ protected Q_SLOTS:
     
 };
 
+class TableToolBarWidgetAction : public QWidgetAction
+{
+    public:
+        QString btnName;
```
sahil21: @GeorgeSabu Makes sense. I made it private.

mzroll/tabledockwidget.cpp
```diff
@@ -2389,3 +2282,155 @@ MatrixXf TableDockWidget::getGroupMatrix(vector<mzSample*>& samples, PeakGroup::
...
+
+
+    if (btnName == "titlePeakTable") {
+ 
+        QLabel *titlePeakTable = new QLabel(parent);
```
rkdahmiwal: Declaring titlePeakTable here can create conflicts because earlier it was declared in the TableDockWidget's constructor and used for title of peak table present in scatterplot.

sahil21: @rkdahmiwal  We'll resolve it when conflict arises

PR #231, opened 2017-05-05
Ftr: Add formula column to peaks and groups format

sahil21:


chubukov:
I'm going to need to make an Agios specific branch of this, because our legacy workflow relies on exactly 13 columns of metadata in the Maven output. If you want to make a universal version, I would get rid of a column that is not very useful, such as goodpeakcount.

shekhjha:
@chubukov Does the order of the 13 Column matter as well? If not, you can save an "Agios Template" and use that to export. No?

sahil21:
@shekhjha We haven't implemented that feature of saving template yet. It is not part of this sprint. Also, just to point out this branch is not merged with develop yet because there is an issue with automatic peak picking without DB. We are fixing that and soon merged it to develop.

PR #238, opened 2017-05-09
Ftr xml parser for cli

sahil21:


chubukov:
Just thinking here: you were also talking about implementing a feature where when you do automatic peak picking from the GUI, all the parameters are saved. Does it make sense to make the correspondence between the way that those are saved and the input to the command line version more explicit? So that if you don't do any manual steps in the GUI, there is a file that you can pass to the command line and get the same output. It might not be perfect correspondence, but I think all the information should be there ( unless user bookmarks peaks or merges tables, etc etc).

chubukov:
Just to reiterate, I don't think what I said above is a critical feature, just thinking about consistency.

sahil21:
@chubukov We are writing XML parser code in such a way that the parameters from the XML file will be loaded into mavenParameters class. mavenParameters class is shared by both CLI and GUI. Loading XML in the GUI won't be that big a task once this code is modularized. Also, we can write a script to save mavenParameters members variables into an XML file.

PR #248, opened 2017-05-23
Fix: Up/Down keys update the treewidget items.

rkdahmiwal:
- Up/Down keys update the compound of compound widget and and
     EIC window of tabledockwidget.

   - Mouse click is also enabled on tree widget items.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,72 @@
-El-Maven
-=====
+<h3 align="center">Elmaven</h3>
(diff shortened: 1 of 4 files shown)


sahil21:
@rkdahmiwal I'm merging this pull request right now because it will decrease performance only when we click using a mouse. Later we'll solve this issue of two different signals calling the same slot.

PR #253, opened 2017-06-07
Ftr: Load CSV file with sample sets

shubhra-agrawal:
- changed default set name to 'NA'
- void loadSetsFile(): load CSV file
- bool loadSetsFile(filename): returns true if file successfully loaded
- loadSetsCSVFile: Reads file


mzroll/projectdockwidget.cpp
```diff
@@ -48,6 +49,12 @@ ProjectDockWidget::ProjectDockWidget(QMainWindow *parent):
     //Trigger to open() in slot to load samples while uploading .mzroll file --@Giridhari
     connect(loadButton,SIGNAL(clicked()),_mainwindow, SLOT(open()));
 
+    QToolButton* loadSetsButton = new QToolButton(toolBar);
+    loadSetsButton->setIcon(QIcon(rsrcPath + "/fileopen.png"));
```
sahil21: @shubhra-agrawal Icon is same as loading samples/project file. Please use some different icon.

mzroll/mainwindow.cpp
```diff
@@ -1639,6 +1656,119 @@ void MainWindow::loadCompoundsFile() {
...
+    if ( filelist.size() == 0 || filelist[0].isEmpty() ) return;
+	if(!loadSetsFile(filelist[0])) {
+		string dbfilename = filelist[0].toStdString();
+		string dbname = mzUtils::cleanFilename(dbfilename);
+		string notFoundColumns = "Following are the unknown column name(s) found: ";
+
+		QMessageBoxResize msgBox;
+		msgBox.setText(tr("Trouble in loading set information %1").arg(QString::fromStdString(dbname)));
+		msgBox.setIcon(QMessageBoxResize::Warning);
+		if (DB.notFoundColumns.size() > 0) {
```
sahil21: DB is for the compound database. You can't use this condition here for loading sets.

mzroll/mainwindow.cpp
```diff
@@ -1639,6 +1656,98 @@ void MainWindow::loadCompoundsFile() {
...
+        }
+
+        string set;
+		QString sampleName;
+        int N=fields.size();
+
+        if ( header.count("sample")&& header["sample"]<N) 	 sampleName = QString::fromUtf8(fields[ header["sample"] ].c_str());
+        if ( header.count("set")&& header["set"]<N)	set = fields[ header["set"] ];
+		else{
+			set = "NA";
```
sahil21: This creates a new set called NA. The column should be empty. The default should be empty or there should be some condition to check if it is NA then don't process it as setname.

PR #258, opened 2017-06-16
Fix: compound charge is used for mass calc unless user enters formula

shubhra-agrawal:
- getCharge returns the appropriate charge for mass calculation (compound charge used if it exists, user input charge used otherwise)
- formulaFlag to track if user searches by formula

### DIFF
--- a/mzroll/eicwidget.cpp
+++ b/mzroll/eicwidget.cpp
@@ -1487,7 +1487,7 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
 
 	if (group == NULL)
 		return;
-	int charge = getMainWindow()->mavenParameters->getCharge();
+	int charge = getMainWindow()->mavenParameters->getCharge(group->compound);
 	if (group->getExpectedMz(charge, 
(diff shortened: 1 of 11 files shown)


mzroll/eicwidget.cpp
```diff
@@ -1486,10 +1487,10 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
 
 	if (group == NULL)
 		return;
-
-	if (group->getExpectedMz(getMainWindow()->getIonizationMode()*getMainWindow()->mavenParameters->charge, 
+	int charge = getMainWindow()->mavenParameters->getCharge();
```
sahil21: No Group added here.

mzroll/eicwidget.cpp
```diff
@@ -1518,10 +1519,11 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
 		eicParameters->_slice.rtmin = bounds.rtmin;
 	if (eicParameters->_slice.rtmax > bounds.rtmax)
 		eicParameters->_slice.rtmax = bounds.rtmax;
+	charge = getMainWindow()->mavenParameters->getCharge();
```
sahil21: No Group added here.

Issue #261, opened 2017-06-16
export to csv: deleted groups and unchecked samples


chubukov:
Possibly related to fixing #196 

- if you delete a child (isotope) group, it still gets exported when you export to CSV, though all the intensities are zero
- if a sample is unchecked, it still gets exported to CSV, with all zero values. (i mean unchecked when the peakgroup was generated, i haven't checked the behavior of unchecking after).

I think both of these are pretty misleading.

rish9511:
@chubukov 

I wanted to understand when should the sample be exported to csv?

* should it be exported if it was used to generate peak groups?

* or should it be exported only if the sample is checked irrespective of the fact whether it was used to generate peak group or not

chubukov:
The only behavior that I 100% want to avoid is that a group gets exported with intensity zero for particular sample when in fact that peak does not have zero intensity in that sample.

You're right that this all leads to some stupid corner cases -- check, do peak detection, uncheck in done random order will make

chubukov:
Don't know why I can't edit comment on phone, continuing above:

You're right that this all leads to some stupid corner cases -- check, do peak detection, uncheck or the reverse will lead to some messy results. See it you can come up with something reasonably consistent, and if not, I will think about it more and comment here. I think the option of using check/uncheck for display only is a reasonable option.

sahil21:
@chubukov @rish9511 I think we should pass "NA" in case a sample is unchecked while peak detection/grouping. Right now the user is not able to understand the difference between zero intensity of a sample or unchecked sample (they both are zero in the output CSV/TAB file). 

This will also take care of the corner case when different groups have different samples in same peaks table. This corner case can happen when the user is bookmarking peaks manually and checking/unchecking samples or in the case of merging of peaks tables.

chubukov:
@sahil21 @rish9511 That sounds fine to me

rish9511:
@chubukov 
There were two sub issues  in this:
- Unchecked samples should not not be exported to csv
To solve this we keep track of the samples that were used while peak detection. 
While exporting we check if the sample was used or not. If it was used we give the correct intensity, otherwise we display NA


- Deleted isotopes should not be exported to csv
The problem with this was, the isotope was not deleted from the actual list that holds the isotopes but was only  deleted from the treewidget. Hence on exporting, all the isotopes were exported
[...]

PR #306, opened 2017-08-01
keep track of the samples used while peak detection

rish9511:
For each group that gets formed during the peak detection, keep track of the samples that were used for it

When exporting the groups to csv, intensities of the samples should be displayed only if they were used while peak detection

Issue: #261

### DIFF
--- a/libmaven/PeakGroup.cpp
+++ b/libmaven/PeakGroup.cpp
@@ -237,7 +237,13 @@ vector<float> PeakGroup::getOrderedIntensityVector(vector<mzSample*>& samples, Q
 
     for( unsigned int j=0; j < samples.size(); j++) {
         sampleOrder[samples[j]]=j;
-        maxIntensity[j]=0;
+        // check if the the sample was used while peak detection or not
+        // samplesUsed is populate during peakDetection
+        auto it = std::find(std::begin(samplesUsed), std::end(samplesUsed), samples[j]);
+        if(it != std::end(samplesUsed))
+            maxIntensity[j]=0;
+        else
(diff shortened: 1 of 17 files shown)


chubukov:
@rish9511 Great idea. Have you considered the potential pitfalls if samples get deleted between peak detection and recall?

libmaven/PeakGroup.cpp
```diff
@@ -237,7 +237,13 @@ vector<float> PeakGroup::getOrderedIntensityVector(vector<mzSample*>& samples, Q
 
     for( unsigned int j=0; j < samples.size(); j++) {
         sampleOrder[samples[j]]=j;
-        maxIntensity[j]=0;
+        // check if the the sample was used while peak detection or not
+        // samplesUsed is populate during peakDetection
+        auto it = std::find(std::begin(samplesUsed), std::end(samplesUsed), samples[j]);
+        if(it != std::end(samplesUsed))
```
shubhra-agrawal: We're setting maxIntensity[j] = -1 here when line 270 is doing the same thing later after normalization. Any particular reason for it?

rish9511: It's just a sanity check but not sure whether it's needed

rish9511:
@chubukov 
Thanks for pointing out. We have taken care of it now. The vector that stores the used samples is made to remove the sample being deleted

Another thing I wanted to ask was, should the sample be exported to csv if it's deleted ? It's not exported as of now

chubukov:
@rish9511 I would say no. If it's deleted, it should be safe to remove all associations with it from all the peakgroups.

PR #308, opened 2017-08-03
Alignment

Raghavdata:


bin/alignment.py
```diff
@@ -0,0 +1,118 @@
+import pandas as pd
+import numpy as np
+from scipy.interpolate import interp1d
+from sys import stdin
+import sys
```
ramb0111: according to python imports convention , the imports must be in order of first python built-in libraries , then 3rd party modules , at last project related modules

bin/alignment.py
```diff
@@ -0,0 +1,118 @@
...
+import json
+import statsmodels.api as sm
+import warnings
+
+
+input_json = ''
```
ramb0111: global variables should be avoided , why you are using it?

bin/alignment.py
```diff
@@ -0,0 +1,118 @@
...
+
+def group_json_to_df(g, group_num, groups_data):
```
ramb0111: No documentation for a any function

bin/alignment.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def group_json_to_df(g, group_num, groups_data):
+    """This function converts the input group json into a pandas dataframe"""
+    return pd.DataFrame({'group': g,'sample': groups_data[g][group_num].keys(), 'rt': groups_data[g][group_num][groups_data[g][group_num].keys()[0]]})
+
+
+
+def groups_json_to_df(g, groups_data, minSample, extraPeaks):
+    """This function converts the input groups json into a pandas dataframe and subsets the data"""
```
sahil21: Put proper doc strings. Explain input arguments and returns.

Issue #356, opened 2017-09-01
PPM window in CLI version


shubhra-agrawal:
Allow user to set the PPM window for compound mass in CLI

shubhra-agrawal:
This has been implemented and merged to develop.

chubukov:
@shubhra-agrawal sorry, this issue slipped by me. As commented on #357 I would look carefully at this code, because in practice, I have been using this feature through the -p parameter.

shubhra-agrawal:
@chubukov Adjusting the value of the '-p' parameter affects the final number of groups in the exports, but it is not the same as adjusting the ppm. 

'-C' or 'CompoundPPMWindow' is the CLI equivalent of compound PPM that is set from the top right box in mainwindow. This value is used in peak detection and has an initial value of 5 ppm, both in the GUI and CLI versions. 

In automated search, duplicate groups are filtered out on the basis of their rt overlap and ppm difference between their mean m/z. The allowed ppm difference depends on the Mass Resolution parameter in Automated Search. The '-p' parameter or 'ppmMerge' is the CLI equivalent of this.

So by adjusting the '-C' parameter, you can change the number of groups that are detected. On the other hand, adjusting the '-p' parameter will only help filter out potential duplicates among the detected groups.

chubukov:
I didn't think there's any filtering on m/z difference in targeted search? If so, this behavior is rather different from the GUI, where it's trivial to get the same group annotated to two different compounds. Let's make them consistent.

Otherwise, if I substitute "untargeted peak detection" into what you wrote, what you describe is basically exactly what _should_ happen. But I know that in older versions I have used the -p parameter and had it work precisely as you describe -C working. So I would just suggesting checking that the value from -p was not also being substituted into the compoundPPMWindow paramter.

Perhaps you've already done this without noticing, or perhaps I'm misunderstanding something still. I can see if there's anything screwy in the new version.

shubhra-agrawal:
@chubukov At present, the CLI and GUI behave differently when it comes to filtering on m/z difference in targeted search. This piece of code is present in both but it's only implemented in CLI. It really should be consistent. I'll make another issue for that. 

I can't find any instance where ppmMerge value is substituted into another variable or where compoundPPMWindow is taking up a value where it should not. The only modules/functions that are using the ppmMerge value are Mass slicing, reduceGroups in CLI and addPeakGroup in PeakDetector. The last one is not being used, as I mentioned above. 

[...]

PR #357, opened 2017-09-01
Ftr: Add PPM window in CLI #356

shubhra-agrawal:
ElMaven GUI allows users to set a range for compound mass in parts per million (ppm) but CLI does not.

This feature has been integrated into CLI in this commit.

Issue: #356


chubukov:
@shubhra-agrawal @sahil21 In practice, I found that the -p (ppmMerge) parameter had exactly this functionality (even though it's supposed to only affect massSlicing I guess). I would look carefully at where the ppmMerge parameter is being used.

PR #365, opened 2017-09-04
Fix: Fix CSV Report when exporting from CLI

IpankajI:
When we were exporting CSV from CLI, we were not getting any data
for for samples because these samples (all samples) were not getting
marked for peak detection

While exporting from CLI, samples are marked for peak detection and we
are getting data instead of NA for all samples

### DIFF
--- a/libmaven/csvreports.cpp
+++ b/libmaven/csvreports.cpp
@@ -61,7 +61,7 @@ void CSVReports::insertGroupReportColumnNamesintoCSVFile(string outputfile,bool 
     if (groupReport.is_open()) {
         QStringList groupReportcolnames;
         groupReportcolnames << "label" << "metaGroupId" << "groupId" << "goodPeakCount"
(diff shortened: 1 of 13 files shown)


libmaven/csvreports.cpp
```diff
@@ -61,7 +61,7 @@ void CSVReports::insertGroupReportColumnNamesintoCSVFile(string outputfile,bool
     if (groupReport.is_open()) {
         QStringList groupReportcolnames;
         groupReportcolnames << "label" << "metaGroupId" << "groupId" << "goodPeakCount"
-                << "medMz" << "medRt" << "maxQuality" << "note" << "compound"
+                << "medMz" << "medRt" << "maxQuality" << "label" << "compound"
```
sahil21: Two columns with the same name "label"

sahil21: @IpankajI Let's put it isotopic label

PR #416, opened 2017-10-03
Fix: Remove binary to new directory <target>

IpankajI:
Previously, binary executables and static libraries were
generated to build folder out side of main directory
which was inconvinient to access.
	Now destination directory for all binary
executables is changed to a new directory <target> which
is a subdirectory of main directory and static libraries
are in <bin> and temprory object files are in <tmp>
folder of this new directory<target>
	These binary are also included in .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,10 +51,10 @@ mzroll/qrc_mzroll.cpp
 # Compiled Binaries
 bin/maven_dev_769
 bin/mzWatcher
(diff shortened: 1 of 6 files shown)


rish9511:
@IpankajI  @sahil21 
This is not making sense to me. Why do we want to use to gitignore to ignore files that should not be part of the src dir(Main repo) in first place ?

PR #417, opened 2017-10-04
Fix grouping parameters

sahil21:


chubukov:
@sahil21 Interesting idea, could be quite helpful. Have you already played with it much? What's the practical difference between adjusting this parameter vs adjusting the smoothing window?

sahil21:
@chubukov I'm still testing it out. Till now it is working well. I was waiting for some fixes in new directory structure branch so that you can compile in windows and test it too. I'll rebase this branch in some time.

sahil21:
@chubukov As smoothing window is doing fine enough job, adding this parameter is redundant. I'm closing this pull request.

chubukov:
@sahil21 sorry, I hope I didn't discourage you -- would also have been totally ok to throw this parameter into the big advanced options list. In general, by all means, please keep thinking about better ways to organize the computations and experiment with new parameters.

sahil21:
@chubukov You are right. We should think about new parameters to make computations easier. While at the same time, after what you suggested on the smoothing parameter, we feel that the parameter we suggested will not add much value for the user. Though, we will still continue to search for new parameters to ease the process.

PR #418, opened 2017-10-04
Ftr widget for same mz n rt

IpankajI:


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -471,14 +494,134 @@ void TableDockWidget::addRow(PeakGroup* group, QTreeWidgetItem* root) {
...
+     * by QEventLoop <loop> and show a prompt to user whether he wants to add
+     * this group to bookmarked groups or not by executing method <showSameGroup>.
+     * <loop> will hold execution of this funtion till user press a button on prompt
+     * dialog.
+    */
+
+
+    int intMz=std::ceil(group->meanMz);
+    int intRt=std::ceil(group->meanRt);
+    int sameMzRtGroupIndexHash=intMz*intRt;
```
shubhra-agrawal: Mz and Rt product can be same for multiple combinations of Mz and Rt. Try another approach maybe.

IpankajI: @shubhra-agrawal QPair is now used as index.

PR #420, opened 2017-10-05
Fix deleted groups exported to json

IpankajI:


IpankajI:
@sahil21 Now working fine for isotopes and compound. see last commit and in-code doc.

sahil21:
@IpankajI Still not working for isotopes.

Issue #431, opened 2017-10-12
Initialization problem in loading some data sets


sahil21:
The data doesn't load properly (EICs are empty) when loading the data just after re-compilation.

Steps to reproduce:
1. Recompile the application (uninstall and run)
2. Load dataset (have seen in data-dependent ms2. Not in ms datasets)
3. Try looking at the compounds from the compound database. EICs will be empty or very few intensity points.
4. Recomputing EICs fixes the problem.

shubhra-agrawal:
@sahil21 Can't reproduce in current develop. Closing the issue.

PR #443, opened 2017-10-26
Fix: Update QSetting on starting #431

IpankajI:
QSettings instance was not updated on starting of ElMaven.

After creating instance of MainWindow, QSettings is updated from
value of GUI forms

Issue: #431

### DIFF
--- a/src/gui/mzroll/main.cpp
+++ b/src/gui/mzroll/main.cpp
@@ -56,6 +56,7 @@ int main(int argc, char *argv[])
     splash.finish(mainWindow);
     mainWindow->show();
     mainWindow->fileLoader->start();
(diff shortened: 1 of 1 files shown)


chubukov:
@IpankajI @sahil21 Is this likely to fix the bug we have had forever where it's necessary to uncheck and re-check the "report isotopic peaks" checkbox every time we start maven on windows? By the way, I noticed that the same thing was happening with the "consider overlap" checkbox in the grouping options tab.

IpankajI:
@chubukov No, this PR is not resolving that issue. We are looking into it.

shubhra-agrawal:
@IpankajI It's crashing either on upload or when I open the Options dialog.

Issue #341, opened 2017-08-21
Mass Spectra not picking up the right scan


sahil21:
When the user clicks on a peak in the EIC window, mass spectra picks up the consecutive scan. Most probably it is an indexing error. 

This issue came up when we were working on the branch **ftr_chooseFilterline**. 
ftr_chooserFilterline branch has a feature in which the user can choose the filterLine (or segments) from a data. We tried to process a data with 3 segments. The data was such that after every scan the segment changes. Every time you choose one segment, the mass spectra show the scan having the consecutive next segment.

IpankajI:
@sahil21 resolved and merger to develop.

PR #447, opened 2017-10-27
Fix: Assign right scan no. for MS/MS #341

IpankajI:
scan no. was updating wrong for ms/ms- having
a increment of 1.

Issue: #341


chubukov:
This was specific for the mzml format? Are we sure there's not the same issue with mzxml or others?

IpankajI:
@chubukov We have tested against couple of data sets( 3 mzxml and 3 mzml data sets) and also have reviewed the code for this.This problem was only with .mzml format file.There, we were getting wrong result for mass spectra while corresponding peak in EIC plot had different scan. See #341

chubukov:
Great. There's basically no relationship between the file format and the data type, so I would call it an mzml issue, not an MS/MS issue. Glad you solved it.

Issue #185, opened 2017-03-22
no isotopes in barplot or bookmark if checkbox is disabled in peak detection dialog


chubukov:
You don't get any isotopic species bookmarked or showing up in the barplot if the "report isotopic peaks" checkmark is disabled in the peak detection dialog.

Is this intentional? I thought the idea was that the options in the peak detection dialog box would only affect the automated search, and everything related to GUI behavior would be in the main options.

I'm working off the develop branch as of 3/22.

sahil21:
@chubukov Yes, this feature was intentional. But having another option in the "Options" Dialog to disable isotope detection makes more sense. We'll fix it as soon as possible.

chubukov:
Probably irrelevant if this is being fixed, but just wanted to note that there's also an initialization bug here. When you first open Maven, you have to uncheck and recheck the box to get isotopes to show up in the barplot and manual bookmarking.

shubhra-agrawal:
@chubukov The initialization bug has been resolved and merged. We will look into moving the option out of Peaks dialog in the next release.

shubhra-agrawal:
@chubukov Just wanted to note that the barplot is independent of the checkbox now. (v0.4.0 and above). 

With respect to moving Report isotopic peaks out of the "Peaks" dialog, we have decided on a design for it. Should be out in the next few releases once we do a feedback session on the new designs. I'm leaving this open till then.

saifulbkhan:
Closing this since all issues discussed here have been resolved. A dedicated dialog has been added to the main toolbar in El-MAVEN v0.9.0.

PR #454, opened 2017-11-10
Fix: Fix initialization bug in report isotopic peaks  #185 #453

shubhra-agrawal:
User had to uncheck and recheck "Report Isotopic Peaks" in the Peaks dialog on opening El-MAVEN.

This commit fixes this issue.

Issues: #453 #185


chubukov:
If this was straightforward after all, can you check if the same thing is happening with the "consider overlap" checkbox in the options? Maybe other ones too, but that's the one I observed.

Issue #424, opened 2017-10-07
Unit tests for MS/MS functions


shubhra-agrawal:
Three sample LCMS files have been uploaded to the repo to run unit tests for which has significantly increased the repo size. Adding ms/ms sample files for unit testing ms/ms specific functions is likely not an ideal solution.

chubukov:
@shubhra-agrawal I think I mentioned this idea before, but you could just create a few minimal mzXML files (pull out just a small range of scans and chromatogram window from some existing file). Those would be easier to use for testing anyway.

PR #472, opened 2017-11-28
Test: Create short files for ms2 testing #424

IpankajI:
For ms/ms testing to run, we need files in repository, but
these files are to large and increasing repo size to great extent.
	Three test files has been created by extracting scans from
existing files which have short size combined

Issue: #424


sahil21:
@IpankajI What all scan types are included in these files?

PR #473, opened 2017-11-28
Fix: Update sorting algo for tree item #466

IpankajI:
Default tree item is sorted lexicographcal order of string
that could be number also.
	It's updated - if there are numbers, then  numbers will be
sorted otherwise string will be given larger value and sorted.

Issue: #466


sahil21:
@IpankajI This branch is not working. 
See here:
image

PR #486, opened 2017-12-08
Fix: Update Isotopes Widget listing mode #453

IpankajI:
Isotopes widget was showing all isotopes for a group when
Report Istopic Peaks was checked and when it was unchecked, it was
showing only parent.
	It's been updated to show all isotopes always.

Issue: #453


chubukov:
Does it still rely on the check boxes in the isotopes tab of the main options window? The idea there was that you could have all isotopes in the widget but only bookmark some of them. Not that this is a critical feature.

IpankajI:
@chubukov This [commit](https://github.com/ElucidataInc/ElMaven/pull/486/commits/e14fff9d8ada78904e043e75bbfac778bf204d04) has only made Isotopes Widget independent of Report Isotopic Peaks check-box. It is still dependent on check-boxes of isotopes tab of the main options window. We'll resolve that as part of [this](https://github.com/ElucidataInc/ElMaven/issues/489) issue.

PR #498, opened 2017-12-20
Ftr cal smooth area

sahil21:


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1768,8 @@ void TableDockWidget::readPeakXML(QXmlStreamReader& xml,PeakGroup* parent) {
     p.pos = xml.attributes().value("pos").toString().toInt();
     p.minpos = xml.attributes().value("minpos").toString().toInt();
     p.maxpos = xml.attributes().value("maxpos").toString().toInt();
+    p.splineminpos = xml.attributes().value("splineminpos").toString().toInt();
```
shubhra-agrawal: @sahil21 This function reads data from mzroll files. Are we handling the case of older versions of .mzroll files which do not have this attribute?

sahil21: @shubhra-agrawal It will work fine for previous versions of mzroll. The spline area will be zero in that case (which is default value)

src/core/libmaven/Peak.h
```diff
@@ -40,6 +42,8 @@ class Peak {
 
         /** non corrected sum of all intensities */
         float peakArea;
+        /** are of spline */
```
rish9511: area of spline *

Issue #469, opened 2017-11-22
BUG - Copy isotopic detail to clipboard copies incorrect values


lparsons:
As an example, when copying NADP+ data, the data for label 21 was instead copied to label 14. The bar plot in the software seems correct, but the table copied is wrong.

shubhra-agrawal:
@lparsons Order of the labels and the data were different. This has been resolved and merged to develop. Let us know when you test it out.

lparsons:
@shubhra-agrawal This does appear to be fixed, at least in my limited testing so far. Thanks!

PR #500, opened 2017-12-20
Fix: Wrong isotope label order in clipboard #469

shubhra-agrawal:
Order of isotope labels did not match the order of isotope information in the clipboard.

Issue: #469


src/gui/mzroll/isotopeswidget.cpp
```diff
@@ -311,7 +311,7 @@ QString IsotopeWidget::groupIsotopeMatrixExport(PeakGroup* group, bool includeSa
                 isotopes.push_back(isotope);
 			}
 		}
-		std::sort(isotopes.begin(), isotopes.end(), PeakGroup::compC13);
+		//std::sort(isotopes.begin(), isotopes.end(), PeakGroup::compC13);
```
rish9511: are we not sure we won't be using it the future ?

shubhra-agrawal: Yes, there's no point in sorting by C13 count here. The output should be the same as what you see in the Peak Table. Plus, this sorting is already disabled in further functions as well.

PR #502, opened 2017-12-20
Ftr pull injection time

sahil21:


src/core/libmaven/mzSample.cpp
```diff
@@ -308,6 +315,20 @@ void mzSample::parseMzML(const char *filename)
 	}
 }
 
+void mzSample::parseMzMLInjectionTimeStamp(xml_node experimentRun)
```
rish9511: use reference instead of just copying the entire object

sahil21: @rish9511 Just trying to be consistent here. There are other parseMLML functions also.

rish9511: that's also bad though. okay, your call

src/core/libmaven/mzSample.h
```diff
@@ -711,6 +711,15 @@ class mzSample
 
     static bool compSampleSort(const mzSample *a, const mzSample *b) { return mzUtils::strcasecmp_withNumbers(a->sampleName, b->sampleName); }
 
+    /**
+    * @brief Compare injection time (in epoch seconds) of two samples
+    * @param a object of class mzSample
+    * @param b object of class mzSample
+    * @return True if mzSample a has lower injection time than mzSample b
+    */
+    static bool compInjectionTime(const mzSample *a, const mzSample *b) { return a->injectionTime < b->injectionTime; }
```
rish9511: why this has to be a static function ?

sahil21: @rish9511 So that we can use the function without object (like mzSample::compInjectionTime)

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1138,6 +1141,18 @@ void MainWindow::setQComboBox() {
 	else ionizationModeComboBox->setCurrentIndex(2);
 }
 
+void MainWindow::setInjectionOrderFromTimeStamp() {
+
+    vector<mzSample*> samples = getSamples();
+    std::sort(samples.begin(), samples.end(), mzSample::compInjectionTime);
```
rish9511: is there a reason for using  a custom function for comparing? std::sort by default sorts in ascending order

src/core/libmaven/mzSample.cpp
sahil21: @rish9511 Updated the code

src/gui/mzroll/mainwindow.cpp
sahil21: As it is a user-defined class, we can't use default sorts here.

rish9511: oh correct. my bad

rish9511: In future we should use lambdas for such things . let's try not to use static functions unless they are really required

sahil21: @rish9511 Yes, makes sense.

PR #503, opened 2017-12-21
Show Peak Area and Natural Abundance fractions in clipboard

shubhra-agrawal:


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2253,7 +2253,7 @@ void MainWindow::readSettings() {
 
     //Pull Isotopes in options
     if (!settings->contains("isotopeC13Correction"))
-        settings->setValue("isotopeC13Correction", 2);
+        settings->setValue("isotopeC13Correction", 0);
```
rish9511: this can still be a problem on the user's system, no ? if they don't uninstall or remove 'Application Settings.conf'  that value won't change for them

shubhra-agrawal: @rish9511 Will this be a problem with all default settings that are not part of mavenParameters?

src/gui/mzroll/isotopeswidget.cpp
```diff
@@ -325,6 +325,19 @@ QString IsotopeWidget::groupIsotopeMatrixExport(PeakGroup* group, bool includeSa
 						}
 						isotopeInfo += groupInfo.join("\t") + "\n";
 				}
+				if (MM.cols() > isotopes.size()) {
+					isotopeInfo += "Natural Abundance\n";
+					for (int i = 0, k = isotopes.size(); i < isotopes.size(); i++, k++) {
+						QStringList groupInfo;
+                        groupInfo << tag + " " + QString(isotopes[i]->tagString.c_str());
+						for( unsigned int j=0; j < vsamples.size(); j++) {
+								//qDebug() << i << " " << j << " " << MM(j,i);
```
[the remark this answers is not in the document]
shubhra-agrawal: @rish9511 Sure. Formatting as well.

src/gui/mzroll/mainwindow.cpp
shubhra-agrawal: @rish9511 Added variable to mavenParameters.

PR #508, opened 2017-12-28
Fix: update initialization of ionizationMode

IpankajI:
ionizationMode was updated with the index value not with the
actual value

### DIFF
--- a/src/core/libmaven/mavenparameters.cpp
+++ b/src/core/libmaven/mavenparameters.cpp
@@ -232,8 +232,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
 {
     mavenSettings[const_cast<char*>(key)] = const_cast<char*>(value);
 
-    if(strcmp(key, "ionizationMode") == 0)
-        ionizationMode = atof(value);
+    if(strcmp(key, "ionizationMode") == 0){
+        int indexOfIonizationMode=*value-'0';
+        /**
+         * map index of ionization-mode to value (-1,0,1) of ionizationMode
+         */
+        if(indexOfIonizationMode==1) ionizationMode=0;
+        else if(indexOfIonizationMode==2) ionizationMode=+1;
+        else if(indexOfIonizationMode==3) ionizationMode=-1;
+    }
 
     if(strcmp(key, "amuQ1") == 0)
         amuQ1 = atof(value);
--- a/src/core/libmaven/mavenparameters.cpp
+++ b/src/core/libmaven/mavenparameters.cpp
@@ -233,13 +233,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
     mavenSettings[const_cast<char*>(key)] = const_cast<char*>(value);
 
     if(strcmp(key, "ionizationMode") == 0){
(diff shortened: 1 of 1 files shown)


src/core/libmaven/mavenparameters.cpp
```diff
@@ -232,8 +232,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
...
     mavenSettings[const_cast<char*>(key)] = const_cast<char*>(value);
 
-    if(strcmp(key, "ionizationMode") == 0)
-        ionizationMode = atof(value);
+    if(strcmp(key, "ionizationMode") == 0){
+        int indexOfIonizationMode=*value-'0';
+        /**
+         * map index of ionization-mode to value (-1,0,1) of ionizationMode
+         */
+        if(indexOfIonizationMode==1) ionizationMode=0;
```
shubhra-agrawal: MainWindow::setTotalCharge() does the same thing. Check that function out.

rish9511: @shubhra-agrawal  we still need to do it here. if we don't then, every call to this function will change the ionizationMode to hold the index and not the actual value

shubhra-agrawal: @rish9511 Yes, we need to do it here. I merely suggested using the already defined function here if possible. Short answer is, we cannot. So this is fine.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -232,8 +232,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
 {
     mavenSettings[const_cast<char*>(key)] = const_cast<char*>(value);
 
-    if(strcmp(key, "ionizationMode") == 0)
-        ionizationMode = atof(value);
+    if(strcmp(key, "ionizationMode") == 0){
+        int indexOfIonizationMode=*value-'0';
```
shubhra-agrawal: @IpankajI Suggest using atoi() or atod() here for consistency.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -232,8 +232,17 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
...
     mavenSettings[const_cast<char*>(key)] = const_cast<char*>(value);
 
-    if(strcmp(key, "ionizationMode") == 0)
-        ionizationMode = atof(value);
+    if(strcmp(key, "ionizationMode") == 0){
+        int indexOfIonizationMode=atoi(value);
+        /**
+         * map index of ionization-mode to value (-1,0,1) of ionizationMode
+         */
+        if(indexOfIonizationMode==1) ionizationMode=0;
```
shubhra-agrawal: @IpankajI Shouldn't the indices be 0,1 and 2?

IpankajI: @shubhra-agrawal 0 is for Auto-Detect, rest is tested.

shubhra-agrawal: @IpankajI I was confusing this drop-down with the one on main window. My bad.

PR #518, opened 2018-01-04
Fix: copy to clipboard for isotopic children shows compound name

shubhra-agrawal:
Copying isotopic information to clipboard for isotopic children copied the tagString twice instead of adding compound name.

### DIFF
--- a/src/gui/mzroll/isotopeswidget.cpp
+++ b/src/gui/mzroll/isotopeswidget.cpp
@@ -347,9 +347,13 @@ QString IsotopeWidget::groupIsotopeMatrixExport(PeakGroup* group, bool includeSa
 			}
 
 			_mw->setStatusText("Clipboard set to isotope summary");
-		} else {
-                isotopeInfo += tag + " | " + groupTextEport(group) + "\n";
-				_mw->setStatusText("Clipboard to group summary");
+		} 
+		else {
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/isotopeswidget.cpp
```diff
@@ -327,7 +328,7 @@ QString IsotopeWidget::groupIsotopeMatrixExport(PeakGroup* group, bool includeSa
 			MatrixXf MM = _mw->getIsotopicMatrixIsoWidget(group);
 			for (int i=0; i < isotopes.size(); i++ ) {
 				QStringList groupInfo;
-                groupInfo << tag + " " + QString(isotopes[i]->tagString.c_str());
+                groupInfo << tag + " | " + QString(isotopes[i]->tagString.c_str());
```
rish9511: If you find peaks without isotopes and then copy details from peaks table , the compound name is copied with "|". eg. 
" Pyridoxamine | 	21918.21	22437.19	16426.92	24983.37"
I think that happens because  isotopes size is never zero.
If you think it should be fixed now, then do it else _ship it!_

shubhra-agrawal: @rish9511 fixed it

PR #519, opened 2018-01-04
Fix polarity assignment

IpankajI:


src/core/libmaven/mavenparameters.cpp
```diff
@@ -26,7 +26,7 @@ MavenParameters::MavenParameters()
         outputdir = "reports" + string(DIR_SEPARATOR_STR);
 
         writeCSVFlag = false;
-        ionizationMode = -1;
+        ionizationMode = 1;
```
shubhra-agrawal: Any particular reason for changing the default?

src/core/libmaven/mavenparameters.cpp
```diff
@@ -240,8 +240,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
...
         else if(indexOfIonizationMode==3) ionizationMode=-1;
-        // set auto to zero
-        else ionizationMode=0;
+        else {
+            // auto-detection
+            if(samples.empty()){
+                ionizationMode=+1;
+            }
+            else{
+                ionizationMode=samples[0]->getPolarity();
```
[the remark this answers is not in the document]
shubhra-agrawal: @rish9511 samples[0] is in the else statement.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -240,8 +240,15 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
         if(indexOfIonizationMode==1) ionizationMode=0;
         else if(indexOfIonizationMode==2) ionizationMode=+1;
         else if(indexOfIonizationMode==3) ionizationMode=-1;
-        // set auto to zero
-        else ionizationMode=0;
```
shubhra-agrawal: We can just call the setIonizationMode() function here, right?

IpankajI: @shubhra-agrawal Requested change is made!

src/core/libmaven/mavenparameters.cpp
IpankajI: @shubhra-agrawal +1 is used as default value at every place.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1117,33 +1116,30 @@ void MainWindow::setUserMassCutoff(double x) {
 
 void MainWindow::setIonizationMode(int x) {
 	_ionizationMode = x;
+	mavenParameters->ionizationMode=x;
```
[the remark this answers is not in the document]
IpankajI: @rish9511 MavenParameters::setIonizationMode is only for auto-detection.

src/core/libmaven/mavenparameters.cpp
shubhra-agrawal: @IpankajI Default was -1 earlier. Why change it? There's a possibility default is assumed to be -1 in other parts of the code. It's a risque move in my opinion. We shouldn't change defaults without a strong reason.

src/core/libmaven/mavenparameters.cpp
rish9511: @IpankajI  @shubhra-agrawal  should we not use setIonizationMode() for handling all the cases? why just auto-detect?  Imo the function should be modified to handle all the cases.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -542,9 +542,8 @@ using namespace mzUtils;
 		setIonizationMode(settings->value("ionizationMode").toInt());
 	}
 
-	setQComboBox();
+	setIonizationModeLable();
```
rish9511: @IpankajI  misspelled  'Label'

src/gui/mzroll/mainwindow.cpp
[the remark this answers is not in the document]
rish9511: it will be better if it's modified to handle the all the cases. what do you think ?

src/core/libmaven/mavenparameters.cpp
IpankajI: @shubhra-agrawal  Previously, It was +1 at few places and -1 at some places. I changed it to +1 everywhere

IpankajI: @shubhra-agrawal While uploading samples, it still will auto detect ionization-mode from samples.

IpankajI: @shubhra-agrawal While uploading samples, it still will auto detect ionization-mode from samples.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -437,11 +431,25 @@ int MavenParameters::getCharge(Compound* compound){
  * MavenParameters::setIonizationMode In this the mode is selected my looking
  * at the first sample polarity
  */
-void MavenParameters::setIonizationMode() {
-        if (samples.size() > 0 && samples[0]->getPolarity() > 0)
-                ionizationMode = +1;
-        else
-                ionizationMode = -1;  //set ionization mode for compound matching
+void MavenParameters::setIonizationMode(Polarity polarity) {
```
rish9511: :+1:

Issue #487, opened 2017-12-09
EIC from text search and compound widget do not match


shubhra-agrawal:
version: Current develop
Data: MRM

EICs from text search look a lot more choppy than those from compound widget/SRM widget.

shubhra-agrawal:
Tracked this down to different EIC vector sizes/patterns.

Text search and compound widget use two different getEIC functions. One each for SRM id and Precursor/Product Mz as input. Both of these return a different sized vector for the same data. This is not to say that either one of them have missing data. Because both these functions scan for data points in a different way using different number of loop runs, the distance between actual data points in the EIC vector are different. Lesser the distance between distinct data points, more the extrapolation, leading to choppy peaks. 

This bug does not affect LCMS and PRM data as text search and compound widget use the same getEIC functions in those cases.

saifulbkhan:
I bet this was actually because of the whole multiple transitions for same precursor-product pair thing again. I will get rid of scans altogether when reading MRM data and create EICs directly, instead. The UX around these issues will still need some thought.

PR #542, opened 2018-01-16
Fix: Remove duplicate Rt from MRM EIC vector#487

shubhra-agrawal:
In case of multiple (n) fragments of the same precursor m/z, Rt vector had n values for every Rt with one corresponding non-zero intensity.

This fix ensures that the getEIC function works taking both MRM and PRM data under consideration.

Issue: #487

### DIFF
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -1256,17 +1256,16 @@ EIC *mzSample::getEIC(float precursorMz, float collisionEnergy, float productMz,
 		}
 		}
 
-		int lastIndex = e->rt.size()-1;
 		//if rt is already present save the higher intensity for that rt
 		//this can happen when there are multiple product m/z for the same precursor
-		if (lastIndex != -1 && scan->rt == e->rt[lastIndex]) {
-			if (eicIntensity <= e->intensity[lastIndex]) 
+		if (!e->rt.empty() && e->rt.back() == scan->rt) {
+			if (eicIntensity <= e->intensity.back()) 
 				continue;
 			else {
-				//save the new intensity
-				e->scannum[lastIndex] = scan->scannum;
-				e->intensity[lastIndex] = eicIntensity;
-				e->mz[lastIndex] = eicMz;
+				//replace old values for the rt
+				e->scannum.back() = scan->scannum;
+				e->intensity.back() = eicIntensity;
+				e->mz.back() = eicMz;
 			}
 		}
 		//save values for new rt


src/core/libmaven/mzSample.cpp
```diff
@@ -1257,16 +1257,19 @@ EIC *mzSample::getEIC(float precursorMz, float collisionEnergy, float productMz,
 		}
 
 		int lastIndex = e->rt.size()-1;
+		//if rt is already present save the higher intensity for that rt
+		//this can happen when there are multiple product m/z for the same precursor
 		if (lastIndex != -1 && scan->rt == e->rt[lastIndex]) {
```
rish9511: Correct me wherever I am wrong. Here we are checking two things:
1. whether the vector is empty:  for this we can simply use the empty() function
2. if eic already has the rt:  for this we can use the std::find() function. This will also help us in getting the index

I would suggest using the standard functions.

src/core/libmaven/mzSample.cpp
```diff
@@ -1257,16 +1257,19 @@ EIC *mzSample::getEIC(float precursorMz, float collisionEnergy, float productMz,
...
 			if (eicIntensity <= e->intensity[lastIndex])
```
rish9511: std::find() can be used to get the index. e.g  vector.begin() - std::find(vector.begin(), vector.end(),  value)
This can stored in the previous step, where we try to find the  rt

src/core/libmaven/mzSample.cpp
```diff
@@ -1257,16 +1257,19 @@ EIC *mzSample::getEIC(float precursorMz, float collisionEnergy, float productMz,
...
 		int lastIndex = e->rt.size()-1;
+		//if rt is already present save the higher intensity for that rt
+		//this can happen when there are multiple product m/z for the same precursor
 		if (lastIndex != -1 && scan->rt == e->rt[lastIndex]) {
 			if (eicIntensity <= e->intensity[lastIndex]) 
 				continue;
 			else {
-				//to prevent duplicate RT and append only relevant intensities
+				//save the new intensity
 				e->scannum[lastIndex] = scan->scannum;
```
rish9511: we can use the index obtained above or use the vector.insert() fucntion

src/core/libmaven/mzSample.cpp
shubhra-agrawal: @rish9511 Thanks. I'll do that.

shubhra-agrawal: @rish9511 done.

src/core/libmaven/mzSample.cpp
```diff
@@ -1256,10 +1256,25 @@ EIC *mzSample::getEIC(float precursorMz, float collisionEnergy, float productMz,
 		}
 		}
 
-		e->scannum.push_back(scan->scannum);
-		e->rt.push_back(scan->rt);
-		e->intensity.push_back(eicIntensity);
-		e->mz.push_back(eicMz);
+		//if rt is already present save the higher intensity for that rt
+		//this can happen when there are multiple product m/z for the same precursor
+		if (!e->rt.empty() && e->rt.back() == scan->rt) {
```
sahil21: Any specific reason to use vector::back?
Here we are assuming that scans are in ascending order, right?

shubhra-agrawal: That is the assumption as we are going scan by scan. If there is a duplicate value, it will be present at the last index. Using std::find would be wasteful.

rish9511: As pointed out by @sahil21 , this will only work if are assumption , that the scans are sorted, is actually correct.
If it's not sorted, then we will end up inserting duplicates in the EIC

Also, why find() would be wasteful? The complexity will still remain the same. i.e O(n^2)

shubhra-agrawal: @rish9511 Sorted scans are less an assumption, more of a rule. 
find() searches the whole vector for the value. Back() will just access the last element which is what we need.

PR #543, opened 2018-01-17
Fix dump load

rish9511:
The main highlights of the change:
- Stop using Qt Settings to store PeakDetection Dialog and Options Dialog settings 
- Use lastRun.xml to store and load back last session settings and load back


chubukov:
@rish9511 it looks like you're storing lastrun.xml in the installation directory. Couple of issues with that
1) user might not have permission to write to that directory in many cases
2) multiple users won't be able to maintain their own settings

Can we find a user-specific location to put this into? Some quick googling suggests that QT can do this in a cross platform way pretty easily.

rish9511:
@chubukov   

> user might not have permission to write to that directory in many cases

Thanks for pointing this out. How about storing this in the config location, in Linux it's the  `~/.config/` directory and on windows it is `C:/Users/<USER>/AppData/Local/<APPNAME>` . This also where we store 'Application Settings.conf' file.  

>multiple users won't be able to maintain their own settings

I am not sure If understand this. Could you please elaborate more ?

>Some quick googling suggests that QT can do this in a cross platform way pretty easily.

I am aware of this . [QStandardpaths](http://doc.qt.io/qt-5/qstandardpaths.html) provides what is required to implement this

@sahil21: we had a small discussion about this. Could you also give your thoughts around this ?

chubukov:
@rish9511 That sounds fine to me. It would take care of the other concern I mentioned.

sahil21:
@rish9511 Branch is failing on travis.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -131,15 +152,38 @@ std::map<string, string>& MavenParameters::getSettings()
...
 
 
     if(strcmp(key, "automatedDetection") == 0 )
         processAllSlices = atof(value);
 
-//     TODO: replace with massCutoffMerge. remove ppmMerge from default_settings_xml as well. and regenerate settings.h 
-//     if(strcmp(key, "ppmStep") == 0 )
-//         ppmMerge = atof(value);
+
+     if(strcmp(key, "massCutoffType") == 0 ) {
```
sahil21: Aren't these 3 separate variables?

src/core/libmaven/mavenparameters.cpp
```diff
@@ -168,12 +212,16 @@ void  MavenParameters::setPeakDetectionSettings(const char* key, const char* val
...
         maxCharge = atof(value);
 
-    if(strcmp(key, "dbDetection") == 0 )
+    if(strcmp(key, "dbDetection") == 0 );
         //TODO
 
-        //TODO: replace with compoundMassCutoffWindow. replace compoundPPMWindow in default_settings_xml. regenerate settings.h
-//     if(strcmp(key, "compoundPPMWindow") == 0 )
-//         compoundPPMWindow = atof(value);
+     if(strcmp(key, "compoundMassCutoffWindow") == 0 ) {
```
sahil21: compoundMassCutoffWindow is being used above also?

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -186,53 +186,8 @@ void BackgroundPeakUpdate::loadSettings(QString fileName) {
             if (xml.name() == "PeakDetectionSettings") {
                 settings->setValue("alignSamplesFlag", xml.attributes().value("alignSamplesFlag").toString().toInt());
```
sahil21: Why hasn't this been removed?

src/core/libmaven/mavenparameters.h
```diff
@@ -20,7 +20,8 @@ class Classifier;
 class MavenParameters 
 {
     public:
-        MavenParameters();
+        MavenParameters(string settingsPath="");
```
sahil21: Why an optional argument rather than a separate constructor? Just asking

src/gui/mzroll/peakdetectiondialog.h
```diff
@@ -25,6 +25,7 @@ class PeakDetectionDialog : public QDialog, public Ui_PeakDetectionDialog
 				 ~PeakDetectionDialog();
 				 void setSettings(QSettings* settings) { this->settings = settings; }
 				 void setMainWindow(MainWindow* w) { this->mainwindow = w; }
+                 MainWindow* getMainWindow() {return mainwindow;}
```
sahil21: Is it a good idea to have Mainwindow pointer here? If no, are we planning to move?

src/gui/mzroll/settingsform.cpp
```diff
@@ -32,7 +32,7 @@ OptionsDialogSettings::OptionsDialogSettings(SettingsForm* dialog): sf(dialog)
     settings.insert("noOfIsotopes", QVariant::fromValue(sf->noOfIsotopes));
     settings.insert("isotopeC13Correction", QVariant::fromValue(sf->isotopeC13Correction));
     settings.insert("minIsotopicCorrelation", QVariant::fromValue(sf->minIsotopicCorrelation));
-    settings.insert("maxIsotopeScanDiff", QVariant::fromValue(sf->minIsotopicCorrelation));
+    settings.insert("maxIsotopeScanDiff", QVariant::fromValue(sf->maxIsotopeScanDiff));
```
IpankajI: Are these insertion and deletion ok?

tests/MavenTests/testCLI.cpp
```diff
@@ -61,8 +61,8 @@ void TestCLI::testProcessXml() {
     QVERIFY(peakdetectorCLI->mavenParameters->processAllSlices == 0);
     QVERIFY(peakdetectorCLI->mavenParameters->C13Labeled_BPE == true);
     QVERIFY(peakdetectorCLI->mavenParameters->N15Labeled_BPE == true);
-    QVERIFY(peakdetectorCLI->mavenParameters->S34Labeled_BPE == false);
-    QVERIFY(peakdetectorCLI->mavenParameters->D2Labeled_BPE == false);
+    QVERIFY(peakdetectorCLI->mavenParameters->S34Labeled_BPE == true);
+    QVERIFY(peakdetectorCLI->mavenParameters->D2Labeled_BPE == true);
```
IpankajI: @rish9511 I don't understand why this change is ok. You have  put false as true.

src/core/libmaven/mavenparameters.cpp
rish9511: no. it tells whether the type is 'ppm' or 'mDa'

src/core/libmaven/mavenparameters.cpp
rish9511: where ? other than here, it's set in constructor.

src/core/libmaven/mavenparameters.h
rish9511: Why have two of them when one is sufficient ? 
If you don't provide the 'settingsPath', which is what happens when mavenParameters is initialized from 'test' module, it will simply load the default settings. Having a different constructor means making the other constructor behave the same way when no 'settingsPath'  is provided, which in this case is handled automatically.

does it make sense ?

src/gui/mzroll/background_peaks_update.cpp
rish9511: I was focusing only  on removing peaks and options dialog settings.
Let's do this when doing the same with alignmentDialog ?

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -398,48 +345,6 @@ void BackgroundPeakUpdate::getPullIsotopeSettings() {
         if (mainwindow) {
                 QSettings* settings = mainwindow->getSettings();
                 if (settings) {
-                        mavenParameters->maxIsotopeScanDiff = settings->value(
```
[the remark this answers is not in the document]
rish9511: yes. the entire function is

src/gui/mzroll/main.cpp
```diff
@@ -49,16 +49,16 @@ int main(int argc, char *argv[])
     app.processEvents();
 
 
-    Controller* contrl = new Controller();
+    Controller contrl;
```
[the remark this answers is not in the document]
rish9511: This is done to take care of what happens when El-maven shuts downs. By having it on stack, it's destructor is called  when it goes out of scope, which only happens when the main event loop returns(when app exits), and therefore allows us to handle what happens during the shutdown

src/gui/mzroll/peakdetectiondialog.h
rish9511: The child widgets generally know who their parent is, in this case it's mainwindow.  I think there should be a class called 'Toolbar'  or maybe something else , and that should ideally be the parent of all the dialogs we have. 
Having everything under mainwindow is not good. For now we don't have better solution

src/gui/mzroll/settingsform.cpp
rish9511: I don't understand. Can you be more specific ?

src/core/libmaven/mavenparameters.cpp
sahil21: Got it

src/core/libmaven/mavenparameters.cpp
sahil21: I confused between type and window

src/core/libmaven/mavenparameters.h
sahil21: I prefer the other way. but this is also fine.

src/gui/mzroll/background_peaks_update.cpp
sahil21: Makes sense.

src/gui/mzroll/peakdetectiondialog.h
sahil21: We can change that later then.

Issue #496, opened 2017-12-19
Calculate Area of smoothed spline 


sahil21:
Related to #474

sahil21:
@chubukov What should be the spline bounds for shift drag integration? min and max position of mouse drag event or actual bounds of spline?

Example - 
![image](https://user-images.githubusercontent.com/21240516/35083145-a26ba546-fc43-11e7-8606-f9ca9651a1c7.png)

chubukov:
@sahil21 there was a comment on this thread which is now gone in which you asked an excellent question about how shift-drag should actually operate. I commented here: #528

sahil21:
@chubukov I was going to post it again. (Earlier I thought the question was not clear)

Here is the question (for documentation purpose):

What should be the spline peak bounds in case of shift-drag integration? Whether it should be min and max position of mouse drag event or actual spline bounds.

shubhra-agrawal:
@sahil21 @chubukov SplineArea is being calculared using the dragged bounds, exactly like peakArea. Available in v0.7.0 and above

Issue #528, opened 2018-01-12
shift-drag and area calculation


chubukov:
I was trying to figure out whether integration by shift-drag actually sets the peak bounds for Area calculation, or whether it only identifies the peak. I took a peak and shift-dragged such that I covered the center but not the entire peak area. 

It turns out it's inconsistent. For the "parent" it looks like it does restrict the left/right bounds for the Area calculation. But for the "C12 Parent" I get the larger area that would get by shift-dragging over the entire peak.

chubukov:
I took a survey of our group, and it was pretty unanimous that the desired behavior was to use the dragged bounds directly to set the peak bounds and calculate area or smoothed (spline) area. (i.e. each sample would get a peak created with the exact same bounds).

chubukov:
That said, I haven't looked carefully at how this might affect some other metrics like quality.

saifulbkhan:
Similar to #344, region of integration for isotopes can be restricted to parent peak's x-range. Closing this.

PR #544, opened 2018-01-18
Fix: Calculate spline area for shift drag integration #496

sahil21:
Peak bounds were not being calculated for peaks found out by
shift drag  integration

Issue: #496


chubukov:
@sahil21 see https://github.com/ElucidataInc/ElMaven/issues/528#issuecomment-358660577 . I think we should probably set the peak bounds to be the dragged region

sahil21:
@chubukov Sure. I'll modify this code. Thanks!

PR #546, opened 2018-01-18
Fix aligment

IpankajI:
This PR is resolving crash issue in alignment, updated method to calculate R-squared, infinite loop in alignment. It is adding a dialog box for error handling if something is not working in alignment. Still python algorithm has to be updated which will be done in different branch.
Python executable(223mb, only for linux) is stored with git lfs.


bin/alignment.py
```diff
@@ -196,16 +196,17 @@ def processData(json_obj):
...
-        sys.stdout.flush()
-        print 'stop'
-        sys.stdout.flush()
-    input_json += line
+# sent back processed data
+sys.stdout.write(processedData)
+sys.stdout.flush()
+
+# must exit the program manully
+sys.exit(0)
```
rish9511: code should  be made to exit from other places too. e.g. if the algorithm fails, then what happens ? Should the code not exit in that case ?

bin/alignment.py
```diff
@@ -196,16 +196,17 @@ def processData(json_obj):
         samples_data)
     output_json = json.dumps(output_dict)
     return output_json
+# read input till the EOF
+for line in iter(sys.stdin.readline,''):
+    input_json += line
 
+# process data
+json_obj = json.loads(input_json)
+processedData = processData(json_obj)
```
rish9511: before passing 'json_obj' to 'processData'  should it not be checked whether it is a valid json or not ? In case if the 'json' is malformed, 'processData()' will complain about it and eventually fail to execute.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -587,68 +571,49 @@ void BackgroundPeakUpdate::align() {
...
                         runPythonProg(&aligner);
-
-                        /**readDataFromPython(data)
-                         * it will wait for python to send the corrected rts
-                         * once we receive the corrected rts, we kill the python exe and update the rts in maven
-                         */
-                        QByteArray data;
-                        readDataFromPython(data);
+                        // wait for processing of data by python program
+                        pythonProg->waitForFinished(-1);
```
rish9511: Why not use [QProcess::finshed()](https://doc.qt.io/qt-5/qprocess.html#finished) to handle things after the process has been completed. We can send proper exit status from python and  use it here  to check if python algo failed/succeeded. 
What do you think ?

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -587,11 +587,21 @@ void BackgroundPeakUpdate::align() {
                         // if jDoc is null that means the json returned from python is malformed
                         // in such a case our rts wont update with new values
                         jDoc = QJsonDocument::fromJson(processedDataFromPython);
+
+                        QString errorMessage=QString::number(processedDataFromPython.size());
+                        errorMessage="Received data: "+errorMessage+" bytes";
```
rish9511: It does not make sense to start forming the error message at this point. We don't know whether we will actually use it

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -587,11 +587,21 @@ void BackgroundPeakUpdate::align() {
...
+
                         if(!jDoc.isNull()){
                                 parentObj = jDoc.object();
                         }
                         else{
-                                qDebug()<<"Error: alignment- data is not complete";
+                                if(processedDataFromPython.size()==0){
```
rish9511: The whole error checking should be performed separately,no ? We should first check the exit status from python and then perform these checks to form the error message. 
Thoughts?

src/gui/mzroll/mainwindow.cpp
```diff
@@ -352,8 +352,10 @@ using namespace mzUtils;
 	logWidget = new LogWidget(this, std::cout);
 	// rconsoleDockWidget = new RconsoleWidget(this);
 	spectralHitsDockWidget = new SpectralHitsDockWidget(this, "Spectral Hits");
-    peptideFragmentation = new PeptideFragmentationWidget(this);
-
+	peptideFragmentation = new PeptideFragmentationWidget(this);
+	alignmentErrorDialog=new QErrorMessage(this);
```
rish9511: Let's not create QErrorMessage unnecessarily. Only create it when we actually want to show the error to the user

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -320,7 +320,7 @@ void BackgroundPeakUpdate::run(void) {
                 quit();
                 return;
         }
-
+        connect(this,SIGNAL(alignmentError(QString)),mainwindow->alignmentErrorDialog,SLOT(showMessage(QString)));
```
rish9511: Why are we asking MainWindow to show an error message related to alignment ? Maybe alignment dialog should handle this. ? 
What do you think ?

bin/alignment.py
IpankajI: @rish9511 Qt handles all this- if python is failing in any case, it will not send complete data and Qt will try to make a json then it'll handle all kind of this error.

bin/alignment.py
IpankajI: @rish9511 No matter where python will fail, all thing will be handled in Qt.

src/gui/mzroll/background_peaks_update.cpp
IpankajI: @rish9511 background_peaks_update.cpp has all kind of job getting done in different thread from GUI thread and that(GUI) is handled by mainwindow.cpp. This architecture make whole ElMaven more responsive.

src/gui/mzroll/mainwindow.cpp
IpankajI: @rish9511 made this as local variable

src/gui/mzroll/alignmentvizwidget.cpp
```diff
@@ -130,17 +130,36 @@ void AlignmentVizWidget::drawMessageBox(PeakGroup newGroup, PeakGroup group) {
...
+float AlignmentVizWidget::calculateRsquare(PeakGroup newGroup,PeakGroup oldGroup) {
+
+    float SSres = 0;
+    float SStot=0;
+    float mean=0;
+
+    vector<mzSample*> oldSamples = getSamplesFromGroup(oldGroup);
+    vector<mzSample*> newSamples = getSamplesFromGroup(newGroup);
+    vector<float> newRts,oldRts;
+    for(int i=0;i<oldSamples.size();++i){
```
shubhra-agrawal: @IpankajI Please take care of code formatting. It is not consistent.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -586,69 +569,63 @@ void BackgroundPeakUpdate::align() {
...
                         // in such a case our rts wont update with new values
-                        jDoc = QJsonDocument::fromJson(data);
-                        if(!jDoc.isNull())
-                        parentObj = jDoc.object();
-
-                        if(!parentObj.isEmpty())
-                        aligner.updateRts(parentObj);
+                        jDoc = QJsonDocument::fromJson(processedDataFromPython);
+
+                        QString errorMessage=QString::number(processedDataFromPython.size());
```
shubhra-agrawal: error message creation should move into the else section, no?

tests/MavenTests/common.cpp
```diff
@@ -83,7 +83,7 @@ vector<PeakGroup> common::getGroupsFromProcessCompounds(){
     mavenparameters->clsf = clsf;
     mavenparameters->ionizationMode = -1;
     mavenparameters->matchRtFlag = true;
-    mavenparameters->compoundRTWindow = 2;
+    mavenparameters->compoundRTWindow = 1;
```
shubhra-agrawal: Cannot see this change reflected in the UI. default_settings.xml needs to be regenerated and pushed. @rish9511 will know more about this

rish9511: I was not sure what this change is actually doing and I am still not. 
In the above changes, Qt settings has been changed to hold '1' for compoundPPMWindow. Well, I did not bother about it as QtSettings no more hold PeakDetection Dialog settings  and here the change has been made in Tests module which is independent of UI.

Anyway, @IpankajI  if you want this to be reflected in the UI then you will have to make changes to default_settings.xml and settings.h .If you not sure about how to do that, ask me or @shubhra-agrawal

src/gui/mzroll/background_peaks_update.cpp
IpankajI: @shubhra-agrawal It's part of error message and common to two places.

src/gui/mzroll/alignmentvizwidget.cpp
```diff
@@ -130,17 +130,36 @@ void AlignmentVizWidget::drawMessageBox(PeakGroup newGroup, PeakGroup group) {
...
 
-        float diff = peak.rt - getRefRt(group);
-        r2 += pow(diff, 2);
+    for(int i=0;i<oldRts.size();++i){
+        mean+=oldRts[i];
     }
+    mean=1.0*mean/oldRts.size();
+    
+    for(int i=0;i<oldRts.size();++i){
+        SSres+=(newRts[i]-mean)*(newRts[i]-mean);
```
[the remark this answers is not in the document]
IpankajI: @shubhra-agrawal changed!

src/gui/mzroll/alignmentvizwidget.cpp
IpankajI: @shubhra-agrawal changed!

src/gui/mzroll/alignmentvizwidget.cpp
```diff
@@ -130,17 +130,36 @@ void AlignmentVizWidget::drawMessageBox(PeakGroup newGroup, PeakGroup group) {
...
+        SStot+=(oldRts[i]-mean)*(oldRts[i]-mean);
+    }
+
+    float RSquared=1-SSres/SStot;
```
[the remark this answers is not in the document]
shubhra-agrawal: @IpankajI

PR #563, opened 2018-01-25
UI: set Alignment w.r.t ExpectedRt visible only for Loess Fit

IpankajI:


shubhra-agrawal:
@IpankajI I think it would be better to move the checkbox to the last panel under Alignment algorithm as this is an algorithm specific option.

Issue #574, opened 2018-01-30
ElMaven not reloading last used settings (0.2.4 beta, Windows)


lparsons:
Restarting ElMaven does not reload the last used settings. This occurs using the 0.2.4 beta build on Windows. This does not appear to be a problem on my build of the develop branch on Linux.

rish9511:
@lparsons  thanks for reporting the issue.
We are looking into it

sp-eldata:
@lparsons Have you seen this on one machine or multiple?

lparsons:
@sp-eldata Since this is the beta version, we've only tested it on the one machine.

shubhra-agrawal:
@rish9511  Tried reproducing this in Windows 7 and 10. Last saved settings is working fine on Win10 for both installer and build versions. Win7 installer version is going back to defaults. Will check the build version too and let you know.

@lparsons @sp-eldata FYI

rish9511:
@lparsons  
when El-Maven shuts down it shows the path where the last ysed settings were saved. e.g `saving the settings at /path_to_folder/lastRun.xml` . Could you please check if the file is getting updated or not? 
This way we will know if it's the problem with loading them back or is it that we are not able to write in that location.

lparsons:
Actually, I'm running Linux, but @LiChenPU has a Windows machine that is exhibiting this issue. Li, can you check to see if the file is getting updated? If you need help figuring it out, let me know (or stop by my office) and we can sort it out. Thanks.

LiChenPU:
ElMaven reports it is attempting to save the settings in: `C:/Program Files (x86)/El-Maven/El-Maven 0.2.4 beta/bin\lastRun.xml`, which doesn't seem right. That file doesn't get created either.

lparsons:
Is there some way we can confirm what specific revision the build is from? Or better yet, is there a documented process to build and create the installer on Windows/MacOS?

rish9511:
@lparsons @LiChenPU 
Writing files in `C:/Program Files (x86)/El-Maven/El-Maven 0.2.4 beta/bin\lastRun.xml`  requires admin rights. You can temporarily solve this by running El-Maven in admin mode or installing it in different directory that does not require Admin rights . 

We have addressed this issues by storing `lastRun.xml` in user-config location. The next release will have this feature

rish9511:
@LiChenPU @lparsons 
Closing this as this has been fixed in v0.2.4

PR #578, opened 2018-02-01
Fix 0.2.4

rish9511:
Fix:
- usability issues (storing lastRun.xml in user-config location, updating mavenParameters on 
'esc' key event)
- issues that cause installer to fail to work properly

### DIFF
--- a/CrashReporter/s3-qt.pri
+++ b/CrashReporter/s3-qt.pri
@@ -1,5 +1,5 @@
 INCLUDEPATH += $$PWD/include
-QMAKE_CXXFLAGS += -Ofast -ffast-math -march=native -std=c++11
+QMAKE_CXXFLAGS += -Ofast -ffast-math  -std=c++11
 QT += network
 
 HEADERS += \
@@ -9,4 +9,3 @@ HEADERS += \
 SOURCES += \
     $$PWD/src/qs3/qs3.cpp \
     $$PWD/src/qs3/qbucket.cpp
-
--- a/src/core/libmaven/libmaven.pro
+++ b/src/core/libmaven/libmaven.pro
@@ -11,7 +11,7 @@ TEMPLATE=lib
 CONFIG += staticlib warn_off console silent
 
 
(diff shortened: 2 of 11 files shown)


lparsons:
Glad to see these issues being addressed, thanks @rish9511! Is the build process for Windows and/or MacOS documented somewhere? I would love to test some of these things out, but at the moment that part seems pretty opaque. I don't know how you guys create the installers and thus can't really test stuff like this.

rish9511:
@lparsons  
build process for windows and mac can be found [here](https://github.com/ElucidataInc/ElMaven#build) 
Since you want to build this branch, you will have to fetch and checkout to `fix_0.2.4` branch  after you have cloned the repo

 > I don't know how you guys create the installers and thus can't really test stuff like this.

This process has not been documented as of now

However If you don't want to go through the build process, we can provide you the installer for this specific branch. Whatever you prefer.

rish9511:
@sahil21  can be merged , right?

lparsons:
@rish9511 Good point about the build process. Assuming that is actually what is used to build the releases, that's great and is quite useful. However, without the installer, I can't really test this fix. I guess it's a manual thing right now? It would be nice to have some CI systems at least build executables, if not the installers. Is there an effort towards that goal? And thanks, looking forward to stable installer I can distribute to group (we've had to hold off getting the lab using the software due to serious issues like this one).

rish9511:
@lparsons 
> I guess it's a manual thing right now? 

Yes, we prepare installers manually.

>  It would be nice to have some CI systems at least build executables, if not the installers. Is there an effort towards that goal?

We want to do something similar in the coming time. As of now we use Travis for CI(automated build + tests) 

> And thanks, looking forward to stable installer I can distribute to group (we've had to hold off getting the lab using the software due to serious issues like this one).

Thanks for reporting all the issues  😃

lparsons:
Thanks, this looks like it addresses #574 then, though I'll need to wait for an installer to truly test.

Issue #553, opened 2018-01-19
Red dashed RT line does not get removed when selecing a compound that has no RT specified


lparsons:
Selecting a compound with no RT value (empty field in the CSV file), the RT line in the main window remains at the position it was drawn from the last compound with an RT value. Also, the position and zoon of the main window remains the same. Instead, the RT line should be removed and the main window should zoom to either the best peak group in the m/z window or zoom all the way out (not sure which is better, lean toward zoom out).

shubhra-agrawal:
@lparsons We've merged a fix for this to develop. When there is no expected Rt information, the Rt line is removed and the window is zoomed out.

There is a corner case here which I'm not able to cover right now. When I upload _MS2 data_, select a metabolite with _no_ expected Rt information, which has an _empty EIC_, the window goes completely blank. There is no title, plot axes etc. This is confusing because the user would not know if there really is no EIC or if there's a bug and they just can't see it. Seems to work for all other possibilities. 

I'm pushing this as a hot-fix for now. Not closing this issue till this corner case is covered.

lparsons:
@shubhra-agrawal I'm not seeing this fix in the latest develop build (`El_Maven_v0.2.4-beta-8-g85d15c8`).
![image](https://user-images.githubusercontent.com/645128/35739891-1e83c91a-0801-11e8-812e-922f93b922db.png)

shubhra-agrawal:
@lparsons There was a file version mix up on my end. Apologies for that.

I have implemented a fix that covers the above mentioned corner case. If there is no expected Rt, the EIC will zoom all the way out and the Rt line will be removed. Even in the case of empty EIC. This has been merged to develop.

PR #580, opened 2018-02-02
Fix: Remove Rt line when expectedRt is not provided #553

shubhra-agrawal:
Rt line of the last selected compound persists when a compound with no expected Rt is selected.

The old Rt line will now be removed and the complete EIC displayed.

Issue: #553

### DIFF
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -1436,6 +1436,13 @@ void EicWidget::setCompound(Compound* c) {
 		setFocusLine(c->expectedRt);
 		selectGroupNearRt(c->expectedRt);
 	}
+	else {
+		//remove previous focusline
+		if (_focusLine && _focusLine->scene()) {
+			scene()->removeItem(_focusLine);
+			resetZoom();
+		}
+	}
 
 //   clock_gettime(CLOCK_REALTIME, &tE);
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/eicwidget.cpp
```diff
@@ -1436,6 +1436,13 @@ void EicWidget::setCompound(Compound* c) {
 		setFocusLine(c->expectedRt);
 		selectGroupNearRt(c->expectedRt);
 	}
+	else {
+		//remove previous focusline
+		if (_focusLine && _focusLine->scene()) {
+			scene()->removeItem(_focusLine);
+			resetZoom();
```
rish9511: @shubhra-agrawal 
resetZoom() clears everything. attaching screenshots to explain what I mean

_without `resetZoom()`_
screen shot 2018-02-02 at 2 14 21 pm

_with `resetZoom()`_
screen shot 2018-02-02 at 2 17 25 pm

Is this intentional ?

shubhra-agrawal: @rish9511 I only tested this with MS data not MS2. Thanks for pointing this out!

shubhra-agrawal: @rish9511 This is one corner case where it's failing i.e. only when we're dealing with MS2 data and there is an empty EIC and no expected Rt information. Seems to work for all other cases. I've checked with all possible combinations of MS1/MS2 data, EIC/no EIC and with/without expectedRt but feel free to test it out once.

PR #582, opened 2018-02-02
Fix: Update peak dialog parameters when report isotopes is checked #453

shubhra-agrawal:


sp-eldata:
@shubhra-agrawal Just trying to understand. Why is this not a clean way? What would a clean solution look like?

shubhra-agrawal:
@sp-eldata All settings for the peak detection dialog should update right as the box is closed or "Find Peaks" is run. For some reason, 'report isotopes' is not taking the changed value with this approach. As a quick fix, I'm indirectly updating the settings as soon as report isotopes is clicked.

A clean solution would involve figuring out why the first approach isn't working and fixing that. @rish9511 and I tried debugging it but were unsuccessful. I have this in my todo list.

Issue #601, opened 2018-02-13
Isotope plot widget not showing all labels for a metabolite


ninjaturtl:
There are two issues here -

1. The isotope plot screen is blank when I click on a metabolite/label in the peak table.
![image](https://user-images.githubusercontent.com/31105033/36135627-714b131a-10b2-11e8-852a-320869426525.png)

2. I have to click on a peak to be able to see any isotopic label in the group. And even then I do not see all the labels for a metabolite in this plot.
![image](https://user-images.githubusercontent.com/31105033/36135670-ab006556-10b2-11e8-811a-f0341f5575bb.png)

PR #604, opened 2018-02-14
Fix: update display slot for istopebarplot #601

IpankajI:
Isotope bar plot should show up whenever we click on any group. I have changed slot to display isotope bar plot


chubukov:
Certainly this is a UI improvement. It's actually the way classic Maven behaved also. But in the earliest days of El-Maven development, we changed the behavior to only show the isotopes plot on clicking on a peak in order to prevent crashes that happened when the user would move too quickly between compounds in the compound widget or groups in the peak table.

Perhaps we have improved threading now and it's not an issue. But I would be careful here and try to test well.

IpankajI:
@chubukov We'll test it thoroughly before merging it to develop.

rish9511:
@IpankajI 
The isotope plots are shown only when we click on the parent group and not when we click on the isotope itself
I am not sure if that's the  desired behavior 

Was this intentional ?

PR #612, opened 2018-02-20
Fix test coverage

IpankajI:
Step to generate test coverage (only for linux)-

1. Run **uninstall.sh**
2. Run **run.sh** in debug mode (type **n** afte running run.sh)
3. Go to parent directory of ElMaven
4. Go to **coverage** directory and click on **index.html**
    While running in debug mode, all temporary files (moc* , *.o, *.gcno and *.gcda) will be created in tmp folder of respective subdirectory not in ../build/tmp folder. Running **uninstall.sh** will delete all these tmp directories.
    While running in release mode, all temporary files will be created in ../build/tmp folder.
Debug mode is generating temporary files in respective folder to generate test-coverage.

### DIFF
--- a/3rdparty/libcdfread/libcdfread.pro
+++ b/3rdparty/libcdfread/libcdfread.pro
@@ -1,8 +1,8 @@
-include($$mzroll_pri)
+
 
 MOC_DIR=$$top_builddir/tmp/cdfread/
 OBJECTS_DIR=$$top_builddir/tmp/cdfread/
-
+include($$mzroll_pri)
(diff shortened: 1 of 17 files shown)


sp-eldata:
@IpankajI Out of curiosity: what kind of test coverage will we get? Do you have an example that you can share here?

IpankajI:
@sp-eldata First row in screen-shot explaining that Compound.cpp file is used 100% while running test cases- 
testcoverage

After clicking on Compound.cpp, we can get which line is getting executed for how many number of times-
compound

PR #614, opened 2018-02-20
Refactor isotope detection

sahil21:
- Add new class Isotope Detection
- Remove duplicate function pullIsotopesBarPlot
- Refactored pullIsotopes function

PS. This is the first phase of refactoring of pullIsotopes function. It still needs a lot of refactoring.


src/core/libmaven/isotopeDetection.cpp
```diff
@@ -432,10 +432,10 @@ void IsotopeDetection::addIsotopes(PeakGroup* parentgroup,
         PeakGroup& child = (*itr2).second;
         child.metaGroupId = parentgroup->metaGroupId;
 
-        bool isotopeAdded = addIsotopes(parentgroup, child, isotopeName, C13Flag, N15Flag, S34Flag, D2Flag);
+        childStatistics(parentgroup, child, isotopeName);
+        bool isotopeAdded = filterLabel(isotopeName, C13Flag, N15Flag, S34Flag, D2Flag);
```
shubhra-agrawal: filterLabel returns true if the isotope should not be added to parent. Maybe change 'isotopeAdded' to 'filterIsotope'? and continue if filterIsotope is true.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -637,10 +624,61 @@ void BackgroundPeakUpdate::setRunFunction(QString functionName) {
...
+                mavenParameters,
+                isoType,
+                C13Flag,
+                N15Flag,
+                S34Flag,
+                D2Flag);
+	isotopeDetection.pullIsotopes(parentgroup);
 }
+
+void BackgroundPeakUpdate::pullIsotopesIsoWidget(PeakGroup* parentgroup) {
```
rish9511: same as pull isotopes(),no ? the only difference is in IsotopeDetectionType

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -637,10 +624,61 @@ void BackgroundPeakUpdate::setRunFunction(QString functionName) {
...
+                mavenParameters,
+                isoType,
+                C13Flag,
+                N15Flag,
+                S34Flag,
+                D2Flag);
+	isotopeDetection.pullIsotopes(parentgroup);
+}
+
 void BackgroundPeakUpdate::pullIsotopesBarPlot(PeakGroup* parentgroup) {
```
rish9511: again, same as pullIstotpes() exception for the IsotopeDetectionType.

src/gui/mzroll/background_peaks_update.cpp
shubhra-agrawal: @rish9511 It's calling the same isotopeDetection::pullIsotopes() but with different flags and isoType

rish9511: @shubhra-agrawal  Just edited my comment

sahil21: @rish9511 @shubhra-agrawal Ideally, we won't be passing any IsoType to the class. 
There will be only one pullIsotopes function and it will only update one vector of isotopic children (Here we are updating 3 vectors depending on the widget used i.e children, childrenIsoWidget & childrenBarPlot). 

**Why we are updating 3 vectors separately?**
Currently, we are passing a pointer to the class PeakGroup. So updating one vector (children in this case) can lead to errors because 3 widgets will try to update the same vector.

**Solution**
Pass a copy of PeakGroup to every widget rather than passing the pointer.

rish9511: @sahil let's leave it for now then

Issue #368, opened 2017-09-10
m+n isotopes misleading


chubukov:
The way the "max number of m+n isotopes" is implemented is really misleading if you have more than one element (e.g. both 13C and 15N) enabled.

It iterates over 13C first and then over 15N, and it stops when it reaches n. So if my molecule has formula C5N4, and I set n=10, I will get something like

C1N0,
C1N1,
....
C2N4 (maybe I'm off by one (it looks like I always get the N0 ones), but the point is I never get to C5N2).

We've discussed this feature a few times -- I don't really think it's necessary, but it's ok if it's implemented well. In this case it actually caused some serious misinterpretation of an experiment.

sahil21:
@chubukov Yes, I think this feature is not necessary. Also, we should try to avoid having lots of redundant parameters in the UI. The UI is already very cluttered.
This feature was implemented to take care of performance issues with isotopic detection. I believe there must be some other way to improve performance.

chubukov:
@sahil21 here's a clear performance enhancement:

currently, for every isotope, the EIC gets pulled twice. Once to check the correlation with the parent, and then again to save it. And the parent EIC could also be saved, so that it doesn't have to be re-pulled to check the correlation with every isotope.

sahil21:
@chubukov We are refactoring isotope detection part in El-Maven. It will be a separate class and not part of PeakDetector.cpp. We'll keep this in mind while refactoring.

shubhra-agrawal:
@chubukov @sahil21 This option has been removed now. Considering that the default value had already been changed to 10000, this parameter was pretty much redundant. 

As for the performance enhancement, we are currently refactoring the isotope detection module and trying to reduce the instances where the same calculation is being done multiple times.

PR #618, opened 2018-02-22
Remove m+ n isotopes from Options dialog

shubhra-agrawal:


lparsons:
I believe this addresses #368.  As a general comment, it would be helpful (esp to those of us who are more passive observers) to reference the issue(s) addressed by PRs.

shubhra-agrawal:
@lparsons We generally tag commits with the issue number. But yes, PR description should have a list of issues addressed as well. Thanks for pointing this out.

PR #619, opened 2018-02-23
[Ref] getExpectedMz

sahil21:


shubhra-agrawal:
Found this while testing the fix_bookmarking branch.

On loading mzroll, isotopic children have expectedMz = 0 as isotopeMass is not calculated. This is resulting in empty EICs for isotopic children in mzroll.

sahil21:
@shubhra-agrawal Raised a PR (#630) to fix the expected mz issue with isotopes in mzroll.

PR #622, opened 2018-02-25
[Fix] Isotopic peak detection in Shift-drag & Bookmarking

sahil21:


shubhra-agrawal:
@sahil21 Isotopic children are giving blank EICs when loading an mzroll. This isn't a problem in v0.2.4 but I'm not sure where it went wrong.

PR #628, opened 2018-03-05
Ref csv report

IpankajI:
CSV exporting is a four step process now -
      1. Create one instance of this class by giving required value of
          all parameters in constructor.      
      2. Add all groups one by one to be exported by calling addItem method of this class
      3. call exportGroup() method of this class which will return true if report is successful
         otherwise false
      4. If returned value from exportGroup() is false call getErrorReport method to get error message

Qt-code is also removed from csvreports.cpp and csvreports.h!

Update:
     step 2 and 3 can be replaced with one method **exportGroup(vector<PeakGroup\*> &groupsToExport)**


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -665,35 +665,31 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
     }
 
-    if(fileName.empty()) return;
-    
-    if (mavenParameters->samples.size() == 0) return;
+    vector<mzSample*> samples = mavenParameters->samples;
+    if ( samples.size() == 0) return;
 
-    csvreports->setUserQuantType(quantitationType);
+    MavenParameters* mp = mavenParameters;
```
sahil21: @IpankajI Why are we making a new variable mp here?

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -665,35 +665,31 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
+    PeakGroup::QType t = quantitationType;
+    CSVReports::ExportType exportType = CSVReports::GroupExport;
+    int selectionFlag = -1;
+    bool includeSetNamesLine = false;
 
-    //Added to pass into csvreports file when merged with Maven776 - Kiran
-    //CLI exports the default Group Summary Matrix Format (without set Names)
-    csvreports->openGroupReport(fileName);
+    CSVReports* csvExport = new CSVReports(samples, mp, t, fileName, exportType, selectionFlag, includeSetNamesLine);
```
sahil21: @IpankajI Line too big.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -665,35 +665,31 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
 
     for(int i=0; i<mavenParameters->allgroups.size(); i++ ) {
 		PeakGroup& group = mavenParameters->allgroups[i];
-		csvreports->addGroup(&group);
+		csvExport->addItem(&group);
```
sahil21: @IpankajI Don't you think addItem is a redundant step? In my opinion, we should only have exportGroups function which can take a vector of groups or a single group and just exports the csv?

IpankajI: @sahil21 Changing!

src/cli/peakdetector/PeakDetectorCLI.cpp
IpankajI: @sahil21 breaking

src/cli/peakdetector/PeakDetectorCLI.cpp
IpankajI: just for consistency and better readability of code, anyway it's a local pointer variable so there won't be any memory hit.

src/cli/peakdetector/PeakDetectorCLI.cpp
IpankajI: @sahil21 Sometimes we are exporting filtered groups and we add one group at a time based on filtering criteria, So I'm letting it be as it is.

src/core/libmaven/csvreports.cpp
```diff
@@ -26,6 +47,88 @@ CSVReports::~CSVReports() {
...
+                }      
+            }
+        }
+
+        outFileStream.close();
+    }
+    
+}
+
+void CSVReports::addColumnNames(){
```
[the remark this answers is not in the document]
IpankajI: @shubhra-agrawal It's good practice to not add any redundancy in code. It's just a **if** statement that is controlling otherwise you have to write a whole new method that will add extra load to compiler and linker.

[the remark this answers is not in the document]
IpankajI: @shubhra-agrawal One more point- all columns are getting added at one place so it is more easy to understand the code base. It we add columns at different places, then it will definitely make one to grasp logics behind the columns- What will be the columns name at what conditions.

src/core/libmaven/csvreports.h
```diff
@@ -22,122 +20,65 @@ class CSVReports {
...
+     * all parameters in below constructor.      
+     * 2. Add all groups one by one to be exported by calling addItem method of this class
+     * 3. call exportGroup() method of this class which will return true if report is successful
+     *  otherwise false
+     * 4. If returned value from exportGroup() is false call getErrorReport method to get error message
+     */ 
     /**
-    *@brief-    destructor, just close all open output files opened for writing csv or tab file
-    */
+     * Note: don't use same instance of this class to export more that one csv file.
```
rish9511: To me it does not make sense to have multiple instances of CsvReports. 
Only one instance of CsvReports should be sufficient.

src/core/libmaven/csvreports.h
```diff
@@ -22,122 +20,65 @@ class CSVReports {
...
+     * Otherwise, it'll export mixed groups.
```
rish9511: Why will it have mixed groups?

src/core/libmaven/csvreports.h
```diff
@@ -22,122 +20,65 @@ class CSVReports {
...
-
+    void addColumnNames();
 
     int groupId;	/**@param-  incremental group numbering. Increment by 1 when a group is added for csv report  */
     string SEP;     /**@param-  separator in output file*/
 
-    QString errorReport;    /**@param-  error message, TODO- QString should not be in libmaven folder, only standard C++ statement should be here*/
+    string errorReport;    /**@param-  error message, TODO- QString should not be in libmaven folder, only standard C++ statement should be here*/
 
     vector<mzSample*> samples;      /**@param-  pointers to all mz samples uploaded*/
```
rish9511: @IpankajI 
Can we not get rid of this pointer ? 
We can send this pointer whenever needed. e.g   exportGroup can receive  a pointer to mzSamples

src/core/libmaven/csvreports.h
```diff
@@ -22,122 +20,65 @@ class CSVReports {
...
     PeakGroup::QType qtype;             /**@param-  user quant type, represents intensity of peaks*/
     MavenParameters * mavenparameters;
-    int selectionFlag;      /**@param-  TODO*/
+    ofstream outFileStream;
+    string _fileName;
+    ExportType _exportType; /**@param- specify either you want to export group info or peak info, check enum ExportType */
+    vector<PeakGroup*> groups;
```
rish9511: again, is this needed?  why should csv reports have a vector of groups ?
if we get rid of this, we won't even need multiple instances of CsvReports and hence there won't be any mixed groups .

src/core/libmaven/csvreports.cpp
```diff
@@ -1,283 +1,150 @@
...
-             groupReport << endl;
-         }
+    if(groups.size() == 0){
+        errorReport = "No groups found!";
+        return 0;
     }
-    else {
-        errorReport = "Unable to write to file \"" + QString::fromStdString(outputfile) + "\"\n";
-        errorReport += "Please check if you have permission to write to the specified location or the file is not in use";
+    if(_fileName.length() < 4){
```
rish9511: So,if I name my file as "foo", it won't get saved ?
Why do we want to force the users to name their file name with more than 3 characters?

src/core/libmaven/csvreports.cpp
```diff
@@ -1,283 +1,150 @@
...
-      insertGroupInformationIntoCSVFile(group);
-
-}
-
-void CSVReports::insertPeakInformationIntoCSVFile(PeakGroup* group) {
-
-      writePeakInfo(group);
+    else{
+        addColumnNames();
+        for(int i=0 ; i<groups.size(); ++i){
```
rish9511: Let's start using [range based for loop](http://en.cppreference.com/w/cpp/language/range-for) 

Same for every other for loop in CsvReports. We can ignore the rest of the classes for now

src/core/libmaven/csvreports.h
chubukov: @rish9511 Does it not make sense to have multiple instances to deal with potential multithreading? E.g. the user could initiate two exports at the same time.

rish9511: @chubukov  having multiple instances is not always the way to get away from multi threading issues. We can use locks 
and IIRC export functionality is not multi threaded

chubukov: @rish ok sure

rish9511: @chubukov you tagged the wrong person :grin:

src/core/libmaven/mzUtils.cpp
```diff
@@ -6,6 +6,24 @@
  */
 namespace mzUtils {
 
+    string replaceAll(string thisOne, string fromThis, string withThis){
```
rish9511: why not use the [standard function](http://en.cppreference.com/w/cpp/algorithm/replace) ?

src/core/libmaven/csvreports.h
IpankajI: @rish9511 @chubukov  We can use same instance, if we can add a reset function to reset all parameters or required parameters including variable holding groups. But this will introduce chances of erroneous result in csv report and overhead on writing code for using that method. Currently, we don't have many requirement to use this class, that's why I didn't write that function. If need be, we'll write that one, however I'll not recommend to use same instance to export multiple csv file to avoid error in result.

src/core/libmaven/csvreports.h
IpankajI: @rish9511 It can have groups from earlier export, because current implementation does not have method to clear groups from that variable.

src/core/libmaven/csvreports.cpp
IpankajI: @rish9511 last three characters are of file extension (like .csv, .tab) not of actual file name.

src/core/libmaven/csvreports.cpp
IpankajI: @sahil21 @shubhra-agrawal what's your opinion here- shifting our traditional **for loop** to recommended by @rish9511

src/core/libmaven/csvreports.h
rish9511: @IpankajI  my point was that we should have groups from earlier export at all.
CsVReports should be provided the groups that have to be written to CSV.
Why should CsVReports contain a vector of groups in first place ? because if we do that, then there is overhead of clearing groups every time we are done exporting.

IpankajI: @rish9511 There are places in Qt-code where we need to export filtered groups which are added one at a time based on filtering criteria. Again, if we are adding group one at a time and writing that group directly to file then file stream has to be remained open for long time which can result in file write exception.

src/cli/peakdetector/PeakDetectorCLI.cpp
sahil21: @IpankajI Then exportGroups function should be able to take PeakGroup or vector of PeakGroup as input, right? I'm talking about function overloading.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -663,35 +663,32 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
     }
-    csvreports->closeFiles();
-
-    if (csvreports->getErrorReport() != "") {
-        cout << endl << "Writing to CSV Failed : " << csvreports->getErrorReport().toStdString() << endl;
+    if (csvExport->exportGroup() == false ) {
```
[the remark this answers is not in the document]
IpankajI: @sahil21 that's what intended here.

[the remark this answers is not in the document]
IpankajI: @sahil21  cool, adding separate variable!

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -663,35 +663,32 @@ void PeakDetectorCLI::saveCSV(string setName) {
     #endif
 
 	string fileName = mavenParameters->outputdir + setName + ".csv";
-
-    CSVReports* csvreports = new CSVReports(mavenParameters->samples);
-    csvreports->setMavenParameters(mavenParameters);
-
+	if(fileName.empty()) return;
+
```
shubhra-agrawal: fileName would never be empty. Use this instead:
if (mavenParameters->outputdir.empty()) return;
string fileName = mavenParameters->outputdir + setName + ".csv";

src/cli/peakdetector/PeakDetectorCLI.cpp
shubhra-agrawal: @IpankajI Now there are two variables that can be used for the same thing. mp and mavenParameters. Future contributors could use either of them. It will be more consistent and readable if we use mavenParameters.

src/cli/peakdetector/PeakDetectorCLI.cpp
sahil21: @IpankajI Why do we have to make temporary vector? We can pass mavenParameters->allgroups, right?

src/core/libmaven/csvreports.cpp
```diff
@@ -1,283 +1,155 @@
...
-        QStringList peakReportcolnames;
-        peakReportcolnames << "groupId" << "compound" << "compoundId" << "formula" << "sample" << "peakMz"
-                << "medianMz" << "baseMz" << "rt" << "rtmin" << "rtmax" << "quality"
-                << "peakIntensity" << "peakArea" << "peakSplineArea" << "peakAreaTop"
-                << "peakAreaCorrected" << "peakAreaTopCorrected"
-                << "noNoiseObs" << "signalBaseLineRatio"
-                << "fromBlankSample";
-        QString header = peakReportcolnames.join(SEP.c_str());
-        peakReport << header.toStdString() << endl;
+    string fileFormat = _fileName.substr(_fileName.length()-4 , 4 );
```
shubhra-agrawal: @IpankajI Just a suggestion, can we set a delimiter type when user chooses which format they want and just use it everywhere we want? It would save on all the string matching etc.

src/core/libmaven/csvreports.cpp
shubhra-agrawal: @IpankajI It's not necessary but it would make for cleaner code with fewer iteration variables. Your call, I guess.

IpankajI:
@sahil21 @shubhra-agrawal @rish9511 **exportGroup** is overloaded, read updated PR message.
Also same instance of this class can be used to export multiple csv file with same controlling params passed in contructor.

PR #629, opened 2018-03-05
Elmaven polly integration

Ky5595:
Please suggest improvements wherever required..


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2509,6 +2518,14 @@ void MainWindow::createToolBars() {
 	ionizationModeLabel->setFrameShape(QFrame::Panel);
 	ionizationModeLabel->setFrameShadow(QFrame::Raised);
 	
+	check_polly_login = new QLabel(hBox);
+	check_polly_login->setToolTip("Status of connection to Polly");
```
IpankajI: use short tool tip message something like - Polly Connection.

src/gui/mzroll/pollyintegration.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+    qDebug()<<"exiting PollyIntegration now....";
+}
+
+QByteArray PollyIntegration::run_qt_process(QString command){
+    QProcess process;
+    process.start(command);
+    process.waitForFinished();
+    QByteArray result = process.readAllStandardOutput();
+    QByteArray result2 = process.readAllStandardError();
```
IpankajI: What's the use of **result2**

src/gui/mzroll/pollyintegration.cpp
```diff
@@ -0,0 +1,263 @@
...
+    return result;
+}
+
+QByteArray PollyIntegration::run_system_process(QString command){
+    QByteArray ba = command.toLatin1();
+    const char *c_str2 = ba.data();
+    system(c_str2);
+    return "";
```
IpankajI: Is there any reason to return **""** always, we can simply ignore return statement?

src/gui/mzroll/pollyintegration.cpp
```diff
@@ -0,0 +1,263 @@
...
+QString PollyIntegration::get_run_id(QByteArray result){
+    QList<QByteArray> test_list = result.split('\n');
+    int size = test_list.size();
+    QByteArray result2 = test_list[size-2];
+    QJsonDocument doc(QJsonDocument::fromJson(result2));
+    // Get JSON object
+    QJsonObject json = doc.object();
+    QVariantMap json_map = json.toVariantMap();
+    QString run_id =  json_map["id"].toString();
+    qDebug() << run_id;
```
IpankajI: There won't be any benefit for printing just a string(run_id) until we print any descriptive word.

src/gui/mzroll/pollyintegration.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+
+void PollyIntegration::login_user(){
+    LoginForm *w=new LoginForm(_tableDockWidget->_mainwindow,this);
+    w->setModal(true);
+    w->show();
+}
+
+void PollyIntegration::get_project_name(){
+    ProjectForm *popup=new ProjectForm(_tableDockWidget->_mainwindow,this);
```
IpankajI: This object is not getting destructed, it's source of memory leak- handle it

src/gui/mzroll/pollyintegration.cpp
```diff
@@ -0,0 +1,263 @@
...
+    else {
+        qDebug("user is already logged in... sending data to polly now..");
+        get_project_name();
+    }
+}
+
+
+
+void PollyIntegration::login_user(){
+    LoginForm *w=new LoginForm(_tableDockWidget->_mainwindow,this);
```
IpankajI: Handle memory leak here.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -779,6 +788,63 @@ void TableDockWidget::exportGroupsToSpreadsheet() {
     }
 }
 
+
+
+
+void TableDockWidget::exportGroupsToSpreadsheet_polly() {
```
IpankajI: As far as I can see, I don't think it's a required function, instead **exportGroupsToSpreadsheet** can be used!

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -812,6 +878,13 @@ void TableDockWidget::exportJson() {
 }
 
 
+void TableDockWidget::exportPolly() {
+    LOGD;
+    _pollyIntegration = new PollyIntegration(this);
```
IpankajI: Object is not handle properly.

src/gui/mzroll/pollyintegration.cpp
Ky5595: In case of process Error we can have a look at this variable..
This is not being used in the function as of Now, but we should keep it there for future references as we want to terminate the uploading process in case of Process error..

src/gui/mzroll/mainwindow.cpp
Ky5595: Changing it to Polly connection status..

src/gui/mzroll/pollyintegration.cpp
Ky5595: All these "system" functions are not being used as of now, We are using Qprocess.
But in some cases where QProcess fails, We may use these system functions ... This is a work in progress so I just returned empty string as of now..

src/gui/mzroll/pollyintegration.cpp
Ky5595: Removing the Qdebug statement..

src/gui/mzroll/pollyintegration.cpp
Ky5595: took care of this issue

src/gui/mzroll/pollyintegration.cpp
Ky5595: done

src/gui/mzroll/tabledockwidget.cpp
Ky5595: No..This one is quite different..

src/gui/mzroll/tabledockwidget.cpp
Ky5595: What do you mean by "not handled properly" ??
Please be more specific

Ky5595:
@IpankajI  @rish9511  @sahil21 Updated the code..please review again..

PR #630, opened 2018-03-06
[Fix] Mzroll loading: Expected mz is zero for isotopes #619

sahil21:
In PR #619, expectedMz variable was introduced in the custom defined datatype "PeakGroup". This datatype was not added to write or read mzroll.


sahil21:
@shubhra-agrawal Backward compatibility of mzroll has not been implemented in this PR.

shubhra-agrawal:
@sahil21 This looks good. We need to add backward compatibility before the release though.

PR #633, opened 2018-03-09
Ref isotope plot

IpankajI:
Now isotopeplot has a define interface, see code doc in .h file to know how to use this class for isotope bar plotting. Strong dependencies on other class have been also been removed.

### DIFF
--- a/src/gui/mzroll/isotopeplot.cpp
+++ b/src/gui/mzroll/isotopeplot.cpp
@@ -97,7 +97,7 @@ void IsotopePlot::showBars() {
     PeakGroup::QType qtype = PeakGroup::AreaTop;
     if ( _mw ) qtype = _mw->getUserQuantType();
 
-    MatrixXf MM = _mw->getIsotopicMatrix(_group);
+    MatrixXf MM = getIsotopicMatrix(_group);
     setBelowAbThresholdMatrixEntries(MM,_mw);
     normalizeIsotopicMatrix(MM);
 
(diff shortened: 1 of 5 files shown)


src/gui/mzroll/isotopeplot.cpp
```diff
@@ -1,27 +1,24 @@
 #include "isotopeplot.h"
 
-// import most common Eigen types
-//USING_PART_OF_NAMESPACE_EIGEN
 using namespace Eigen;
 
+IsotopePlot::IsotopePlot(QCustomPlot* customPlot, float width, float height, float stackingZValue,
```
sahil21: @IpankajI Do we need to pass customPlot here? Can't we make customPlot as private member of IsotopePlot class?

sahil21: @IpankajI Also, why do we need to pass width and height even?

src/gui/mzroll/isotopeplot.cpp
```diff
@@ -1,27 +1,24 @@
 #include "isotopeplot.h"
 
-// import most common Eigen types
-//USING_PART_OF_NAMESPACE_EIGEN
 using namespace Eigen;
 
+IsotopePlot::IsotopePlot(QCustomPlot* customPlot, float width, float height, float stackingZValue, 
+                vector<mzSample*> samples, float abundanceThresold, PeakGroup::QType qtype){
```
sahil21: @IpankajI Don't you think every argument should not be part of the constructor?

src/gui/mzroll/isotopeplot.cpp
IpankajI: customPlot is used to hold bars and manage these, but this needs to be placed somewhere else in DockWidget which in turn has to be placed in a GUI element, not as independent entity. So it has to be public and it's height and width has to be specified based on holding GUI element.

src/gui/mzroll/isotopeplot.cpp
IpankajI: Right but here it's all parameter have to be  passed, as these need to be accessed at many places.  We can pass parameters as method arguments if these are only used by that method only otherwise pass these at time of creating instance of that class.

src/gui/mzroll/isotopeplot.cpp
IpankajI: @sahil21 done!

PR #636, opened 2018-03-22
Refactor get isotope

shubhra-agrawal:
Abundance error and correlation check are done both in isotopeDetection and isotopeWidget. Moving the code block into a separate function to get rid of duplicate code later.


src/core/libmaven/isotopeDetection.h
```diff
@@ -26,6 +26,7 @@ class IsotopeDetection
 		bool D2Flag);
 
 	void pullIsotopes(PeakGroup *group);
+	bool filterIsotope(Isotope x, bool C13Flag, bool N15Flag, bool S34Flag, bool D2Flag, float parentPeakIntensity, float isotopePeakIntensity, mzSample* sample, PeakGroup* parentGroup = NULL);
```
IpankajI: break this line with no. of character not more than 80

shubhra-agrawal: @IpankajI Changed in future branches. Most of the parameters have been removed as well.

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -202,6 +164,58 @@ map<string, PeakGroup> IsotopeDetection::getIsotopes(PeakGroup* parentgroup, vec
     return isotopes;
 }
 
+bool IsotopeDetection::filterIsotope(Isotope x, bool C13Flag, bool N15Flag, bool S34Flag, bool D2Flag, float parentPeakIntensity, float isotopePeakIntensity, mzSample* sample, PeakGroup *parentGroup)
```
IpankajI: @shubhra-agrawal Break all lines with no. of character not more than 80

shubhra-agrawal: @IpankajI Done in future branches.

PR #637, opened 2018-03-22
Refactor isotope widget

shubhra-agrawal:
- Unify duplicate code
- Move logical functions to isotopeDetection (part of core). IsoWidget should only have UI functionalities eventually
- Remove redundant functions


src/core/libmaven/isotopeDetection.cpp
```diff
@@ -164,58 +152,83 @@ map<string, PeakGroup> IsotopeDetection::getIsotopes(PeakGroup* parentgroup, vec
...
     return false;
 }
 
+std::pair<float, float> IsotopeDetection::getIntensity(Scan* scan, float mzmin, float mzmax)
+{
+    float highestIntensity = 0;
+    float rt = 0;
+    mzSample* sample = scan->getSample();
+    //TODO: use maxIsotopeScanDiff instead of arbitrary number
+    for (int i = scan->scannum - 2; i < scan->scannum + 2; i++) {
```
rish9511: @shubhra-agrawal 
it's not clear as to why we are starting from scan number - 2 and end at scan number + 2. 
Document this in the comments, no  ?

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -164,58 +152,83 @@ map<string, PeakGroup> IsotopeDetection::getIsotopes(PeakGroup* parentgroup, vec
...
+		Scan* s = sample->getScan(i);
+		vector<int> matches = s->findMatchingMzs(mzmin, mzmax);
+		for (unsigned int i = 0; i < matches.size(); i++) {
```
rish9511: @shubhra-agrawal 
just a suggestion. we can use [range based for](http://en.cppreference.com/w/cpp/language/range-for) here

src/core/libmaven/isotopeDetection.cpp
shubhra-agrawal: @rish9511 That's old code. To make sure isotopic peak is within 2 scans of the parent. We have a variable called maxIsotopeScanDiff which should be used here instead but it was messing up the results a bit so I changed it back. I've added a TODO and will take care of this when possible. 
For now, I'll add a comment if you think that's useful.

src/gui/mzroll/isotopeswidget.cpp
```diff
@@ -166,9 +166,19 @@ void IsotopeWidget::computeIsotopes(string f) {
...
-		if (isotopeDetector->filterIsotope(x, isotopePeakIntensity, parentPeakIntensity, sample))
-			continue;
+		if (isotopeParameters->_group)
+		{
+			filterIsotope = isotopeDetector->filterIsotope(x, isotopePeakIntensity, parentPeakIntensity, sample, isotopeParameters->_group);
+		}
+		else
+		{
+			filterIsotope = isotopeDetector->filterIsotope(x, isotopePeakIntensity, parentPeakIntensity, sample);
+		}
```
rish9511: this can be converted to one line of code. 
just use it as this:
`filterIsotope = isotopeDetector->filterIsotope(x, isotopePeakIntensity, parentPeakIntensity, sample, isotopeParameters->_group);}`

and make sure that isotopeParameters->_group: 
-  is initialized to null in `IsotopeLogic` 
-  is only used in `filterIsotope` when it's not null

src/core/libmaven/isotopeDetection.cpp
rish9511: @shubhra-agrawal 
let's leave it then

src/core/libmaven/isotopeDetection.cpp
shubhra-agrawal: Yep. Makes sense. I'll add it.

src/gui/mzroll/isotopeswidget.cpp
shubhra-agrawal: My bad. Missed this somehow. I've pushed the requested changes; you can review this again.

src/gui/mzroll/point.cpp
```diff
@@ -144,6 +144,11 @@ void EicPoint::mousePressEvent (QGraphicsSceneMouseEvent* event) {
 
     if ( _group && _group->isIsotope() == false ) {
         _mw->isotopeWidget->updateIsotopicBarplot(_group);
+        _mw->isotopeWidget->setPeakGroupAndMore(_group);
```
sahil21: @shubhra-agrawal Can you document this part? Why this else condition?

shubhra-agrawal: @sahil21 done.

PR #638, opened 2018-03-23
[Add] Append Sample number to Sample Name for mzML files

sahil21:


rish9511:
@sahil21 
I did some quick testing and found that this branch is crashing

Steps to reproduce(tested with mzML files)
1. upload some mzMl files 
2. find peaks (not sure if this is important )
3. delete all the samples
4. upload the samples again

Back trace report: https://pastebin.com/xXBqJbeS

Also, after deleting the samples the EIC widget looks weird. 
This is also the case with develop though 
sam_del

sahil21:
@rish9511 If it reproducible on develop then we should file it as a separate issue, right?

rish9511:
@sahil21 
The crash is not reproducible on Develop.
The issue in Eic Widget can be seen both in your branch and Develop

sahil21:
@rish9511 I can reproduce the same crash on Develop.

PR #646, opened 2018-03-30
Init exception

rish9511:
The change introduced in the four commits  are the following:

commit 42157ef (first commit)
- new Error handling library 
- new exception class
- introduce exception while importing files

commit 5e147d4
- introduce exceptions in mzSample class
- handle error while parsing different sample formats

commit 48a9f9d
- introduce two new exception classes.
- Don't bother about this commit. We decided to have only one exception class and hence
  these classes were removed 

commit c2d351b
- use a single exception class
- do all the logging in exception class
- store logs in generic config location

commit ce8762d
- link Error handling library with Cli and Tests

### DIFF
--- a/3rdparty/ErrorHandling/MavenException.cpp
+++ b/3rdparty/ErrorHandling/MavenException.cpp
@@ -1,0 +1,42 @@
+#include "MavenException.h"
+#include <Log.h>
+
+ErrorMsg* ErrorMsg::errMsg = nullptr;
+
+
+ErrorMsg::ErrorMsg()
+{
+    errMessages[ErrorCodes::FileNotFound] = "File not found";
+    errMessages[ErrorCodes::Blank] =  "Blank file";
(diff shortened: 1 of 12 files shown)


3rdparty/ErrorHandling/MavenException.cpp
```diff
@@ -0,0 +1,42 @@
+#include "MavenException.h"
+#include <Log.h>
```
sahil21: Why not in header?

src/gui/mzroll/mzfileio.cpp
```diff
@@ -658,7 +681,8 @@ void mzFileIO::readThermoRawFileImport() {
 
 void mzFileIO::addFileToQueue(QString f)
 {
-    if (isKnownFileType(f)) filelist << f;
+//    if (isKnownFileType(f)) filelist << f;
```
sahil21: Why has this been removed?

Issue #493, opened 2017-12-15
Expected mz in Peak Table shows NA 


shubhra-agrawal:
Current develop

I have only observed this for MRM data for now. Sometimes on bookmarking or automated peak detection, peaks table shows NA value in the Expected Mz column.

saifulbkhan:
Not encountered this before. Will reopen if found again.

PR #649, opened 2018-04-02
[Fix] Stop mass adjustment if mass is zero #493

shubhra-agrawal:
Compound mass is set to zero for MS/MS fragments. It is used as a check for MS2 in multiple places.
AdjustMass was adding or removing proton mass even when mass is zero. This caused some of these checks to fail.
One prominent result: expectedMz showing "NA" in Peaks Table

Issues: #493


rish9511:
@shubhra-agrawal 
I have tested this with Mimosa datasets on two systems but it does not seem to work for me.
Please check the screen shot attached. 
Can you confirm that the commit id is correct ?

Did you test it on linux or windows ?

image

PR #652, opened 2018-04-03
clean up Error Message

rish9511:
Major highlight
- Move the error message class to separate file


3rdparty/3rdparty.pro
```diff
@@ -2,7 +2,7 @@ TEMPLATE = subdirs
 CONFIG += ordered qt thread
 
 
-SUBDIRS += libneural libcdfread  libcsvparser pugixml/src  libpillow libpls
+SUBDIRS += libneural libcdfread  libcsvparser pugixml/src  libpillow libpls ErrorHandling
```
sahil21: @rish9511 Let's move ErrorHandling library out of 3rd party? Can you add this to the TODO?

3rdparty/ErrorHandling/errorcodes.h
```diff
@@ -0,0 +1,29 @@
...
+        ParsemzXml,
+        ParsemzMl,
+        ParsemzData,
+        ParseCsv
+    };
+
+};
+
+//singleton
+class ErrorMsg
```
sahil21: @rish9511 Why do we need to have ErrorCodes class along with MavenException class? Is there a specific reason for both of them being separate? can we combine them?

sahil21: @rish9511 Class name and file should be same.

src/core/libmaven/mzSample.cpp
```diff
@@ -155,7 +160,16 @@ void mzSample::loadSample(const char *filename)
 {
 
 	//Loading and Decoding the file
-	loadAnySample(filename);
+    //catch any error while parsing
+    try {
```
sahil21: @rish9511 Where's the case where no exception has been caught but loading sample failed?

src/gui/mzroll/mzfileio.cpp
```diff
@@ -658,7 +681,8 @@ void mzFileIO::readThermoRawFileImport() {
 
 void mzFileIO::addFileToQueue(QString f)
 {
-    if (isKnownFileType(f)) filelist << f;
+//    if (isKnownFileType(f)) filelist << f;
```
shubhra-agrawal: @rish9511 why are we removing this check?

3rdparty/3rdparty.pro
rish9511: @sahil21 
sure

3rdparty/ErrorHandling/errorcodes.h
rish9511: The input to MavenException class is the type of Error that can occur in ElMaven. Depending on the type of Error received, MavenException class may decide to act differently(To be done in future). For e.g,  if the type of the error received is a CRITICAL error, it will log the error differently compared to when the type will be DEBUG. 
Point being, both Classes have a different role to play 
Also, in future, we will  add more error types in the ErrorCode class.

Therefore I believe it's better if we keep them separate

Does this make sense ?

rish9511: Yes.
should I add this to TODO  or do you want me to change it now ?

src/core/libmaven/mzSample.cpp
rish9511: @sahil21 That won't happen. If for any reason uploading of sample fails, we will know about it,
either here or in [mzFileIo](https://github.com/ElucidataInc/ElMaven/pull/652/files/f9ccbce4192ae66e4d95b23e1044e4df20a6211b#diff-92d3a1d8ef6524a4f786409ff4f221c5R480)

3rdparty/ErrorHandling/errorcodes.h
sahil21: @rish9511 Makes sense. Thanks for the explanation.

src/core/libmaven/mzSample.cpp
sahil21: @rish9511 Okay. got it

PR #654, opened 2018-04-03
Upload logs on sentry

rish9511:
Major highlights:

- store logs in generic config location 

- upload logs to sentry when El-Maven crashes


CrashReporter/mainwindow.cpp
```diff
@@ -9,78 +9,171 @@ MainWindow::MainWindow(QWidget *parent) :
...
+void MainWindow::startElMaven()
+{
+    if (this->windowState == 1) {
+        QProcess * myProcess = new QProcess();
+        myProcess->start(this->restartApplicationPath);
+    }
+
 }
 
 void MainWindow::uploadToS3Done() {
```
shubhra-agrawal: @rish9511 Lets remove this function?

src/gui/mzroll/main.cpp
```diff
@@ -36,12 +37,27 @@ Database DB;
...
 
+void initializeLogger()
+{
+    QDir dir;
+    QString path = QStandardPaths::writableLocation(QStandardPaths::GenericConfigLocation) + QDir::separator() + \
+                   qApp->organizationName() + QDir::separator() + qApp->applicationName() + QDir::separator() + "logs" \
+                   + QDir::separator() ;
+
+    // Logs  won't be saved anywhere if this mkpath returns False;
+    dir.mkpath(path);
```
shubhra-agrawal: @rish9511 Shouldn't we at least inform the user? Maybe later, but I'd suggest you add a TODO here.

CrashReporter/mainwindow.cpp
rish9511: @shubhra-agrawal 
Let's do this later ?

PR #655, opened 2018-04-03
Get rid of PII

rish9511:
- Do not store any PII 

- Merge Error Code and Error message


3rdparty/ErrorHandling/errorcodes.h
```diff
@@ -3,31 +3,26 @@
...
-        ParseCsv
-    };
-
-};
-
 //singleton
 class ErrorMsg
 {
     public:
+        enum Errors {
```
sahil21: @rish9511 Why are we not assigning any integer value?

rish9511: They all assigned an Integer value implicitly. 
You want to assign some other values to them ?

sahil21: @rish9511 Just for convenience for developers. This is also fine.

PR #656, opened 2018-04-03
Fix windows and mac

rish9511:
Major highlights

- Fix issues on Mac and win


src/gui/mzroll/mainwindow.cpp
```diff
@@ -71,56 +70,51 @@ void MainWindow::setValue(int value)
 }
 
 void MainWindow::printvalue() {
```
sahil21: @rish9511 Don't you think functions like these should not be part of mainwindow.cpp? Let's not change it right now but it's a good idea to not it somewhere. maybe add TODO comment?

rish9511: Yes, this one is pretty useless. 
Will add to this to TODO

PR #663, opened 2018-04-05
Elmaven polly integration with cli

Ky5595:


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -130,7 +132,28 @@ void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
...
 			mavenParameters->massCutoffMerge->setMassCutoffAndType(atof(optarg),"ppm");
 			break;
-			
+
+		case 'P':
+			//parse polly specific arguments here.
+			uploadToPolly_bool = true;
+			{
+			pollyArgs = QString(optarg);
+			test_list = pollyArgs.split(",");
```
sahil21: @Ky5595 test_list? The variable name doesn't seem right.

IpankajI:
@Ky5595 crash point-

1. Start El-Maven
2. Click **Polly** button without uploading sample
3. Close Login Form Window
4. Click **Load project** or **Upload** in **Load from Polly** and **Upload to Polly** respectively

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -585,7 +608,7 @@ string PeakDetectorCLI::cleanSampleName(string sampleName) {
         return out.toStdString();
 }
 
-void PeakDetectorCLI::writeReport(string setName) {
+void PeakDetectorCLI::writeReport(string setName,QString jsPath,QString nodePath) {
```
sahil21: @Ky5595 You can't pass these arguments here. Not related to writeReport function.

IpankajI:
@Ky5595 **Upload to Polly** tab has **or** to choose from two set, it should be changed to display one option at a time, better talk to UI/UX team to improve this.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -639,6 +685,47 @@ void PeakDetectorCLI::saveJson(string setName) {
 	}
 }
 
+QString PeakDetectorCLI::UploadToPolly(QString jsPath,QString nodePath,QStringList filenames) {
```
sahil21: @Ky5595 The whole idea was not to write a lot of logic in the CLI or GUI of El-MAVEN and have a separate library which takes care of that. I believe this function (uploadToPolly) should be part of the library .

src/cli/peakdetector/PeakDetectorCLI.h
```diff
@@ -52,6 +54,14 @@ class PeakDetectorCLI {
 	public:
 		PeakDetectorCLI();
 
+		QString username;
```
sahil21: @Ky5595 Lots of these member variables should be just local to some function and rather avoid them to be part of the class.

src/pollyCLI/pollyintegration.cpp
```diff
@@ -0,0 +1,362 @@
...
+// PURPOSE:
+//    This function parses the output of "createproject" command returned by run_qt_process
+// this command is used to create a new project on polly
+// this function Returns project id for the new project..
+// CALLS TO: none
+//
+// CALLED FROM: createProjectOnPolly
+
+
+QString PollyIntegration::get_run_id(QByteArray result){
```
IpankajI: @Ky5595 pass **result** by **const** reference becauze size of **result** can be larger which will unnecessarily be copied every time.

IpankajI:
@Ky5595 **QStringList PollyIntegration::get_projectFiles_download_url_commands** and **QStringList PollyIntegration::get_project_upload_url_commands** share same code, please redesign according to DRY

src/pollyCLI/pollyintegration.cpp
```diff
@@ -0,0 +1,362 @@
...
+// name OF FUNCTION: check_already_logged_in
+// PURPOSE:
+//    This function checks if the user is already logged in or not.
+// Returns log in status of current user
+
+// CALLS TO: run_qt_process
+//
+// CALLED FROM: authenticate_login
+
+int PollyIntegration::check_already_logged_in(){
```
IpankajI: @Ky5595 return value should be **bool**

IpankajI:
@Ky5595 Logic of **int PollyIntegration::check_already_logged_in()** should be in **int PollyIntegration::authenticate_login(QString username,QString password)** because these two methods are not big and sharing related purpose.

src/cli/peakdetector/PeakDetectorCLI.cpp
Ky5595: @sahil21  For this sprint, we can pass these as it is, it will require significant changes in the code. 
In the next sprint, we are going to remove the node_modules from repo in any case.

src/cli/peakdetector/PeakDetectorCLI.h
Ky5595: @sahil21 I will add it to my TODO list and take up in next sprint

src/pollyCLI/pollyintegration.cpp
```diff
@@ -0,0 +1,362 @@
...
+    arg << args; // required params by that command . eg username, password
+
+    // nodePath = "PATH_OF_MAVEN/bin/node.exe"
+    process.setProgram(nodePath);
+    process.setArguments(arg);
+
+    qDebug () << "program: " <<  process.program()  << "args: " << process.arguments();
+
+    process.start();
+    process.waitForFinished(-1);
```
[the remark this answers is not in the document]
Ky5595: @IpankajI  This Is something that I am aware of and in the next sprint, we are coming up with a better workflow that will fix this issue.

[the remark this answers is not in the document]
sp-eldata: Don't think we should let this pass right? Might lead to frustrating user experience no?

src/cli/peakdetector/PeakDetectorCLI.h
sahil21: @Ky5595 Can you add TODO(s) as code comments? Personal to-do lists have the ability to get lost :p

Ky5595: @sahil21  Yes, Did that only.. You can find TODO(s) in the updated version of this PR..

sahil21: @Ky5595 Thanks!

src/pollyCLI/pollyintegration.cpp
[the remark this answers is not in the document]
rish9511: @sp-eldata 
it's less likely(will fail 1 out of 10 times) that the function will wait for infinite time. In most of the cases the function will return back due to some process error.
 
We will take care of this in the next release

PR #666, opened 2018-04-06
Ftr obi warp

IpankajI:
1. I have taken source-code of OBI-Warp from [here](http://obi-warp.sourceforge.net/) and used it as 3rd party library.
2. This alignment algorithm align retention times of samples without peak picking and grouping.
3. It takes retention times, m/z-values and intensity-values as input from samples (read [this](http://obi-warp.sourceforge.net/tutorials/mass_spectrometry_alignment_example.html) to understand input format)
4. One sample is used as reference sample and others are aligned with this (in our implementation, reference sample is chosen randomly from uploaded samples, option to choose sample will be added shortly)
5. **Undo Alignment** button is fixed to work with OBI-Warp.
6. This algorithm is written in C++ and does not have any other dependency( will work on Windows, Mac and Linux smoothly)
7. **Alignment visualization for Groups** will not be shown with this one as this algorithm is not dependent on grouping.

There are few advance params to control alignment-
1. Most relevant parameter is **Response**( Noisier the data, higher the value(0-100) ).
2. Bin Size is used to calculate mz-points as mentioned [here](http://obi-warp.sourceforge.net/tutorials/mass_spectrometry_alignment_example.html)
3. To understand uses of other parameters, refer to the paper shared earlier
Notes: Use default parameters' value and if need be, adjust **Response** only


3rdparty/obiwarp/obiwarp.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    _mat.take(mat);
+
+}
+
+vector<float> ObiWarp::align(vector<float> &rtPoints, vector<float> &mzPoints, vector<vector<float> >& intMat){
+    
+    vector<float> alignedRts;
+    if(rtPoints.size() == 0 || mzPoints.size() == 0 || _tm.len() == 0 || _mz.len() == 0)
+        return alignedRts;
```
rish9511: The function that takes `alignedRts`(i.e `Aligner::alignSampleRts`) should make sure to use the vector only if the size  of vector is not zero

3rdparty/obiwarp/obiwarp.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "vec.h"
+#include "mat.h"
+#include "dynprog.h"
+
+using namespace std;
+
+class ObiWarp{
+public:
+    ObiWarp(string score="cor",bool local = false, float factor_diag = 2.f, float factor_gap = 1.f,
```
rish9511: Should we not use a`struct` here ? 
You have added a struct ,`obiParams`, in `mzAligner.cpp` , can we not pass that struct here ?

3rdparty/obiwarp/obiwarp.h
```diff
@@ -0,0 +1,46 @@
...
+#include "mat.h"
+#include "dynprog.h"
+
+using namespace std;
+
+class ObiWarp{
+public:
+    ObiWarp(string score="cor",bool local = false, float factor_diag = 2.f, float factor_gap = 1.f, 
+            float gap_init = 0.2f,float gap_extend = 3.4f, float init_penalty = 0.f, 
+            float response = 100.f, bool nostdnrm = false  );
```
rish9511: Should we not use astruct here ?
You have added a struct ,obiParams, in mzAligner.cpp , can we not pass that struct here ?

3rdparty/obiwarp/obiwarp.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void ObiWarp::tm_axis_vals(VecI &tmCoords, VecF &tmVals,VecF &_tm ,int _tm_vals){
+    VecF tmp(tmCoords.length());
+    for (int i = 0; i < tmCoords.length(); ++i) {
+        if (tmCoords[i] < _tm_vals) {
+            tmp[i] = _tm[tmCoords[i]];
+        }
+        else {
+            printf("asking for time value at index: %d (length: %d)\n", tmCoords[i], _tm_vals);
+            exit(1);
```
rish9511: So, we simply exit from here ?
We should get back to: 
1. `mzAligner` and inform that the alignment process failed
2.  tell the user that the alignment was not performed ?(can be done later ?)

src/core/libmaven/Scan.cpp
```diff
@@ -15,6 +15,7 @@ Scan::Scan(mzSample* sample, int scannum, int mslevel, float rt, float precursor
 	this->precursorCharge=0;
 	this->precursorIntensity=0;
     this->sampleNumber = -1;
+    this->originalRt = -1;
```
shubhra-agrawal: @IpankajI Better to assign the originalRt here instead of doing this in mzAligner.

IpankajI: @shubhra-agrawal done!

3rdparty/obiwarp/obiwarp.h
IpankajI: @rish9511 done!

3rdparty/obiwarp/obiwarp.cpp
IpankajI: @rish9511 we'll do it when will be adding code for error handling.

3rdparty/obiwarp/obiwarp.cpp
IpankajI: @rish9511 removed from here!

PR #686, opened 2018-04-24
[Ref] Add a slice to each PeakGroup

sahil21:
This PR is meant to introduce the following changes:
- `Compound` member of the `PeakGroup` has been replaced with `mzSlice`. The compound associated with a group can still be queried.
- Slices and compounds will be accessible through getters/setters. Furthermore, every `PeakGroup` will hold its copy of a slice that is only wholly modifiable (its attributes will not be allowed to set through the access of a raw pointer).
-  EIC widget now sports a toggle button which allows two modes of group display. By default (sync toggled off), the behavior will be the same as it has been until now - changes to global mass tolerance values will change the visualized slice of a peak group. Once toggled on, however, groups will be shown using the original slice from which they have been created.
- Save slice attributes in emDB save files, enabling sessions to be restored as they were originally.

- [x] **TODO**: Add icons for the tolerance-sync toggle button.


src/core/libmaven/groupFiltering.cpp
```diff
@@ -64,7 +64,7 @@ bool GroupFiltering::filterByMS1(PeakGroup &peakgroup)
         return true;
 
     if (compound)
-        peakgroup.compound = compound;
+        peakgroup.setCompound(compound);
```
shubhra-agrawal: @saifulbkhan Compound assignment should happen in Group stats or when the group is created. Not in filtering. I know this is not directly related to this PR so we can do this in a separate branch too, just mentioning it here for reference.

shubhra-agrawal: wait, it's taking `_slice->compound` and assigning it again to `_slice` in `setPeakGroup`

saifulbkhan: Ah correct. Will do.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -203,12 +197,50 @@ bool PeakGroup::isMS1()
...
+
+Compound* PeakGroup::getCompound()
+{
+    if (hasSlice()) {
+        return _slice.compound;
+    }
+    return NULL;
+}
+
+void PeakGroup::setCompound(Compound* compound)
```
shubhra-agrawal: Would be safer to check whether a slice exists before setting a compound.

src/core/libmaven/datastructures/mzSlice.cpp
```diff
@@ -3,6 +3,31 @@
 #include "masscutofftype.h"
 #include "mzMassCalculator.h"
 
+mzSlice::mzSlice(float minMz, float maxMz, float minRt, float maxRt)
+{
+	mzmin = minMz;
+	mzmax = maxMz;
+	rtmin = minRt;
+	rtmax = maxRt;
+	mz = minMz + (maxMz - minMz) / 2;
```
shubhra-agrawal: This can be reduced to `mz = (minMz + maxMz)/2` and `rt = (minRt + maxRt)/2`

src/core/libmaven/groupFiltering.cpp
shubhra-agrawal: This still needs to be removed.

src/projectDB/projectdatabase.cpp
```diff
@@ -186,7 +186,12 @@ int ProjectDatabase::saveGroupAndPeaks(PeakGroup* group,
                      , :fragmentation_spearman_rank_corr   \
                      , :fragmentation_tic_matched          \
                      , :fragmentation_num_matches          \
-                     , :sample_ids                         )");
+                     , :sample_ids                         \
```
shubhra-agrawal: If user loads an emDB file from an older version of El-MAVEN, the slice will be zero for every saved group. As discussed offline, let's disable the Tolerance Lock for groups with zero slice.

src/core/libmaven/PeakGroup.cpp
saifulbkhan: An empty slice (having zeros as its bounds) is the default state. So there is no chance of an invalid assignment (that leads to a crash).

src/core/libmaven/datastructures/mzSlice.cpp
saifulbkhan: Yes it can.

src/projectDB/projectdatabase.cpp
saifulbkhan: Will make the change.

saifulbkhan:
@shubhra-agrawal I've made the changes and even added icons for tolerance switch. Please take a look.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3924,9 +3924,30 @@ QWidget* MainWindowWidgetAction::createWidget(QWidget *parent) {
...
+                [=](PeakGroup* selectedGroup) {
+                    if (selectedGroup == nullptr)
+                        return;
+
+                    auto& slice = selectedGroup->getSlice();
+                    if ((mzUtils::almostEqual(slice.mzmin, 0.0f)
+                         && mzUtils::almostEqual(slice.mzmax, 0.0f))
+                        || (mzUtils::almostEqual(slice.rtmin, 0.0f)
+                            && mzUtils::almostEqual(slice.rtmax, 0.0f))) {
+                        setChecked(false);
```
shubhra-agrawal: The tolerance switch is not being unchecked and disabled here.

shubhra-agrawal:
@saifulbkhan Tested this on Mac today and I don't think this is there yet. Untargeted, especially is not working as expected. I understand that this branch works well with fix_untargeted but I don't know when that one will be merged so let's iterate over this a bit more.

saifulbkhan:
@shubhra-agrawal I have fixed the issue with untargeted peak-groups and the tolerance switch, changed enabling, disabling and checking methods to be called as members of `toleranceSyncSwitch` and also disabled the switch for compounds as agreed from our earlier discussion that it does not make sense for them. Please let me know if you face any other blockers.

sahil21:
Awesome! :D

PR #688, opened 2018-04-26
Reset settings

rish9511:
allow users to reset settings of options and peaks dialog to defaults

### DIFF
--- a/src/core/libmaven/mavenparameters.cpp
+++ b/src/core/libmaven/mavenparameters.cpp
@@ -133,6 +133,8 @@ MavenParameters::MavenParameters(string settingsPath):lastUsedSettingsPath(setti
     * the 'lastUsedSettingsPath'is not empty and the file
     * actually exits. If not, then load defaultSettingsData
     */
+
+
     if(!lastUsedSettingsPath.empty()) {
         ifstream ifs(lastUsedSettingsPath, std::ios_base::in);
         if(ifs.is_open()) {
@@ -465,6 +467,26 @@ bool MavenParameters::loadSettings(const char* data)
     return true;
 }
 
+
+void MavenParameters::reset(std::list<string> keys)
+{
+
+    pugi::xml_document xmlDoc;
+    pugi::xml_parse_result parseResult = xmlDoc.load_string(defaultSettingsData);
+
+    if(parseResult.status != pugi::xml_parse_status::status_ok)
+        std::cerr << "parsing error  : " << parseResult.status ;
+
+    pugi::xml_node pnode = xmlDoc.child("Settings");
+
+    for(pugi::xml_node_iterator it = pnode.begin(); it != pnode.end(); ++it) {
+
+        if( std::find(keys.begin(), keys.end(), it->name()) != keys.end() ) {
+            setOptionsDialogSettings(it->name(), it->text().get());
+            setPeakDetectionSettings(it->name(), it->text().get());
+        }
+    }
+}
 vector<mzSample*> MavenParameters::getVisibleSamples() {
 
 	vector<mzSample*> vsamples;
--- a/src/core/libmaven/mavenparameters.cpp
+++ b/src/core/libmaven/mavenparameters.cpp
@@ -468,7 +468,7 @@ bool MavenParameters::loadSettings(const char* data)
 }
 
 
-void MavenParameters::reset(std::list<string> keys)
+void MavenParameters::reset(const std::list<string>& keys)
 {
 
     pugi::xml_document xmlDoc;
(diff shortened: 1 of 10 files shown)


src/core/libmaven/mavenparameters.cpp
```diff
@@ -465,6 +467,28 @@ bool MavenParameters::loadSettings(const char* data)
     return true;
 }
 
+
+void MavenParameters::reset(const std::list<string>& keys)
```
sahil21: @rish9511 vector will be much faster than list but this is also fine as we have few keys. Pointing out for future usage.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -157,6 +158,11 @@ PeakDetectionDialog::PeakDetectionDialog(QWidget *parent) :
 }
 
 
+void PeakDetectionDialog::onReset()
```
sahil21: @rish9511 Why an extra layer of signal and slots?

sahil21:
@rish9511 @shubhra-agrawal
Which is better

Reset settings for individual dialogs
Reset settings for the whole tool
Both

src/gui/mzroll/peakdetectiondialog.cpp
rish9511: @sahil21 how is this an extra layer ?

src/core/libmaven/mavenparameters.cpp
rish9511: @sahil21 ya, we have hardly 20-30 keys. so won't make a difference

rish9511:
@sahil21 
> Which is better
Reset settings for individual dialogs
Reset settings for the whole tool
Both

The reset button in dialog boxes should only reset settings for that dialog . 
If we want to allow users to reset settings for the whole tool, then we should  provide that option separately

src/core/libmaven/mavenparameters.cpp
sahil21: @rish9511 Cool

src/gui/mzroll/peakdetectiondialog.cpp
sahil21: @rish9511 Sorry my bad. Ignore.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -157,6 +158,11 @@ PeakDetectionDialog::PeakDetectionDialog(QWidget *parent) :
...
+{
+    emit resetSettings(pdSettings->getSettings().keys());
```
shubhra-agrawal: @rish9511 There's another variable called odSettings for options dialog. Typo issue possibilities in the future. Can we change the variable names to something safer?

rish9511: @shubhra-agrawal 
can you suggest something better ? 
od = options dialog
pd = peaks dialog

shubhra-agrawal: @rish9511 peakSettings and optionSettings works as well.

src/gui/mzroll/controller.cpp
```diff
@@ -19,6 +21,19 @@ Controller::~Controller()
     delete mw;
 }
 
+void Controller::resetMP(QList<QString> keys)
+{
+
+    // we need to change the 'type' from QString to std::string since maven parameters
+    // needs to independent of Qt
+    QList<std::string> _keys;
```
shubhra-agrawal: @rish9511 Why not create a std list instead of converting a QList to std list before passing it to mavenParameters

src/gui/mzroll/forms/settingsform.ui
```diff
@@ -1404,6 +1404,13 @@ p, li { white-space: pre-wrap; }
      </widget>
     </widget>
    </item>
+   <item row="1" column="0">
+    <widget class="QPushButton" name="resetButton">
+     <property name="text">
+      <string>Reset defaults</string>
```
shubhra-agrawal: @rish9511 I'd suggest changing this to "Use Default Parameters". Much clearer for users

src/gui/mzroll/controller.cpp
rish9511: @shubhra-agrawal  
cant do much about this. resetMp receives the list of keys from peaks/options dialog. and both these dialogs give QList.

src/gui/mzroll/forms/settingsform.ui
rish9511: @shubhra-agrawal  okay

src/gui/mzroll/controller.cpp
shubhra-agrawal: @rish9511 I meant _keys could have been a std list

PR #694, opened 2018-05-02
Move isotope Plot functions out of eicwidget

shubhra-agrawal:
Changes made:

- Refactor to move isotope plot functions out of eicWidget
- move isotope plot to a dockwidget
- Plot does not clear unless group is changed
- Horizontal plot instead of vertical
- Add checkboxes for which labels should be displayed
- Sort samples in the same order as barplot
- Natural sort for labels
- Label information on hover at mouse position
- Opaque background for label info for clarity
- Use distinct colors for label bars 
- Add black separators between labels

### DIFF
--- a/src/gui/mzroll/isotopeplot.cpp
+++ b/src/gui/mzroll/isotopeplot.cpp
@@ -5,11 +5,11 @@
 using namespace Eigen;
 
 
-IsotopePlot::IsotopePlot()
+IsotopePlot::IsotopePlot(MainWindow *mw)
     :QGraphicsItem() {
 	// Initialised existing values - Kiran
 	_barwidth=10;
-	_mw=NULL;
+	_mw=mw;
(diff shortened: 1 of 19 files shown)


src/gui/mzroll/isotopeplot.cpp
```diff
@@ -5,17 +5,17 @@
 using namespace Eigen;
 
 
-IsotopePlot::IsotopePlot(QGraphicsItem* parent, QGraphicsScene *scene)
-    :QGraphicsItem(parent) {
+IsotopePlot::IsotopePlot(MainWindow *mw)
```
sahil21: @shubhra-agrawal Pass QWidget instead of MainWindow.

sahil21: QWidget *parent

PR #697, opened 2018-05-07
Fix crash in isowidget

shubhra-agrawal:
mavenParameters->_group is not thread safe at the moment. It is set by eicwidget, isotopewidget and isotopePlot in different threads. As a temporary fix, we are using selectedGroup which is thread-safe. 

Thread safety for mavenParameters->_group will be implemented in the next release.


src/gui/mzroll/eicwidget.cpp
```diff
@@ -1452,7 +1452,7 @@ void EicWidget::setCompound(Compound* c) {
 		getMainWindow()->mavenParameters->setPeakGroup(NULL);
 		resetZoom();
 	}
-
+return eicParameters->selectedGroup;
```
sahil21: Correct the indentation please

PR #704, opened 2018-05-10
[Fix] Load compound database through mzroll in MSMS datasets

sahil21:


src/gui/mzroll/ligandwidget.cpp
```diff
@@ -179,9 +179,9 @@ void LigandWidget::readCompoundXML(QXmlStreamReader& xml, string dbname) {
     rt = xml.attributes().value("rt").toString().toFloat();
     charge = xml.attributes().value("Charge").toString().toInt();
     formula = xml.attributes().value("Formula").toString().toStdString();
-    precursormz = xml.attributes().value("Precursor Mz").toString().toFloat();
-    productmz = xml.attributes().value("Product Mz").toString().toFloat();
-    collisionenergy = xml.attributes().value("Collision Energy").toString().toFloat();
+    precursormz = xml.attributes().value("precursorMz").toString().toFloat();
```
rish9511: will it never be "Precursor Mz" ? for any data set (lc-ms)?

rish9511: sorry! did not think it through before commenting
can you explain your change ?

shubhra-agrawal: We write this attribute in saveProject for all datasets. It is either empty in case of lcms or has a value for msms. @sahil21 correct me if i'm wrong.

sahil21: @rish9511 @shubhra-agrawal In the previous versions attributes for precursor m/z and product m/z were written as "Precursor Mz" with space. Adding space in an attribute is not valid XML format.

The change I have made has removed the space in the attribute name.

sahil21: @rish9511 @shubhra-agrawal For previous versions, this invalid format was not working and hence msms databases were not being loaded. This change will fix the issue.

PS we don't need to take care of backward compatibility here as in previous version it was not working at all.

rish9511: :+1:

Issue #693, opened 2018-04-27
Method to properly separate multiple peak groups for a single compound in labeled data export


lparsons:
The automated peak picking has been useful for us, and we are hoping to use it more in our workflows going forward. However, we have run into an issue with some of our downstream processing. There are times that the automated peak detection will detect two or more peak groups for a single compound. When the peak data is exported to a CSV file, it can be difficult to properly group the rows for the isotopes according to the compound *and* the peak group.

For example, the table below, there were two peaks detected for the `C17:0` compound. This is evident from the fact that there are two rows with `C12 PARENT` as the **isotopeLabel**, both with the same value (`C17:0`) for the *compound*. Currently, the only way I can see to properly group the isotopes together is to use the sort order (which is rather fragile). It would be great if there was a column that could be used to group together the isotopes for a given parent peak group. I had thought that **groupId** or **metaGroupId** would do this, but it appears not.

| label | metaGroupId | groupId | goodPeakCount | medMz | medRt | maxQuality | isotopeLabel | compound | compoundId | formula | expectedRtDiff | ppmDiff | parent | F-2-3 | F-2-2 | F-2-1 | F-1-3 | F-1-2 | F-1-1 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| | 0 | 19 | 3 | 269.2483 | 10.66541 | 0.3812263 | C12 PARENT | C17:0 | C17:0 | C17H34O2 | 0.0554142 | 1.246778 | 269.2483 | 968530.8 | 877007.3 | 798940.5 | 916631.2 | 835455.4 | 890473.7 |
| | 0 | 20 | 0 | 270.2516 | 10.66841 | 0.1329716 | C13-label-1 | C17:0 | C17:0 | C17H34O2 | 0.05840874 | 1.355072 | 269.2483 | 145705.1 | 179558.1 | 157521.6 | 187857.9 | 165838.1 | 188400.8 |
| | 0 | 21 | 0 | 271.2553 | 10.66063 | 0.1363112 | C13-label-2 | C17:0 | C17:0 | C17H34O2 | 0.0506258 | 0.112505 | 269.2483 | 0 | 21641.75 | 37308.98 | 0 | 38462.94 | 31614.97 |
[...]

PR #707, opened 2018-05-16
Fix groupID for correct isotopic parent #693

mayankmtg:
Multiple peak groups for a single compound in labeled data export were all given different labels even when  belonging to the same isotopic parent. The desirable outcome was to have groupIDs such that we could group together the isotopes for a given parent peak group.

### DIFF
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -79,7 +79,6 @@ void CSVReports::insertGroupReportColumnNamesintoCSVFile(string outputfile,bool 
         QString header = groupReportcolnames.join(SEP.c_str());
         groupReport << header.toStdString();
         for (unsigned int i = 0; i < samples.size(); i++) {
-            qDebug() << "mayank " << sanitizeString(samples[i]->sampleName.c_str());
             groupReport << SEP << sanitizeString(samples[i]->sampleName.c_str()).toStdString();
         }
         groupReport << endl;
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -79,6 +79,7 @@ void CSVReports::insertGroupReportColumnNamesintoCSVFile(string outputfile,bool 
         QString header = groupReportcolnames.join(SEP.c_str());
         groupReport << header.toStdString();
         for (unsigned int i = 0; i < samples.size(); i++) {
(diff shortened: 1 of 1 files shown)


src/core/libmaven/csvreports.cpp
```diff
@@ -79,6 +79,7 @@ void CSVReports::insertGroupReportColumnNamesintoCSVFile(string outputfile,bool
         QString header = groupReportcolnames.join(SEP.c_str());
         groupReport << header.toStdString();
         for (unsigned int i = 0; i < samples.size(); i++) {
+            qDebug() << "mayank " << sanitizeString(samples[i]->sampleName.c_str());
```
rish9511: haha
that's funny

shubhra-agrawal:
@chubukov This branch is a temporary fix to help internal users. 

We are still discussing the final implementation. Two of these are:
a. Every group has a different groupId. metaGroupId is same for all children of a parent group.
image
b. Every group has a different groupId but metaGroupId resets for every parent.
image

chubukov:
@shubhra-agrawal I think (a) is how it's historically been done, and what I've relied on. In fact my branch behaves correctly for the flat csv export (exactly as you show in (a). I think the bug was in the long peak details format, where metaGroupId just doesn't get set.

chubukov:
@shubhra-agrawal i guess also at some point if was fixed? https://github.com/ElucidataInc/ElMaven/pull/277 And then maybe re-broken? :)

shubhra-agrawal:
@chubukov That branch wasn't merged because of some corner cases related to deleting groups. We'll make sure the new implementation takes care of all such cases. In any case, we will discuss the final implementation with you before merging.

mayankmtg:
@chubukov @shubhra-agrawal I have not yet seen why metagroupID has changed. It seems to be corrupted on the way being exported to the csv. However, getting the groupId of the isotopic parent of a compound will essentially performs the same role as the metagroupId in this case. The csv has been updated with updated metagroupId. Now the csv is exported just like the one shown in (a).

shubhra-agrawal:
@mayankmtg It's crashing for untargeted run. Please look into it.

mayankmtg:
@shubhra-agrawal I am looking into this

shubhra-agrawal:
@mayankmtg Tested for untargeted, labeled and unlabeled LCMS and MRM data. Group summary and Peak detailed formats. Seems to be working well!

PR #713, opened 2018-05-25
Isotope Plot aesthetics

shubhra-agrawal:
Changes in this PR:

- Horizontal bars instead of vertical
- Black separator for every label fraction

### DIFF
--- a/src/core/libmaven/PeakGroup.h
+++ b/src/core/libmaven/PeakGroup.h
@@ -455,6 +455,14 @@ class PeakGroup{
         static bool compMz(const PeakGroup& a, const PeakGroup& b ) { return(a.meanMz > b.meanMz); }
 
         /**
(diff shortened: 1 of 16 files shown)


chubukov:
@shubhra-agrawal I like the cleaner look, that's great.

I clicked through two compounds and on the third I got a crash :( I believe that this happens only after I drag the isotope plot out of the docked position, but there doesn't seem to be any clear pattern regarding what compound it happens on.

shubhra-agrawal:
@chubukov Did you experience the crash on develop? We got a few crashes on Mac that we are working on right now. Linux seems to be stable but haven't tested much on Windows yet. We'll get these resolved before the beta release.

chubukov:
@shubhra-agrawal Yes, I just pulled and built `develop` a few hours ago.

PR #714, opened 2018-05-29
Breakpad mac

rish9511:


CrashReporter/file_uploader.cpp
```diff
@@ -19,23 +25,54 @@ FileUploader::~FileUploader()
...
+#ifdef Q_OS_WIN
     typedef std::basic_string<wchar_t> wstring;
     std::map<wstring, wstring> files;
     std::map<wstring, wstring> parameters;
     parameters[L"Logs"] =  QString(additionalData).toStdWString();
     files[L"upload_file_minidump"] = dmpFilePath.toStdWString();
     qDebug() << "uploading file: " << dmpFilePath;
 
-    return google_breakpad::HTTPUpload::SendRequest(
+    google_breakpad::HTTPUpload::SendRequest(
```
sahil21: @rish9511 SendRequest might be an asynchronous call. So before even completing SendRequest, emit uploadDone() might get executed in that case.

bin/sup.sh
```diff
@@ -0,0 +1 @@
+curl -X POST   'https://sentry.io/api/294375/minidump?sentry_key=5428a76c424142128a3ff1c04e5e342e'   -F upload_file_minidump=@$1 -F logs=$2
```
sahil21: @rish9511 Is this write only?

src/gui/mzroll/mzroll.pro
```diff
@@ -45,6 +45,14 @@ win32 {
     LIBS += -lgoogle-breakpad -pthread
 }
 
+mac {
+    INCLUDEPATH  += $$top_srcdir/3rdparty/google-breakpad/src/
+    QMAKE_LFLAGS += -L$$top_builddir/libs/
+    LIBS += /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
```
sahil21: @rish9511 Not sure what this is. Can you add comments in the code?

bin/sup.sh
sahil21: @rish9511 Can we do POST request using QT Network libraries instead of using bash script?
If yes, then let's add TODO in the code to do it in future.

CrashReporter/file_uploader.cpp
rish9511: @sahil21  it's a synchronous request

src/gui/mzroll/mzroll.pro
rish9511: @sahil21  google-breakpad uses native apis. hence we need to link against those on all the platforms

bin/sup.sh
rish9511: yes, i'll add this in TODO

CrashReporter/mainwindow.cpp
```diff
@@ -31,25 +31,27 @@ MainWindow::MainWindow(QWidget *parent, const QString& path) :
...
             nodePath = qApp->applicationDirPath() + QDir::separator() + "node";
-        #endif
 
-        #ifdef Q_OS_MAC
-        QString binDir = qApp->applicationDirPath() + QDir::separator() + ".." + QDir::separator() + ".." + QDir::separator() + ".." + QDir::separator();
-        if(!QStandardPaths::findExecutable("node", QStringList() << binDir + "node_bin" + QDir::separator() ).isEmpty())
-            nodePath = binDir + "node_bin" + QDir::separator() + "node";
-        #endif
+
+//        #ifdef Q_OS_MAC
```
shubhra-agrawal: @rish9511 Can we remove all the commented out code? Assuming it's not going to be used from now on.

PR #716, opened 2018-05-30
Filtering Chromatogram ID (aka SRM ID in code)

sahil21:
A clean way of filtering SRM ID
- Added a separate function to do filtering
- Added a vector<string> which takes attributes to be removed as input

### DIFF
--- a/src/core/libmaven/Scan.h
+++ b/src/core/libmaven/Scan.h
@@ -3,6 +3,7 @@
 #include "mzSample.h"
 #include <QRegExp>
 #include <QString>
+#include <QStringList>
 
 class mzSample;
 class ChargedSpecies;
--- a/src/core/libmaven/Scan.h
+++ b/src/core/libmaven/Scan.h
@@ -1,6 +1,9 @@
 #ifndef SCAN_H
 #define SCAN_H
 #include "mzSample.h"
+#include <QRegExp>
+#include <QString>
+
 class mzSample;
 class ChargedSpecies;
 class MassCutoff;
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -416,7 +416,7 @@ void mzSample::parseMzMLChromatogramList(xml_node &chromatogramList)
 
 int mzSample::getSampleNoChromatogram(const string &chromatogramId) {
 
-	std::regex rxSampleNumber("sample\ *\=\ *([0-9]+)");
+	std::regex rxSampleNumber("sample\ *\=\ *([0-9]+)\ ");
 	std::vector<int> results;
     for(std::sregex_iterator i = std::sregex_iterator(chromatogramId.begin(), chromatogramId.end(), rxSampleNumber);
             i != std::sregex_iterator(); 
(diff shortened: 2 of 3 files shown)


src/core/libmaven/Scan.h
```diff
@@ -1,6 +1,10 @@
 #ifndef SCAN_H
 #define SCAN_H
 #include "mzSample.h"
+#include <QRegExp>
+#include <QString>
+#include <QStringList>
+
```
rish9511: unnecessary includes I guess. did you forget to remove them ?

src/core/libmaven/mzSample.cpp
```diff
@@ -427,6 +414,36 @@ void mzSample::parseMzMLChromatogromList(xml_node &chromatogramList)
 	}
 }
 
+int mzSample::getSampleNoChromatogram(string chromatogramId) {
```
rish9511: const reference

src/core/libmaven/mzSample.cpp
```diff
@@ -427,6 +414,36 @@ void mzSample::parseMzMLChromatogromList(xml_node &chromatogramList)
...
+    for(std::sregex_iterator i = std::sregex_iterator(chromatogramId.begin(), chromatogramId.end(), rxSampleNumber);
+            i != std::sregex_iterator(); 
+            ++i) 
+    { 
+        std::smatch m = *i;
+        results.push_back(std::stoi( m[1].str().c_str() ));
+    } 
+
+    for (auto n: results)
+        return n;
```
rish9511: it will always return after the first loop. so we can simply do `return results[0]` 
and if that's the case why do we need a `std::vector<int> results` ?

did you intend to do something else ?

src/core/libmaven/mzSample.cpp
```diff
@@ -427,6 +414,36 @@ void mzSample::parseMzMLChromatogromList(xml_node &chromatogramList)
...
+    
+    return -1;
+}
+
+string mzSample::filterChromatogramId(string chromatogramId) {
```
rish9511: const reference

src/core/libmaven/Scan.h
sahil21: @rish9511 No, it is being used by Scan class.

src/core/libmaven/mzSample.cpp
sahil21: @rish9511 Does that matter? We are not passing pointer anyway.

rish9511: yes it does. 
passing parameters this way performs a copy operation which essentially creates a new object and then copies everything in that object. 

using references make sure that no new object is created.

sahil21: @rish9511 But object size if small here, right? Anyway, it is a good practice to use a const reference. Thanks for pointing out. I'll change here

src/core/libmaven/Scan.h
rish9511: I just compiled the branch without these includes and it compiled fine

sahil21: @rish9511 It is failing on Travis. Not on my system also.

src/core/libmaven/mzSample.cpp
sahil21: @rish9511 

Consider a string, where we are trying to match some regex. There are 2 corner cases we need to take care of: 
- **What if there are two substrings which match the regex** -- Hence there is a vector of results
- **What if there is no substring which matches the regex** -- The for loop for results won't execute and hence it will skip that part.

I agree there will be numerous ways to do this. Let me know if this still confuses you. I'll make changes in the code.

sahil21: @rish9511 We are taking here the assumptions that we only want the first result.

Issue #724, opened 2018-06-04
debug statements slow down mzML import


chubukov:
When I got rid of the debug output in the mzML parser, import speed increased by an order of magnitude. 

Bunch of other debug output could be cleaned up as well.

V

shubhra-agrawal:
@chubukov The sample import speed took a hit in v0.3 but it's back to normal now. We've also done a few optimisations to make it a little faster for bigger datasets.

Irrespective of that, cleaning up debug statements should help. We have removed a few of these. Will clean it up more soon.

shubhra-agrawal:
Merged to develop. Will be available as part of v0.8

PR #728, opened 2018-06-06
[Fix] Increase mzML import speed  #724

omi02:


shubhra-agrawal:
@omi02 Not sure if it's resulting in any speed-up but it's good to clean up some debug statements. Let's look into the mzML import speed in more detail soon.

Issue #152, opened 2017-02-28
look into mass slicing


chubukov:
It happens all the time that the result of untargeted peak detection includes a bunch of peaks that overlap absolutely perfectly in m/z - rt space. 

I would like to try to understand why this happens and if it could be improved.

chubukov:
Assigning to myself for now, but happy to have anyone look into it.

r-el-maya:
@chubukov 
   
-  First we will go through the algorithm with the help of code documentation.
-  Check for overlapping of the mass slices.
-  If such a probability should exist, will have to modify the algo accordingly.

chubukov:
@r-el-maya @sahil21 one additional question on this: am I right that the "Auto-detect and ignore isotopes" radio button does absolutely nothing?

sahil21:
@chubukov Yes, it is not being used anywhere. This option was not part of Maven build 769 so it came in between 769 and 775.

r-el-maya:
@chubukov  Replying to https://github.com/ElucidataInc/ElMaven/commit/6b12810eaa5cd838619d065fccb759ca576943e8#commitcomment-23020303

Yes, this function is called after creating the slices. Currently, we are not changing the way how the slices are being created. Though we know how and why the overlapping(sometimes 100%) occurs, we can’t avoid overlaps just by changing some parts of the algorithm.

The reason behind that being, while iterating over samples and in turn scans, for each observation, we are creating slices, if there is no already existing slice in the range ‘mz-mztolerance’ to ‘mz+mztolerance’. This case doesn’t create any problem. But, if there already exists any such slice, we merge this new one with the existing one, expanding the original one. This expansion when done iteratively, creates problem causing huge percentage of overlaps.

Overlaps are indeed necessary to facilitate the range(mz & rt) broad enough to have all the data of a particulate peakgroup in one slice. But we only have 
Problem when the overlap is more than some extent.

This implementation is just a temporary fix. We are thinking of a new algorithm that will avoid such unnecessary overlaps.

chubukov:
@r-el-maya Thanks very much for the more detailed explanation. I'm still not sure I fully understand (when it merges two slices, does it not eliminate one of them?). But I think your approach of a temporary fix that simply detects hugely overlapping slices is totally reasonable.

[...]

PR #730, opened 2018-06-12
Mass Slicing correlation based post processing #152

mayankmtg:
El-maven reports duplicates because of similar overlapping slice templates. The number is observed to be as high as 10% to 20% of the reported groups. Traversing through the slices and finding the duplicates according to some metric could be the way to go with no considerable performance overhead. We can traverse through the slices and find duplicates in three steps:
1. mz-difference <0.01. Compare the differences in meanMZ values and choose the ones with difference less than this.
2. rt-difference <0.1. Compare the differences in meanRT values and choose the ones with difference less than this.
3. Vector correlation between sample wise intensity values > 0.99.

These must be performed in the same order to reduce the overhead caused by correlation calculation. The constant values may be varied to find duplicates better and faster.

### DIFF
--- a/src/core/libmaven/groupFiltering.cpp
+++ b/src/core/libmaven/groupFiltering.cpp
@@ -132,3 +132,45 @@ bool GroupFiltering::quantileFilters(PeakGroup *group) {
     }
     return false;
 }
+
+void GroupFiltering::deleteDuplicateGroup(){
+    double mzdiff=0.01;
+    double rtdiff=0.1;
+    double corrthresh=0.99;
+    vector_intensity_array.clear();
+    vector_mz_array.clear();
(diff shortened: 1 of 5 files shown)


src/core/libmaven/groupFiltering.cpp
```diff
@@ -131,4 +131,46 @@ bool GroupFiltering::quantileFilters(PeakGroup *group) {
         return true;
     }
     return false;
+}
+
+void GroupFiltering::deleteDuplicateGroup(){
+    double mzdiff=0.01;
```
shubhra-agrawal: @mayankmtg all variable names should be in camel case. This is for consistency across this code-base not a general rule.

PR #733, opened 2018-06-13
El-MAVEN-Polly Integration

shubhra-agrawal:
Requirements:
- Add New Project option should be selected by default every time it opens
- Loading gif while projects are being fetched on opening polly interface anytime after the login
- Advanced settings show up in another dialog
- Go To Polly button takes user to the first view app of the last uploaded project

### DIFF
--- a/src/gui/mzroll/ligandwidget.cpp
+++ b/src/gui/mzroll/ligandwidget.cpp
@@ -444,8 +444,9 @@ QTreeWidgetItem* LigandWidget::getItem(Compound* compound) {
 }
 
 void LigandWidget::saveCompoundList(){
+
     QSettings *settings = _mw->getSettings();
-    string dbname = databaseSelect->currentText().toStdString();
+    QString dbname = databaseSelect->currentText();
     QString dbfilename = databaseSelect->currentText() + ".tab";
     QString dataDir = settings->value("dataDir").value<QString>();
     QString methodsFolder =     dataDir +  "/"  + settings->value("methodsFolder").value<QString>();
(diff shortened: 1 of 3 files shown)


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -677,55 +696,124 @@ void PeakDetectorCLI::saveJson(string setName) {
 		#endif
 
 		jsonReports=new JSONReports(mavenParameters);
-		jsonReports->saveMzEICJson(mavenParameters->outputdir + setName + ".json",
+		string fileName = mavenParameters->outputdir + setName + ".json";
+		if (uploadToPolly_bool){
+			string fileName = setName;
```
sahil21: @shubhra-agrawal @Ky5595 fileName scope will end after the if statement. You should not initialize it again.

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -514,6 +514,67 @@ void LigandWidget::saveCompoundList(){
     }
 }
 
+void LigandWidget::saveCompoundListToPolly(QString fileName,QString dbname){
```
sahil21: @Ky5595 @shubhra-agrawal Why are we doing this? Why can't just take the file from user's system?

Ky5595: @sahil21  Basically, we want to upload the compound database that was used to generate the peak table, to Polly. We could have had a mapping of the compound databases present in the ligand widget and their actual location on the system.
But, To validate it before uploading, we had to write a function separately. Personally, I feel that will add one more redundant function to the code, so i used already existing functions from Elmaven code.

This takes care of the following scenarios -
1. If the user uploads some compound dataset to Elmaven, does some analysis and for some reason that file is deleted from the system.
If we use fixed mapping, Elmaven will crash. 
2. If the file is modified on the system, Elmaven will upload the file but further workflow on Polly will break.

The current code takes care of both these scenarios by writing to a temporary file from RAM (Elmaven session variables) and then uploading it to Polly.

sahil21: @Ky5595 saveCompoundList & saveCompoundListToPolly is sharing around 95% of code. Please refactor this.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -795,29 +795,75 @@ void TableDockWidget::exportGroupsToSpreadsheet() {
     }
 }
 
-void TableDockWidget::exportGroupsToSpreadsheet_polly(QString fileName) {
-    
+void TableDockWidget::prepareDataForPolly(QString writable_temp_dir,QString export_format,QString user_filename) {
```
shubhra-agrawal: @Ky5595 exportGroupstoSpreadSheet already exists in tabledockwidget. The code is almost the same. Lets refactor this before merging.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -838,6 +884,26 @@ void TableDockWidget::exportGroupsToSpreadsheet_polly(QString fileName) {
     }
 }
 
+void TableDockWidget::exportJsonToPolly(QString writable_temp_dir,QString jsonfileName) {
```
shubhra-agrawal: @Ky5595 Similar function exists

PR #734, opened 2018-06-13
Fix crash on develop

shubhra-agrawal:
Reverting the commit for natural sorting of isotopes. The PeakGroup vector is not threadsafe and is resulting in random crashes on Windows and Mac. 

The order of labels has been changed while adding isotopes to the parent group. Still need to debug where the order is getting changed again on its way to the isotope plot.


shubhra-agrawal:
@sahil21 Reverting commit fba2ec1

PR #736, opened 2018-06-18
Bug fix in isotope detection

shubhra-agrawal:
Certain isotopic children had misleading RTs and/or zero intensities. This fix will reject isotopes that have 0 intensity within +/- 2 scans from the parent peak or if no m/z matches were found within that scan range.


chubukov:
I'd like to understand this a little better. What's the case you're catering to that's not covered by the other filters? It seems like this way if you have one bad scan you would miss the peak. Or do you mean zero intensity summed over the 5 scans?

I guess if you're using areaTop that's a bad case anyway... that was the motivation for some of the smoothed metrics, but that's another discussion.

shubhra-agrawal:
@chubukov This filter is for the case where the isotopic m/z is not found at all within +/-2 scan range of the parent peak. This range is hard-coded at the moment but in the future it will be replaced by maxIsotope scan difference set by the user.

PR #745, opened 2018-06-22
fix: fix getting started page

rish9511:
- show getting started page after mainwindow


src/gui/mzroll/gettingstarted.cpp
```diff
@@ -1,59 +1,50 @@
...
     }
-    return false;
 }
 
-void GettingStarted::setFlag()
+void GettingStarted::setFlag(int state)
 {
-    GettingStarted::close();
+
+    qDebug() << "state : " << state;
```
shubhra-agrawal: @rish9511 Please remove the debug statement.

PR #762, opened 2018-06-29
fix: use default model file shipped with the application

rish9511:


src/gui/mzroll/mainwindow.cpp
```diff
@@ -226,7 +226,20 @@ using namespace mzUtils;
       _massCutoffWindow->setMassCutoffAndType(massCutoff,massCutoffType);
     */
 
-    QString clsfModelFilename = settings->value("clsfModelFilename").value<QString>();
+
+    // always load model file present in the bin directory.
+    QString clsfModelFilename;
+
+    #if defined(Q_OS_WIN) || defined(Q_OS_LINUX)
+      clsfModelFilename = QApplication::applicationDirPath() + "/" + "default.model";
```
sahil21: Use this instead 

QDir::cleanPath(QApplication::applicationDirPath() + QDir::separator()  + "default.model");

rish9511: 👍 thanks for pointing out

PR #796, opened 2018-07-17
[Fix] Check for peaks before bookmarking a group

shubhra-agrawal:
The check for peaks has been moved to the bookmark function as shift-drag is the only way to get a group with no peaks. 

In case of mzrolls, groups are read and added before the peaks are assigned which led to filtering out of all peaks.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -1193,6 +1193,8 @@ PeakGroup* MainWindow::bookmarkPeakGroup() {
 PeakGroup* MainWindow::bookmarkPeakGroup(PeakGroup* group) {
 
 	if ( bookmarkedPeaks == NULL ) return NULL;
+	//TODO: User feedback when group is rejected
+	if (group->peakCount() == 0) return NULL;
 
     if ( bookmarkedPeaks->isVisible() == false ) {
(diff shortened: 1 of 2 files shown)


rish9511:
@shubhra-agrawal 
I think that the check that was added in the last commit did make sense. Any  group that does not have any peaks should not be added in the peaks table ideally.But that check has been removed now and has been moved to `MainWindow::bookmarkPeakGroup` which means we only perform this check when we are adding groups to bookmark table, which is okay but Imo the same check should also be present in `TableDockWidget::addPeakGroup`. 

As to why on loading mzRoll peaks table were not being loaded is because of the way we parse mzRoll. 
check this: https://github.com/ElucidataInc/ElMaven/blob/develop/src/gui/mzroll/tabledockwidget.cpp#L2077

Essentially it:
1. read the Groups
2. adds them to `allgroups` by calling `TableDockWidget::addPeakGroup`. The reason why peaks table did not show any groups. 
3. add peaks to the groups 

So, imo we should change the way we parse mzRoll and have that check at both the places
What do you think ?

rish9511:
@shubhra-agrawal 

> We can take that up when we're working on the mzroll issues

cool. lets leave it right now then

shubhra-agrawal:
@rish9511 I'll create a low priority issue for this so we don't lose track of it.

PR #808, opened 2018-07-24
Profile sample uploading

rish9511:
Reasons that led to slow down of sample uploading:
1. enumerateSRMScan() and  calculateMzRtRange() were called thrice for every sample
2. Ofast and ffast-math flag were removed and hence compiler was not optimizing parts of code involved in sample uploading process

Solution
1. removed duplicated calls 
2. added compiler flags for windows and linux

Tested on:
Mac
Linux
Windows

### DIFF
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -305,7 +305,7 @@ void mzSample::parseMzML(const char *filename)
 	xml_node experimentRun = doc.first_child().first_element_by_path("mzML/run");
 	if (!experimentRun.empty())
 	{
-		parseMzMLInjectionTimeStamp(experimentRun);
+        parseMzMLInjectionTimeStamp(experimentRun.attribute("startTimeStamp"));
 	}
 
 	//Get a spectrumstore node
@@ -322,9 +322,8 @@ void mzSample::parseMzML(const char *filename)
 	}
 }
 
-void mzSample::parseMzMLInjectionTimeStamp(xml_node &experimentRun)
+void mzSample::parseMzMLInjectionTimeStamp(const xml_attribute &injectionTimeStamp)
 {
-	xml_attribute injectionTimeStamp = experimentRun.attribute("startTimeStamp");
 
 	if (!injectionTimeStamp.empty())
 	{
@@ -339,7 +338,7 @@ void mzSample::parseMzMLInjectionTimeStamp(xml_node &experimentRun)
 	}
 }
 
-void mzSample::parseMzMLChromatogramList(xml_node &chromatogramList)
+void mzSample::parseMzMLChromatogramList(const xml_node &chromatogramList)
 {
 
 	int scannum = 0;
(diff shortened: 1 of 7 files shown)


src/core/libmaven/mzSample.cpp
```diff
@@ -305,7 +305,7 @@ void mzSample::parseMzML(const char *filename)
 	xml_node experimentRun = doc.first_child().first_element_by_path("mzML/run");
 	if (!experimentRun.empty())
 	{
-		parseMzMLInjectionTimeStamp(experimentRun);
+        parseMzMLInjectionTimeStamp(experimentRun.attribute("startTimeStamp"));
```
saifulbkhan: Please correct indentation

saifulbkhan: Please ignore this. This appears totally fine in the editor.

src/core/libmaven/mzSample.cpp
```diff
@@ -899,7 +898,7 @@ void mzSample::populateFilterline(string filterLine, Scan *_scan)
 	}
 }
 
-void mzSample::parseMzXMLScan(const xml_node &scan, int scannum)
+void mzSample::parseMzXMLScan(const xml_node &scan, const int& scannum)
```
shubhra-agrawal: @rish9511 Why pass int value as a const ref? pass-by-value would surely be same or faster in this case. Not really an issue, just curious.

rish9511: @shubhra-agrawal  
references make the code memory efficient not necessarily faster(unless it's a really large object)
if the parameters are just being read, they should be received as const references

shubhra-agrawal: I get that. I was reading about passing by const ref recently and most sources were divided on what is better for native datatypes: sending by value or const ref. Just wanted to know why you chose one over the other. Seems you're going with convention here, which is good. Thanks!

src/core/libmaven/mzSample.h
```diff
@@ -706,7 +706,7 @@ class mzSample
 
     void setInstrumentSettigs(xml_document &doc, xml_node spectrumstore);
 
-    void parseMzXMLData(xml_document &staticdoc, xml_node spectrumstore);
+    void parseMzXMLData(const xml_node& spectrumstore);
```
shubhra-agrawal: Good eye! :)

src/gui/mzroll/mzfileio.cpp
```diff
@@ -583,16 +574,8 @@ void mzFileIO::fileImport(void) {
             QString filename = samples.at(i);
             mzSample* sample = loadSample(filename);
             if (sample) {
-                    sample->enumerateSRMScans();
-                    sample->calculateMzRtRange();    //set min and max values for rt
-                    sample->fileName = filename.toStdString();
-
-                    if ( filename.contains("blank",Qt::CaseInsensitive))
```
shubhra-agrawal: @rish9511 This should be removed from the else block as well, then

rish9511: @shubhra-agrawal 
Thanks for pointing out

Issue #723, opened 2018-06-01
slowdown with large peak table


chubukov:
I did an untargeted peak picking with a very low threshold, so I ended up with 50k peaks in a table.

I found that as a result of this, interaction with the sample widget (checking/unchecking samples) slowed the app to a crawl (on the order of a minute to check/uncheck a sample). This was with only about 30 samples loaded.

Is there some recalculation or update being done on the entire peak table? Is it obvious this needs to happen?

Not a huge deal, but I found it odd.

saifulbkhan:
This has been investigated, and culprit found (alignment widget, see #809 for discussion). While the existence of current alignment widgets and their behaviour will be discussed eventually, they should no longer slow down the application when samples are selected with a large peak table present (at least not when they are hidden).

PR #809, opened 2018-07-24
Fix sample widget slowdown with large peak table

saifulbkhan:
Fixing an issue where having a large peak table would cause the
sample widget, when interacted with, to perform a heavy (blocking)
computation within the alignment visualisation widget(s).

Now this computation will only be performed if the alignment widget
is visible, and replotted whenever it is made visible using the
sidebar widget.

Issue: #723


chubukov:
@saifulbkhan sounds great. I think it's also worth asking if it's worth keeping the widget as is at all. If it's not that useful but has the potential to really slow down the app (a user may have it visible but not actually be actively using it) it might be worth rethinking.

shubhra-agrawal:
@chubukov Agreed. We're trying to clean up the UI a bit and the topic of 3 different alignment visualisations did come up. Will need some user input before we decide anything. For v0.4.1, we're just making minor changes in the main window but we'll get to this soon.

Issue #795, opened 2018-07-16
Slowdown with navigating between multiple large peak tables


46navdeep:
**Describe the bug**
There is a slowdown when navigating between multiple large peak tables.

**To Reproduce**
Create multiple large peak tables.
Try navigating between two compounds on different tables.

**Expected behavior**

**Screenshots**

**System (please complete the following information):**
 - MacOs,Windows,Ubuntu
 - v 0.4.0

**Additional context**

PR #812, opened 2018-07-26
Fix slowdown peaks

46navdeep:
Issue Number: #795
Reasons that led to slow down during navigating between large peak tables:
1) A new comparison between compound item list and group list was triggered every time the user switched between the peaks. This leads to a complexity of O(n^2)  when navigating between the peaks.

Solution:
1) Creating a hash to store the list of compounds. So now searching a group item in the compound list takes O(1), thus reducing the complexity of the operation to O(n).

### DIFF
--- a/src/gui/mzroll/ligandwidget.cpp
+++ b/src/gui/mzroll/ligandwidget.cpp
@@ -411,37 +411,31 @@ void LigandWidget::showTable() {
     
     Hash.clear();
     QTreeWidgetItemIterator itr(treeWidget);
-    cerr<<"-------------------------Creating new hash table-----------------------------"<<"\n";
-    while(*itr)
-    {     
-    QTreeWidgetItem* item =(*itr);
-    QVariant v = item->data(0,Qt::UserRole);
-    Compound*  c =  v.value<Compound*>();                                                                                   
-    Hash.insert(c , item);
-     ++itr;
-    }
-
+        while(*itr)
+        {     
+        QTreeWidgetItem* item =(*itr);
+        QVariant v = item->data(0,Qt::UserRole);
+        Compound*  c =  v.value<Compound*>();                                                                                   
+        Hash.insert(c , item);
+        ++itr;
+        }
 treeWidget->setSortingEnabled(true);
 }
 
 void LigandWidget::markAsDone(Compound* compound) {
+    if(compound != NULL)
+    {
+        QHash<Compound *, QTreeWidgetItem *>::const_iterator i = Hash.find(compound);
+        if (i != Hash.end() & i.key() == compound) {
+            QTreeWidgetItem* item = i.value();  
 
-
-if(compound != NULL)
-{
-    QHash<Compound *, QTreeWidgetItem *>::const_iterator i = Hash.find(compound);
-if (i != Hash.end() & i.key() == compound) {
-     QTreeWidgetItem* item = i.value();  
-
-        if (item != NULL) {
-             
-            for (int col = 0; col < treeWidget->columnCount(); col++) {
-                item->setBackgroundColor(col, QColor(61, 204, 85, 100));
-            }
+                if (item != NULL) {
+                    for (int col = 0; col < treeWidget->columnCount(); col++) {
+                        item->setBackgroundColor(col, QColor(61, 204, 85, 100));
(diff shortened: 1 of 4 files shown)


src/gui/mzroll/ligandwidget.cpp
```diff
@@ -403,25 +404,38 @@ void LigandWidget::showTable() {
...
-            }
+    if(compound != NULL)
+    {
+        QHash<Compound *, QTreeWidgetItem *>::const_iterator i = Hash.find(compound);
+        if (i != Hash.end() & i.key() == compound) {
+            QTreeWidgetItem* item = i.value();  
+
+                if (item != NULL) {
+                    for (int col = 0; col < treeWidget->columnCount(); col++) {
+                        item->setBackgroundColor(col, QColor(61, 204, 85, 100));
```
rish9511: use https://doc.qt.io/qt-5/qtreewidgetitem.html#setBackground 
setBackgroundColor is [obsolete](https://doc.qt.io/archives/qt-4.8/qtreewidgetitem-obsolete.html)

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -403,25 +404,38 @@ void LigandWidget::showTable() {
...
-            for (int col = 0; col < treeWidget->columnCount(); col++) {
-                item->setBackgroundColor(col, QColor(61, 204, 85, 100));
-            }
+    if(compound != NULL)
+    {
+        QHash<Compound *, QTreeWidgetItem *>::const_iterator i = Hash.find(compound);
+        if (i != Hash.end() & i.key() == compound) {
+            QTreeWidgetItem* item = i.value();  
+
+                if (item != NULL) {
```
rish9511: use [nullptr](https://en.cppreference.com/w/cpp/keyword/nullptr)

src/gui/mzroll/ligandwidget.h
```diff
@@ -112,6 +113,7 @@ private Q_SLOTS:
     QToolButton *loadButton;
     QLineEdit*  filterEditor;
     QPoint dragStartPosition;
+    QHash<Compound *, QTreeWidgetItem *> Hash;
```
rish9511: use a better name

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -403,25 +404,38 @@ void LigandWidget::showTable() {
...
-
             }*/
 
     }
-    treeWidget->setSortingEnabled(true);
+  
+    
+    CompoundsHash.clear();
+    QTreeWidgetItemIterator itr(treeWidget);
+        while(*itr)
```
shubhra-agrawal: @46navdeep Please correct the indentation of the whole block.

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -403,25 +404,38 @@ void LigandWidget::showTable() {
...
 void LigandWidget::markAsDone(Compound* compound) {
-
-    if (compound != NULL) {
-        QTreeWidgetItem* item = getItem(compound);
-
-        if (item != NULL) {
-            for (int col = 0; col < treeWidget->columnCount(); col++) {
-                item->setBackgroundColor(col, QColor(61, 204, 85, 100));
-            }
+    if(compound != NULL)
```
shubhra-agrawal: @46navdeep I would suggest moving the code out of the 'if' block and just returning early.

Read more here: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -403,25 +404,38 @@ void LigandWidget::showTable() {
                 QTreeWidgetItem *item = new QTreeWidgetItem(parent, PathwayType);
                 item->setText(0,QString(compound->category[i].c_str()));
                 //parent->setData(0,Qt::UserRole,QVariant::fromValue(QString(pathway_id.c_str())));
-
             }*/
 
     }
-    treeWidget->setSortingEnabled(true);
+
```
shubhra-agrawal: @46navdeep Move this into another function, perhaps?

PR #815, opened 2018-07-26
UI: Hide unnecessary fields

saifulbkhan:
The fields "Ratio Change" and "P-value" are not required within regular peak tables and bookmarks table. These fields should be only visible when the table instance is showing scatter-plot groups. This PR addresses this issue.

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1,6 +1,6 @@
 #include "tabledockwidget.h";
 
-TableDockWidget::TableDockWidget(MainWindow* mw, QString title, int numColms, int bookmarkFlag) {
+TableDockWidget::TableDockWidget(MainWindow* mw, QString title, int numColms, tableType type) {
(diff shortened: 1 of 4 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -302,6 +303,23 @@ void TableDockWidget::setupPeakTable() {
     treeWidget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);
     treeWidget->header()->adjustSize();
 
+    /* Remove "Ratio Change" and "P-value" fields from non-scatterplot
+     * tables where they are not relevant.
+     */
+    if (viewType == groupView && type != scatterplotTable) {
+        QRegExp rx1("Ratio Change");
+        QRegExp rx2("P-value");
+        treeWidget->setColumnHidden(colNames.indexOf(rx1), true);
```
rish9511: QRegExp is not really needed. IndexOf() can directly take string

PR #818, opened 2018-07-30
UI: Club export options under one popover button

saifulbkhan:
Export buttons should not be occupying extra space in the toolbar
for each option available. This commit groups them under one drop-
down menu, freeing up space, and hence one step closer to a cleaner
EIC toolbar.


src/gui/mzroll/mainwindow.cpp
```diff
@@ -3249,42 +3239,30 @@ QWidget* MainWindowWidgetAction::createWidget(QWidget *parent) {
...
+		
+		QToolButton* btnExport = new QToolButton(parent);
+		btnExport->setIcon(QIcon(rsrcPath + "/exportOptions.png"));
+		btnExport->setToolTip("Export as…");
+		QMenu* btnExportMenu = new QMenu("Export Image as…");
+		btnExport->setMenu(btnExportMenu);
+		btnExport->setPopupMode(QToolButton::InstantPopup);
+		QAction* pdfAction = btnExportMenu->addAction(QIcon(rsrcPath + "/exportpdf.png"),
+													  "Export EIC as PDF file");
+		connect(pdfAction, SIGNAL(triggered(bool)), mw, SLOT(exportPDF()));
```
rish9511: lets try to use the new  syntax of signals and slots wherever possible.  In Qt5 the same will look something like this  `connect(pdfAction, &QAction::triggered, mw, &MainWindow::exportPdf)`. 

One of the pros is that compile-time check is enabled for the new syntax. Read more [here](https://wiki.qt.io/New_Signal_Slot_Syntax)

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3249,42 +3239,30 @@ QWidget* MainWindowWidgetAction::createWidget(QWidget *parent) {
...
-		return btnPNG;
-
-	}
-	else if (btnName == "btnPrint") {
-
-		QToolButton *btnPrint = new QToolButton(parent);
-		btnPrint->setIcon(QIcon(rsrcPath + "/fileprint.png"));
-		btnPrint->setToolButtonStyle(Qt::ToolButtonTextUnderIcon);
-		btnPrint->setToolTip(tr("Print EIC (Ctr+P)"));
-		btnPrint->setShortcut(tr("Ctrl+P"));
```
rish9511: You forgot to add this. The shortcut should still remain

src/gui/mzroll/mainwindow.cpp
saifulbkhan: Done.

src/gui/mzroll/mainwindow.cpp
saifulbkhan: Sorry about that. Fixed.

PR #834, opened 2018-08-07
Testpolly installer bootcamp

Ky5595:
This PR has the following features -

- Adding Polly emailer in Elmaven now.
 - - previously, users had an option to send email using any email id and password. Users did not prefer this, because it required exposing there gmail password. So now, an email (with redirection button) from polly will be sent every time users have successfully uploaded data to polly. 
- Filtering based on timestamp -
 - - Elmaven now provides timestamp for one particular upload. this is needed on polly to filter only the files corresponding to the given upload.
- providing an option to upload sample cohort file from CLI itself.
- changed the way users provide polly project to CLI - 
 - - initially we were taking polly project from the credentials file, now users can specify their project using -N argument. If this project exists on polly, EPI will upload to that project else it will create the project first and then upload to it.
- validating the cohort file on Elmaven layer - 
 - - CLI checks if the cohort file provided by the user has exact same samples that were loaded in elmaven or not. Also, the cohort file should not have more than 9 cohorts.
- redirection url changes -
 - - if the user has not provided cohort file or if the file is incorrect, elmaven will redirect to gsheet interface otherwise it will directly redirect to visualisation dashboard where users can get insights from their mass spec data.

### DIFF
--- a/src/cli/peakdetector/PeakDetectorCLI.cpp
+++ b/src/cli/peakdetector/PeakDetectorCLI.cpp
@@ -691,7 +691,7 @@ void PeakDetectorCLI::writeReport(string setName,QString jsPath,QString nodePath
 				// jspath and nodepath are very important here..node executable will be used to connect to polly, with the help of index.js script..
 				QString upload_project_id = UploadToPolly(jsPath,nodePath,files_to_be_uploaded,creds,pollyProject); 
 				if (upload_project_id!=""){ //That means the upload was successfull, in that case, redirect the user to polly..
-					QString redirection_url = QString("<a href='http://testpolly.elucidata.io/main#project=%1&auto-redirect=%2&elmavenTimestamp=%3'>Go To Polly</a>").arg(upload_project_id).arg(redirect_to).arg(datetimestamp);
+					QString redirection_url = QString("<a href='https://polly.elucidata.io/main#project=%1&auto-redirect=%2&elmavenTimestamp=%3'>Go To Polly</a>").arg(upload_project_id).arg(redirect_to).arg(datetimestamp);
 					qDebug()<<"redirection url - \n"<<redirection_url;
 					QString filename=QStandardPaths::writableLocation(QStandardPaths::QStandardPaths::GenericConfigLocation) + QDir::separator()+datetimestamp+"_redirection_url.txt";
 					qDebug()<<"writing url to this file -"<<filename;
--- a/src/cli/peakdetector/PeakDetectorCLI.cpp
+++ b/src/cli/peakdetector/PeakDetectorCLI.cpp
@@ -788,10 +788,11 @@ QString PeakDetectorCLI::UploadToPolly(QString jsPath,QString nodePath,QStringLi
 			}
 		}
 	if (projectId==""){
-		//In case no project matches with the user defined name or the user has not provided any project name,
-		// upload to default project..
-		_pollyIntegration->exportData(filenames,defaultprojectId);
-		upload_project_id = defaultprojectId;
+		//In case no project matches with the user defined name,
+		// Create the project and upload to it..This makes the project name to be mandatory..
+		QString new_project_id = _pollyIntegration->createProjectOnPolly(creds["polly_project"]);
+		_pollyIntegration->exportData(filenames,new_project_id);
+		upload_project_id = new_project_id;
 	}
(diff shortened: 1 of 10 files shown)


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -144,6 +146,20 @@ void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
...
+			{
+			pollyProject = QString(optarg);
+			pollyProject_provided = true;
+			}
+			break;
+		case 'S':
+			//Upload sample cohort file from the CLI to polly.
+			{
+			sample_cohort_file = QString(optarg);
+			sample_cohort_present  =true;
```
shubhra-agrawal: @Ky5595 Why not check for empty string instead of using a global boolean variable? Is there some corner case here?

Same for case 'P' and 'N' too.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -627,58 +671,84 @@ void PeakDetectorCLI::writeReport(string setName,QString jsPath,QString nodePath
...
+													sample_cohort_filename+".csv";
+
+			//jspath and nodepath are very important here..node executable will be used to connect to polly, with the help of index.js script..
+			QString upload_project_id = UploadToPolly(jsPath,nodePath,files_to_be_uploaded,creds,pollyProject); 
+			if (upload_project_id!="") { //That means the upload was successful, in that case, redirect the user to polly..
+				QString redirection_url = QString("<a href='https://polly.elucidata.io/main#project=%1&auto-redirect=%2&elmavenTimestamp=%3'>Go To Polly</a>").arg(upload_project_id).arg(redirect_to).arg(datetimestamp);
+				qDebug() << "redirection url - \n" << redirection_url;
+				QString filename=QStandardPaths::writableLocation(QStandardPaths::QStandardPaths::GenericConfigLocation) + QDir::separator()+datetimestamp+"_redirection_url.txt";
+				qDebug() << "writing url to this file -" << filename;
+				QFile file(filename);// redirection url will be written to a file..In future it will be replaced by email feature..
```
shubhra-agrawal: @Ky5595 The email feature has been added. Can this be removed now?

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -717,105 +848,83 @@ QMap<QString, QString> PeakDetectorCLI::readCredentialsFromXml(QString filename)
...
-    process.waitForFinished(-1);
-    QByteArray result = process.readAllStandardOutput();
-    QByteArray result2 = process.readAllStandardError();
-    qDebug()<<"pyemailer output  - "<<result;
-    qDebug()<<"pyemailer error, if any  - "<<result2;
-	status=true;
-	if (result2!=""){
-		status=false;
-	}
+	QString user_email = creds["polly_username"];
```
shubhra-agrawal: @Ky5595 jsPath is not required in the function anymore.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -627,57 +671,92 @@ void PeakDetectorCLI::writeReport(string setName,QString jsPath,QString nodePath
...
+				QString redirection_url = QString("<a href='https://polly.elucidata.io/main#project=%1&auto-redirect=%2&elmavenTimestamp=%3'>Go To Polly</a>").arg(upload_project_id).arg(redirect_to).arg(datetimestamp);
+				qDebug() << "redirection url - \n" << redirection_url;
+				QString filename = QStandardPaths::writableLocation(
+									QStandardPaths::QStandardPaths::GenericConfigLocation) + 
+									QDir::separator() + 
+									datetimestamp + 
+									"_redirection_url.txt";
+				
+				qDebug() << "writing url to this file -" << filename;
+				//redirection url will be written to a file..In future it will be replaced by email feature..
```
shubhra-agrawal: @Ky5595 Email feature has been added. Can we remove the file now?

Ky5595: @shubhra-agrawal 
In case users don't get an email, we still want to have some source from where they can be redirected to polly.
So, I believe we should retain this functionality.

PR #848, opened 2018-08-22
Cli obiwarp

rish9511:
Allow users to choose  alignment algorithm  from command line

### DIFF
--- a/src/cli/peakdetector/PeakDetectorCLI.cpp
+++ b/src/cli/peakdetector/PeakDetectorCLI.cpp
@@ -3,13 +3,14 @@
 PeakDetectorCLI::PeakDetectorCLI() {
 	parseOptions = new ParseOptions();
 	_pollyIntegration = new PollyIntegration();
+    alignType = NO_TYPE;
 }
 
 void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
 
 	//command line options
 	const char * optv[] = {
-							"a?alignSamples: Enter non-zero integer to run alignment <int>",
+                            "a?alignSamples: 1=obi-warp, 2=poly fit",
 							"b?minGoodGroupCount: Enter minimum number of good peaks per group <int>",
 							"c?matchRtFlag: Enter non-zero integer to match retention time to the database values <int>",
 							"C?compoundPPMWindow: Enter ppm window for m/z <float>",							
@@ -52,7 +53,15 @@ void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
 
 		case 'a':
 			mavenParameters->alignSamplesFlag = true;
-			if (atoi(optarg) == 0) mavenParameters->alignSamplesFlag = false;
+
+            if (atoi(optarg) == 1)
+                alignType = OBI_WARP;
+
+            else if (atoi(optarg) == 2)
+                alignType = POLY_FIT;
+
+            else
+                mavenParameters->alignSamplesFlag = false;
 			break;
 
 		case 'b':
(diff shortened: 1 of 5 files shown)


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -3,13 +3,14 @@
...
 	_pollyIntegration = new PollyIntegration();
+    alignType = NO_TYPE;
 }
 
 void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
 
 	//command line options
 	const char * optv[] = {
-							"a?alignSamples: Enter non-zero integer to run alignment <int>",
+                            "a?alignSamples: 1=obi-warp, 2=poly fit",
```
shubhra-agrawal: "Enter 1 for obi-warp alignment, 2 for polyfit"
Consistent with the other help messages.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -50,8 +51,18 @@ void PeakDetectorCLI::processOptions(int argc, char* argv[]) {
...
+
+            if (atoi(optarg) == 1)
+                alignType = OBI_WARP;
+
+            else if (atoi(optarg) == 2)
+                alignType = POLY_FIT;
+
+            else
+                alignType = NO_TYPE;
+                mavenParameters->alignSamplesFlag = false;
```
shubhra-agrawal: @rish9511 won't the flag always be set as false? else block needs to be within brackets

src/core/libmaven/PeakDetector.cpp
```diff
@@ -253,24 +253,43 @@ vector<mzSlice*> PeakDetector::processCompounds(vector<Compound*> set,
...
-                mavenParameters->writeCSVFlag = true;
+void PeakDetector::alignSamples(const int& method) {
+
+    // only called from CLI
+    if(mavenParameters->samples.size() > 1 ) {
+
+        switch(method) {
+
+        case 1: {
+            ObiParams params("cor", false, 2.0, 1.0, 0.20, 3.40, 0.0, 20.0, false, 0.60);
```
shubhra-agrawal: @rish9511 add a print statement for case 1 and 2. Important to let the user know what's happening in CLI. Especially because Obi-warp takes time.

shubhra-agrawal: @rish9511 where are these arguments coming from?

shubhra-agrawal: I'm not too comfortable hard-coding these values here. Might update the defaults in the ui file because of some issue and forget to update them in this function. But I can't see a good way to solve this either. Your call.

rish9511: @shubhra-agrawal 
There are  print statements in the alignment algos somewhere. 
In case of obi-warp cli shows `Aligning Sample Retention times...` in the terminal
Do you think there is still a need for print statements ?

src/cli/peakdetector/PeakDetectorCLI.cpp
rish9511: @shubhra-agrawal 
Yes, this needs to be within brackets. 
Forgot somehow

src/cli/peakdetector/PeakDetectorCLI.cpp
rish9511: yes, makes sense

rish9511:
@shubhra-agrawal 
Have pushed the relevant changes
Please check again

src/core/libmaven/PeakDetector.cpp
shubhra-agrawal: @rish9511 Just make sure those are not debugging statements. User should know alignment is happening, that's all.

rish9511: Just by looking at it, default constructor seems to be the best place to define such values
But again, i have not thought about this well enough

PR #849, opened 2018-08-23
UI peak table deletion

saifulbkhan:
With this PR, the user will be able to delete peak tables if they do not need them during a session.

### DIFF
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -873,6 +873,14 @@ void EicWidget::clearPlot() {
 	scene()->setSceneRect(10, 10, this->width() - 10, this->height() - 10);
 }
 
+void EicWidget::unSetPeakTableGroup(PeakGroup* group)
+{
+    // Peak table is being deleted. Making sure the selected group is not holding any garbage value;
+    if(eicParameters->selectedGroup ==  group) {
+        eicParameters->selectedGroup = nullptr;
+    }
+}
+
 void EicWidget::replot(PeakGroup* group) {
 	//qDebug <<" EicWidget::replot(PeakGroup* group )";
 
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -878,7 +878,10 @@ void EicWidget::unSetPeakTableGroup(PeakGroup* group)
     // Peak table is being deleted. Making sure the selected group is not holding any garbage value;
     if(eicParameters->selectedGroup ==  group) {
         eicParameters->selectedGroup = nullptr;
-    }
+        mzSlice slice(0, 0, 0, 0);
+        setMzSlice(slice);
+
+	}
 }
 
 void EicWidget::replot(PeakGroup* group) {
@@ -893,7 +896,7 @@ void EicWidget::replot(PeakGroup* group) {
 	setupColors();	  //associate color with each EIC
 
 	clearPlot();
-	
+
 	setSelectedGroup(group);
 	setTitle();
 	addEICLines(_showSpline, _showEIC);
@@ -1455,12 +1458,6 @@ void EicWidget::setMzSlice(const mzSlice& slice) {
 		eicParameters->_slice = slice;
 	}
 	replot(NULL);
-}
-
-void EicWidget::setMzRtWindow(float mzmin, float mzmax, float rtmin,
-		float rtmax) {
-	mzSlice slice(mzmin, mzmax, rtmin, rtmax);
-	setMzSlice(slice);
 }
 
 void EicWidget::setPeakGroup(PeakGroup* group) {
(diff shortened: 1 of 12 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -2424,16 +2425,26 @@ PeakTableDockWidget::PeakTableDockWidget(MainWindow *mw) : TableDockWidget(mw) {
   toolBar->addAction(btnX);
 
   setTitleBarWidget(toolBar);
+
+  deletionDialog = new PeakTableDeletionDialog(this);
+  connect(deletionDialog, SIGNAL(accepted()),
+          this, SLOT(destroy()));
+  connect(deletionDialog, SIGNAL(rejected()),
+          deletionDialog, SLOT(hide()));
```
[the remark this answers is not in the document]
rish9511: >Maybe the "rejected" signal can be connected internally within PeakDetectionDialog, but "accepted" needs to be connected with the "destroy" slot of an instance of PeakTableDockWidget.

Well, the "accpeted" signal can still be connected to `PeakTableDockWidget::destroy()`  from `PeakTableDeletionDialog` class because  `PeakTableDockWidget` class is the parent of `PeakTableDeletionDialog` . 


> So I connected both the response signals in the same place -- looks better that way. I can of course move the "accepted" connection if you think it makes more sense to not make it explicit.

I suggested this because, to me  it makes more sense to have all the connections of a class defined in the ctor of that class. It makes easier to grep the code base. So, if I am looking for the all connections of a class I just have to check the ctor of that class. 
What do you think ? Maybe we should have some sort of **coding conventions** where we can define such things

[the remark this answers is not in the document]
saifulbkhan: Right! So while we do not have any coding conventions around this, I think your way makes sense. I had forgot that the dialog's parent is passed as a param, and is accessible from within. I'll make changes accordingly.

src/gui/mzroll/eicwidget.cpp
```diff
@@ -873,6 +873,15 @@ void EicWidget::clearPlot() {
 	scene()->setSceneRect(10, 10, this->width() - 10, this->height() - 10);
 }
 
+void EicWidget::unSetPeakTableGroup(PeakGroup* group)
+{
+    // Peak table is being deleted. Making sure the selected group is not holding any garbage value;
+    if(eicParameters->selectedGroup ==  group) {
+        eicParameters->selectedGroup = nullptr;
+		setMzRtWindow(0, 0, 0, 0);
```
shubhra-agrawal: @saifulbkhan setMzRtWindow is the same as setMzSlice. Let's remove this function, create a slice with (0,0,0,0) parameters and call setMzSlice here.

saifulbkhan: Alright. Will do.

src/gui/mzroll/mzroll.pro
```diff
@@ -1,289 +1,292 @@
...
-  images
- sources.path =  ./src
- target.path =  ../bin
-
-build_all {
-  !build_pass {
-   CONFIG -=    build_all
-  }
-}
+MOC_DIR=$$top_builddir/tmp/mzroll/
```
shubhra-agrawal: @saifulbkhan why is the whole file showing up in diff?

saifulbkhan: Strange. I only made additions to the SOURCES and HEADERS list. This should not be showing up.

src/gui/mzroll/scatterplot.cpp
```diff
@@ -121,9 +123,17 @@ void ScatterPlot::setupToolBar() {
 ScatterPlot::~ScatterPlot() { delete(compareSamplesDialog); }
 
 
-void ScatterPlot::setTable(TableDockWidget* t) { 
-	_table=t; 
-	compareSamplesDialog->setTableWidget(t); 
+void ScatterPlot::setTable(TableDockWidget* t) {
```
shubhra-agrawal: @saifulbkhan better variable name, please

saifulbkhan: Did you mean `TableDockWidget *t`? It can have a better name.

src/gui/mzroll/scatterplot.cpp
```diff
@@ -720,4 +728,8 @@ void ScatterPlot::showPeakTable() {
     _peakTable->showAllGroups();
     _peakTable->setVisible(btnPeakTable->isChecked());
 
+}
+
+void ScatterPlot::peakTableVisibilityChanged(bool visible) {
```
shubhra-agrawal: the signal could directly connect to btnPeakTable's setChecked, no?

saifulbkhan: Right. Should have checked if `setChecked` was a slot.

src/gui/mzroll/mzroll.pro
shubhra-agrawal: just make sure there are no other changes than what you intended. Cursory look didn't turn up anything problematic.

saifulbkhan: I do not know why the entire file would turn up in diff, but I am pretty sure that no unintended changes are being included. Here's the diff for the file (before vs after) using an online tool: http://linediff.com/?id=5b991604687f4bfd7e8b4567

PR #854, opened 2018-08-29
Uploading in separate thread

Ky5595:
This PR bring two major changes on table - 

- Persistent login to Polly, In El-maven

- Now, uploading to Polly will happen in a separate thread. This will ensure that UI doesn't  freeze and also uploading process can be terminated while it is running.

How to test all the changes in this PR -

- [x] Verify that Users only have to log in once, unless they log out.

- [x] verify that GUI doesn't get hanged while uploading.Also, multiple threads can be started in parallel.

TODO - 
implement keychain in EPI, As of now skipping it because users will need their admin password every time they access EPI.
We need it for better security.

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -52,9 +52,66 @@ function read_id_token(token_filename){
     return saved_token;    
 }
 
+function refreshToken(token_filename,email){
+    if (!email) {
+        console.log(chalk.red.bold("Email is required to refresh the token"));
+        return;
+    } else {
+        if (!(email.includes("@"))){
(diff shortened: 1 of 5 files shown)


shubhra-agrawal:
@Ky5595 works well for me!

Small UX issues that need to be handled now. Like, what if the user changes certain selections while the upload is in progress. Certain elements on the dialog might have to be disabled while uploading etc. Just mentioning that here so it doesn't get lost. We'll take care of that. Thanks @Ky5595 !

PR #856, opened 2018-08-31
Ftr fluxomics

shubhra-agrawal:
- Move sample cohort validation check to pollyIntegration as it is a polly-specific check and common to CLI and GUI
- New UI for Polly Dialog: added an app page for fluxomics and related functions
- A fluxomics prompt shows up on exporting all groups that opens the flux app when clicked
- Flux app uploads a sample cohort file to polly
- if validation of sample cohort file fails, the redirection url takes the user to Google sheets on polly not the pollyPhi workflow

Code for changing between apps in Polly is a little crude atm. Will refactor that later when adding finishing touches.


src/gui/mzroll/pollyelmaveninterface.h
```diff
@@ -142,6 +149,9 @@ class PollyElmavenInterfaceDialog : public QDialog, public Ui_PollyElmavenInterf
                  * 5. Populate all combo boxes on this UI, with the data obtained from above steps..
                  */
                 QVariantMap startup_data_load();
```
rish9511: This returns QVariantMap but it is not used anywhere. 
Can be changed to `void`

Thought it would be good if we keep cleaning things like these in parallel

src/gui/mzroll/mainwindow.h
```diff
@@ -100,6 +101,7 @@ class LogWidget;
 class SpectralHit;
 class SpectralHitsDockWidget;
 class PeptideFragmentationWidget;
+class Notificator;
```
rish9511: Well, either we need to include the file or we forward declare it. Both of them are not required
Since we do not require any info related to Notificator here, just including it in the cpp file will do the job

src/gui/mzroll/mainwindow.h
```diff
@@ -52,6 +52,7 @@
 //Added when merged with Maven776 - Kiran
 #include "remotespectrahandler.h"
 #include "messageBoxResize.h"
+#include "notificator.h"
```
rish9511: Not required here
Just include it in the cpp file

src/gui/mzroll/pollyelmaveninterface.h
shubhra-agrawal: projectnames_id was made a global parameter recently so yes. Can be changed to void. Will do it while refactoring.

rish9511:
Tested on: Linux
During my testing, I found that the application is crashing while trying to upload data to Polly.
It's the same crash as reported by @saifulbkhan 

Re-posting the backtrace and steps to reproduce here

**Steps to reproduce** 
1. Start El-MAVEN, log in to Polly, exit.
2. Start El-MAVEN again, load sample, do peak detection, and upload to either Fluxomics or FIrstView.

**Backtrace**: https://pastebin.com/raw/DUJwkNwC

shubhra-agrawal:
> Tested on: Linux
> During my testing, I found that the application is crashing while trying to upload data to Polly.
> It's the same crash as reported by @saifulbkhan
> 
> Re-posting the backtrace and steps to reproduce here
> 
> **Steps to reproduce**
> 
> 1. Start El-MAVEN, log in to Polly, exit.
> 2. Start El-MAVEN again, load sample, do peak detection, and upload to either Fluxomics or FIrstView.
> 
> **Backtrace**: https://pastebin.com/raw/DUJwkNwC

Can't reproduce this on develop. Will look into this asap

shubhra-agrawal:
@saifulbkhan @rish9511 Update:
- Crash has been resolved (happened due to rebasing after changes in another branch)
- UI changes are done. The notification always shows up at the top right of your screen. The decision was made by the product team after looking at other software notification systems.
- Close button has been added to the prompt to hide it
- Pompt only shows up for tables that have at least one labeled group
- Prompt shows up for all export options in a peak table

I'm refactoring this in a separate branch so reviewing is not a challenge. Would recommend testing the functionality in this branch and the code in the ref_fluxomics branch

PR #857, opened 2018-09-03
Fix missing C12 label

saifulbkhan:
This should fix missing "C12 PARENT" labels in the exported CSV, when PeakDetector CLI is run on Masha-1 (neg C12) dataset. After writing to the CSV file, a second pass is done to check which ones are missing labels. For these group's, if their meanMz is within the set ppm range around Compound's computed mz value, then they are given a C12 PARENT child that is a copy of the group itself. This ensures that at least for the groups passing this check, isotope label is not empty in the exported (again) CSV.

TL;DR - This is literally a patch, the underlying issue still exists.

### DIFF
--- a/src/cli/peakdetector/PeakDetectorCLI.cpp
+++ b/src/cli/peakdetector/PeakDetectorCLI.cpp
@@ -1093,6 +1093,59 @@ void PeakDetectorCLI::saveCSV(string setName) {
     }
     csvreports->closeFiles();
 
+	// NOTE: The following validation is being done to prevent a workflow breaking
+	// issue (SYP-24) caused by missing C12 PARENT labels for labelled data. This
+	// is only being done for the CLI.
+
+	// Check whether there's a missing C12 PARENT label in the written CSV file.
+    bool labeledDataPresent = false;
+    bool fileNeedsCorrection = false;
+    vector<int> groupsWithMissingLabels;
+    CsvParser *csvParser = CsvParser_new(fileName.c_str(), ",", 1);
+    CsvRow *currentRow = NULL;
+    // Iterates over each row, continuously assigning a pointer to an unread CsvRow.
+    while (currentRow = CsvParser_getRow(csvParser)) {
+        // eighth field is the isotope label.
+        string isotopeLabel = string(currentRow->fields_[7]);
+        if (isotopeLabel.empty()) {
+            int groupId = atoi(currentRow->fields_[2]);
+            groupsWithMissingLabels.push_back(groupId);
+			fileNeedsCorrection = true;
+        } else {
+			labeledDataPresent = true;
+        }
+    }
+
+	if (!labeledDataPresent) fileNeedsCorrection = false;
+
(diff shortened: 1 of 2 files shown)


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -1070,6 +1070,52 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
+    CsvRow *currentRow = NULL;
+    // Iterates over each row, continuously assigning a pointer to an unread CsvRow.
+    while (currentRow = CsvParser_getRow(csvParser)) {
+        // eighth field is the isotope label.
+        string isotopeLabel = string(currentRow->fields_[7]);
+        if (isotopeLabel.empty()) {
+            int groupId = atoi(currentRow->fields_[2]);
+            groupsWithMissingLabels.push_back(groupId);
+            atLeastOneIsotopeLabelIsEmpty = true;
+        } else {
```
shubhra-agrawal: @saifulbkhan both flags signify the same thing, no? as soon as you find a missing label, fileNeedsCorrection goes true. We don't even need the else block

shubhra-agrawal: actually, we don't need either of the flags. Just check for the groupsWithMissingLabels vector size. If there are elements then we need to correct the file.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -1070,6 +1070,52 @@ void PeakDetectorCLI::saveCSV(string setName) {
...
+			if (binary_search(groupsWithMissingLabels.begin(), groupsWithMissingLabels.end(), csvreports->groupId + 1)) {
+				PeakGroup newGroup = group;
+				Compound *compound = group.getCompound();
+				float compoundMz = MassCalculator::computeMass(compound->formula, mavenParameters->getCharge(compound));
+				float cutoffDist = massCutoffDist(group.meanMz, compoundMz, mavenParameters->massCutoffMerge);
+		    	if (cutoffDist < mavenParameters->massCutoffMerge->getMassCutoff()) {
+					PeakGroup childGroup = group;
+					childGroup.tagString = "C12 PARENT";
+					newGroup.children.push_back(childGroup);
+					csvreports->addGroup(&newGroup);
```
shubhra-agrawal: @saifulbkhan would suggest adding a comment here that groupId is updated in the addGroup function. Not necessary, just helps readability.

src/cli/peakdetector/PeakDetectorCLI.cpp
saifulbkhan: We will need two flags, albeit there is a simpler way of doing this. I'll make the changes as discussed.

src/cli/peakdetector/PeakDetectorCLI.cpp
saifulbkhan: Will add.

PR #860, opened 2018-09-05
Notify compounds DB file not provided for targeted analysis using CLI

saifulbkhan:
This commit adds a message to CLI output when user opts for
targeted peak detection but does not provide a compounds database
file. The program will also terminated before reading in any sample
files in this case.


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -533,6 +533,9 @@ void PeakDetectorCLI::loadCompoundsFile() {
 		int loadCount = DB.loadCompoundCSVFile(mavenParameters->ligandDbFilename);
 		mavenParameters->compounds = DB.compoundsDB;
 		cout << "Total Compounds Loaded : " << loadCount << endl;
+	} else if (!mavenParameters->processAllSlices) {
+		cerr << "\nPlease provide a compound database file to proceed with targeted analysis." << endl;
```
shubhra-agrawal: @saifulbkhan Also add that they can use the -h option to get help. User might not remember all parameters and their usage.

PR #863, opened 2018-09-13
Add explicit warning when compound database is empty

saifulbkhan:
If no compounds are loaded from the provided compound database using CLI, the user should be notified as such.

### DIFF
--- a/src/cli/peakdetector/PeakDetectorCLI.cpp
+++ b/src/cli/peakdetector/PeakDetectorCLI.cpp
@@ -555,7 +555,11 @@ void PeakDetectorCLI::loadCompoundsFile() {
 		cout << "\nLoading ligand database" << endl;
 		int loadCount = DB.loadCompoundCSVFile(mavenParameters->ligandDbFilename);
 		mavenParameters->compounds = DB.compoundsDB;
-		cout << "Total Compounds Loaded : " << loadCount << endl;
+		if (loadCount == 0) {
+			cerr << "Warning: Given compound database is empty!" << endl;
+		} else {
+			cout << "Total Compounds Loaded : " << loadCount << endl;
+		}
 	} else {
 		cerr << "\nPlease provide a compound database file to proceed with targeted analysis."
 		     << "Use the '-h' argument to see all available options." << endl;


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -533,6 +533,9 @@ void PeakDetectorCLI::loadCompoundsFile() {
 		int loadCount = DB.loadCompoundCSVFile(mavenParameters->ligandDbFilename);
 		mavenParameters->compounds = DB.compoundsDB;
 		cout << "Total Compounds Loaded : " << loadCount << endl;
+		if (loadCount == 0) {
```
Ky5595: It should be like this - 


if (loadCount == 0) {
			cerr << "Warning: Uploaded compound database is empty!" << endl;
		}
		else{
			cout << "Total Compounds Loaded : " << loadCount << endl;
		}

shubhra-agrawal: @saifulbkhan While we're at it, also remove the word 'uploaded'. We're not uploading the data anywhere.

src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -533,6 +533,9 @@ void PeakDetectorCLI::loadCompoundsFile() {
 		int loadCount = DB.loadCompoundCSVFile(mavenParameters->ligandDbFilename);
```
Ky5595: @saifulbkhan 
Have you tested this change in different scenarios ?
In some cases, the CLI will crash before reaching your changes, while executing the 2 lines below.
You should make sure that this piece of code doesn't crash.

saifulbkhan: I have tested this for two cases:
1. When CSV file has only headers and no entries.
2. When CSV file is completely empty.

There is a scenario where, if the CSV file has blank spaces and newlines, our parser crashes the application. But this is an issue that is not related to this PR. We need to make some changes to our parser to catch these edge cases.

PR #867, opened 2018-09-24
Warn when compounds are discarded while loading

saifulbkhan:
This pair of commit makes changes on both CLI and GUI side to ensure that the user is informed about any compounds discarded while loading a compounds DB file.

Tested the changes with:
- DB having rows with missing m/z and formula.
- DB having rows with missing name, m/z and formula.

Got expected behavior on both GUI and CLI.

### DIFF
--- a/src/core/libmaven/databases.cpp
+++ b/src/core/libmaven/databases.cpp
@@ -10,6 +10,9 @@ int Databases::loadCompoundCSVFile(string filename) {
     map<string, int> header;
     vector<string> headers;
 
+    // reset the contents of the vector containing the names of invalid rows
+    invalidRows.clear();
+
     //assume that files are tab delimited, unless matched ".csv", then comma delimited
     string sep="\t";
     if(filename.find(".csv") != -1 || filename.find(".CSV") != -1) sep=",";
@@ -138,6 +141,10 @@ Compound* Databases::extractCompoundfromEachLine(vector<string>& fields, map<str
         
         return compound;
     }
+
+    if (!name.empty())
+        id = name;
+    invalidRows.push_back(id);
 
     return NULL;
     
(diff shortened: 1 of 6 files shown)


src/core/libmaven/databases.cpp
```diff
@@ -10,6 +10,9 @@ int Databases::loadCompoundCSVFile(string filename) {
     map<string, int> header;
     vector<string> headers;
 
+    // recent the contents of the vector containing the names of invalid rows
```
shubhra-agrawal: @saifulbkhan "recent". Typo?

saifulbkhan: Ehem. I think I meant to write "reset" maybe. Will correct.

src/gui/mzroll/database.cpp
```diff
@@ -570,6 +571,10 @@ int Database::loadCompoundCSVFile(string filename){
             for(int i=0; i < categorylist.size(); i++) compound->category.push_back(categorylist[i]);
             if (addCompound(compound))
                 loadCount++;
+        } else {
+            if (!name.empty())
```
shubhra-agrawal: @saifulbkhan why check for name not being empty? id will be set to name if name is available anyway.

saifulbkhan: So usually ID is something like "cmpd21" or something, but if the name is empty it will become an empty string and the list of discarded compounds will look like: 

- UMP
- ATP
- 
- Succinate

I thought giving a compound ID would make more sense at least than letting it be blank.

shubhra-agrawal: @saifulbkhan Got it!

saifulbkhan:
Yes let's merge them. I'll create an issue.

src/gui/mzroll/database.cpp
saifulbkhan: We will soon work on merging these two database files. I recently opened an [issue](https://github.com/ElucidataInc/ElMaven/issues/869) for this.

PR #870, opened 2018-09-26
Update GUI for Polly changes

shubhra-agrawal:
In this PR, I have:
- Changed the datetimestamp to numeric values (same as CLI) to remove dependency on the user's system language preference
- Updated the redirection URLs for fluxomics (same as CLI)


src/gui/mzroll/loginform.cpp
```diff
@@ -57,40 +58,39 @@ void LoginForm::on_pushButton_clicked()
...
-void LoginForm::handleResults(QStringList results){
-    QString status_inside=results.at(0);
-    QString username=results.at(1);
-    QString password=results.at(2);
-    if (status_inside=="ok"){
-        qDebug()<<"Logged in, moving on now....";
+void LoginForm::handleResults(QStringList results)
+{
+    QString status_inside = results.at(0);
+    QString username = results.at(1);
```
Ky5595: @shubhra-agrawal 
Neither username nor password now needs to be stored. Please remove them.

src/gui/mzroll/loginform.cpp
```diff
@@ -57,40 +58,39 @@ void LoginForm::on_pushButton_clicked()
...
+    QString password = results.at(2);
+    if (status_inside == "ok") {
+        qDebug() << "Logged in, moving on now....";
         ui->login_label->setText("Fetching data from Polly..");
         QCoreApplication::processEvents();
         _pollyelmaveninterfacedialog->credentials = QStringList()<< username << password;
```
Ky5595: @shubhra-agrawal 
We don't need to store credentials in this pointer now. Initially we were storing these to achieve persistent login in one session but now we have persistent login functionality till the users explicitly click on "logout" button.
Please remove this line as it imposes security risk of exposing user credentials.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -210,29 +215,28 @@ void PollyElmavenInterfaceDialog::call_initial_EPI_form() {
...
+        _loadingDialog->statusLabel->setText("Authentication failed. Please login again.");
+        QCoreApplication::processEvents();
+        logout();
     }
 }
 
-void PollyElmavenInterfaceDialog::handleResults(QVariantMap projectnamesIdMap){
+void PollyElmavenInterfaceDialog::handleResults(QVariantMap projectnamesIdMap)
+{
     projectNamesId = projectnamesIdMap;
```
Ky5595: @shubhra-agrawal 
In case of No internet connection, projectNamesId will be an empty QMap().
We should check for it in later steps, I mean in the startupDataLoad function.
Otherwise EPI will not be able to detect internet connection issues which is what is happening right now, as reported by @rish9511 and @saifulbkhan .

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -210,29 +215,28 @@ void PollyElmavenInterfaceDialog::call_initial_EPI_form() {
...
     startupDataLoad();
```
Ky5595: @shubhra-agrawal 
This function must have at least one argument projectNamesId, otherwise how are we going to pass information generated from the thread to main UI ?

It should be startupDataLoad(projectNamesId).

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -277,9 +281,9 @@ void PollyElmavenInterfaceDialog::startupDataLoad()
     }
```
Ky5595: @shubhra-agrawal 

First of all, this function should be startupDataLoad(QVariantMap projectNamesId)
second, 
If projectNamesId.isEmpty(), then we have no internet connection issue and the same should be communicated to the user instead of trying to get user projects again.

This will resolve comments made by @rish9511 and @saifulbkhan regarding EPI returning "ok" even if there is no internet connection.

By the way, this was happening because -

- Once anyone logs into Elmaven, it will store the token information for persistent login.

- Now, while authenticating it will check for token expiry. If someone tries to use EPI before one hour (token expiry time), it will return "ok" because actually it has all the authentication information to connect to Polly.

src/gui/mzroll/loginform.cpp
```diff
@@ -31,9 +31,10 @@ WorkerThread::WorkerThread()
     
 };
 
-void WorkerThread::run(){
-    QString status_inside = _pollyintegration->authenticateLogin(username,password);
-    emit resultReady(QStringList()<<status_inside<<username<<password);
+void WorkerThread::run()
+{
+    QString status_inside = _pollyintegration->authenticateLogin(username, password);
+    emit resultReady(QStringList() << status_inside << username <<password);
```
shubhra-agrawal: @Ky5595 Don't need to add username and password to this qstringlist either right? It can just say status_inside?

src/gui/mzroll/pollyelmaveninterface.cpp
shubhra-agrawal: @Ky5595 The problem with it always sending 'ok' should be resolved by this. Making those changes.
The error reported by @rish9511 and @saifulbkhan happened further ahead on Polly. Will relay that to you on personal chat.

src/gui/mzroll/pollyelmaveninterface.cpp
shubhra-agrawal: @Ky5595 Empty QMap of project list is not a good indicator of no internet connection. The authentication function should check for this and return an error.

shubhra-agrawal:
@Ky5595 As discussed offline, the checks for internet connection are being made in another branch. I have resolved the 2 comments related to removing the credentials variable. Please take a look.

Issue #589, opened 2018-02-07
No compound name in Peak Detailed Format in case of automated peak detection


sunil20dhakad:
I am running untargeted metabolomics dataset where I am using Automated Peak Detection and saving the output in Peak Detailed Format (*.csv) but it does not describe groups, the 'compound' column is blank. This makes difficult to distinguish between groups.

shubhra-agrawal:
@sunil20dhakad The group name displayed in the Peak Table will now be available under the compound name column in CSV exports (both Peak detailed and Group summary formats). This is available on develop.

saifulbkhan:
Since this is not yet available in the current 0.4.1 release (probably broke somewhere in between the intermediate releases), we were thinking (re)enabling it.

The argument in favor of exporting group name (which is a combination of mz and rt values for untargeted workflow), is that all information visible in groups table, _should_ be exported out to CSV. Apparently it also helps to identify different groups downstream.

The argument not in favor of this would be, that exporting group name as `compound` field values does not make sense, and `compoundId` should be unique and mz@rt values may not be unique for different groups. We could add another field for storing these values but that could cause conflicts with existing downstream scripts either way.

I am re-opening this issue so that users can discuss the usefulness of exporting group name in CSV format. We are open to suggestions (please elaborate on your opinion).

sunil20dhakad:
@saifulbkhan 
As per I know  there is no case where compoundId column is blank in El-Maven output whether you provide id in compound database or not. If id is absent in compound database then it is filled with compound name, so in my opinion compoundId should also be mz@rt as compound (as @chubukov suggested).
@shubhra-agrawal FYI

shubhra-agrawal:
@chubukov @sunil20dhakad right. compoundId is never blank in El-MAVEN. It uses the compound name or just serial number if nothing is available. 

We'll implement the following changes then.
compound and compoundId : mz@rt
groupId : unique for every group.

saifulbkhan:
An {mz}@{rt} pair will now be exported for `compoundId` and `compound` fields while exporting groups of untargeted curation. Will be available from v0.5

PR #871, opened 2018-09-27
Add mz@rt strings for untargeted group and peak exports

saifulbkhan:
Instead of leaving the "compound" and "compoundId" fields blank
when exporting groups from an untargeted search, an {mz}@{rt}
string will now be exported.

Issue: #589


shubhra-agrawal:
@saifulbkhan Can you add some test cases around this? This keeps breaking from time-to-time.

src/core/libmaven/csvreports.cpp
```diff
@@ -359,6 +364,12 @@ void CSVReports::writePeakInfo(PeakGroup* group) {
         compoundName = sanitizeString(group->compound->name.c_str()).toStdString();
         compoundID   = sanitizeString(group->compound->id.c_str()).toStdString();
         formula = sanitizeString(group->compound->formula.c_str()).toStdString();
+    } else {
+        // absence of a group compound means this group was created using untargeted detection,
+        // we set compound name and ID to {mz}@{rt} strings for untargeted sets.
+        cout << "here" << endl;
```
rish9511: Not needed I guess

saifulbkhan: My bad. Was using this for debugging. Will remove

PR #878, opened 2018-09-30
Fix internet connection issue

Ky5595:


src/cli/peakdetector/PeakDetectorCLI.cpp
```diff
@@ -969,7 +969,12 @@ QString PeakDetectorCLI::UploadToPolly(QString jsPath, QString nodePath,
 
 	QString status = _pollyIntegration->authenticateLogin(creds["polly_username"], creds["polly_password"]);
 	if (status != "ok") {
-		cerr << "Incorrect credentials. Please check." << endl;
+		if (status == "error"){
+			cerr << "There was a Problem while authenticating to Polly. Please check your internet connection and try again" << endl;
```
shubhra-agrawal: "Error in authentication to Polly. Please check your internet connection and try again"
The error message sounds a little odd. Would also suggest taking care of random capitalisation.

src/pollyCLI/pollyintegration.cpp
```diff
@@ -244,11 +244,18 @@ QVariantMap PollyIntegration::getUserProjectsMap(QByteArray result2) {
...
 
-
 QVariantMap PollyIntegration::getUserProjects() {
+    QVariantMap userProjects;
     QString getProjectsCommand = "get_Project_names";
     QList<QByteArray> resultAndError = runQtProcess(getProjectsCommand, QStringList() << credFile);
-    QVariantMap userProjects = getUserProjectsMap(resultAndError.at(0));
+    // Putting EpiErrorCode key to detect possible error due to internet connection issues 
+    // or if Polly servers are Down for some reason.
+    if (resultAndError.at(1)!=""){
```
rish9511: Error stream can be populated for n number of reasons. Non empty error stream does not mean that the user is not connected to internet. Error messages like 'package x not found' also get reported on error streams.

That said, Imo the best place to check whether the user is connected to internet or not  would be `index.js`. A quick google search tells me that there is a package called [connectivity](https://www.npmjs.com/package/connectivity) that can be used for this purpose.

Also, If such errors are reported by Node, both GUI and CLI will be able to use them since they use the same Interface i.e `PollyIntegration`

PR #880, opened 2018-10-01
check_internet_connection function added.

Ky5595:
- Added a check for active internet connection in index.js
- Check the status before authentication, before file prep and before upload starts
- Error messages in the appropriate places for every error
- Also explicitly set the label color everytime

### DIFF
--- a/src/pollyCLI/pollyintegration.cpp
+++ b/src/pollyCLI/pollyintegration.cpp
@@ -121,6 +121,18 @@ QStringList PollyIntegration::get_projectFiles_download_url_commands(QByteArray 
     return patchId;
 }
 
+bool PollyIntegration::activeInternet()
+{
+    QString command = QString("check_internet_connection");
+    QList<QByteArray> results = runQtProcess(command, QStringList()).at(0).split('\n');
+    QString status = results[1];
+    if (status == "Connected") {
+        return true;
+    }
+
+    return false; 
+}
+
 // name OF FUNCTION: checkLoginStatus
 // PURPOSE:
 //    This function checks if the user is already logged in or not.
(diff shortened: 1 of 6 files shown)


src/pollyCLI/pollyintegration.cpp
```diff
@@ -121,6 +121,18 @@ QStringList PollyIntegration::get_projectFiles_download_url_commands(QByteArray
     return patchId;
 }
 
+bool PollyIntegration::activeInternet()
+{
+    QString command = QString("check_internet_connection");
+    QList<QByteArray> results = runQtProcess(command, QStringList()).at(0).split('\n');
+    QString status = results[1];
+    if (status == "Connected") {
```
rish9511: I am not sure if this would work. 
The `check_internet_connection` function that has been defined  in **index.js** outputs both the messages on **stdout**.  However the `status` variable that takes it value from `results[1]` is picking the value from **stderr**.  

`check_internet_connection` function should be using `console.error` when the user is not connected to internet

I have not tested this. Please correct me If I am wrong

shubhra-agrawal: @rish9511 The status variable is taking the value for stdout. Stdout outputs "undefined\nConnected\n". The results[1] is the string "Connected" so it does work.

I agree with the you that console.error should be used when there is no internet connection, though. Would be cleaner.

rish9511: Oh right!
I did not read the statement 
`QList<QByteArray> results = runQtProcess(command, QStringList()).at(0).split('\n')`  carefully.

> console.error should be used when there is no internet connection, though. Would be cleaner.

Yes. And then we would not have to split the string as we are doing it now

shubhra-agrawal: @rish9511 On second thoughts, console.log only has information we define. Error console could have a number of errors. Log sounds more reliable to me. What do you think?

rish9511: @shubhra-agrawal Alright, lets keep using console.log then

PR #881, opened 2018-10-01
Fix: Inconsistent CSV output format

saifulbkhan:
While testing #871, we discovered some inconsistent behavior in our group summary and peak detailed formats across different operating systems. While there might be more variable behavior which we have not yet encountered (or recorded) yet, this PR should aims to fix two of them:

1. **Order of peak export**: For peak detailed format, it seems that the order in which each peak is exported for different samples can be different. This might be due to the order in which each sample is loaded and thus their peaks are pushed to a group's peak vector in that fashion. To correct this, the peak vector is sorted according to the name's of the sample before it is iterated upon to write output.
2. **Precision of floating point values**: Until now we have not been enforcing any type (fixed or scientific) or precision for the floating point values that are exported in CSV files. This led to different formats for fractional values in different systems (tried in Mac vs Ubuntu), probably because the default state of output streams responsible for writing these values are in different states or have different rules of formatting floating values. To solve this issue, we will now exactly define precision and type for the output stream writing to CSV files.

These changes should ideally be pushed only with proper tests, but the current state of CSV report testing is too poor to cover these scenarios. The tests in #871 should be sufficient for the changes but they themselves depend on consistency of output to pass CI check. Therefore, #871 can only be merged after this PR has been merged.

### DIFF
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -365,9 +365,12 @@ void CSVReports::writePeakInfo(PeakGroup* group) {
         if(group->label !='g') return;
     } else if (selectionFlag == 3) {
         if(group->label !='b') return;
-    } else {
-
     }
+
+    // sort the peaks in the group according to the sample names using a comparison function
+    // this ensures that the order in which the peaks are written is same across different systems.
+    std::sort(group->peaks.begin(), group->peaks.end(), Peak::compSampleName);
+
     for (unsigned int j = 0; j < group->peaks.size(); j++) {
         Peak& peak = group->peaks[j];
         mzSample* sample = peak.getSample();
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -279,7 +279,7 @@ void CSVReports::writeGroupInfo(PeakGroup* group) {
     tagString = sanitizeString(tagString.c_str()).toStdString();
     char label[2];
     sprintf(label, "%c", group->label);
-    groupReport << label << SEP << setprecision(7) << parentGroup->groupId << SEP
+    groupReport << label << SEP << fixed << setprecision(6) << parentGroup->groupId << SEP
             << groupId << SEP << group->goodPeakCount << SEP << group->meanMz
             << SEP << group->meanRt << SEP << group->maxQuality << SEP
             << tagString;
@@ -333,6 +333,9 @@ void CSVReports::writeGroupInfo(PeakGroup* group) {
         groupReport << SEP << group->meanMz;
     }
 
+    // for intensity values, we only write two digits of floating point precision
+    // since these values are supposed to be large (in the order of > 10^3).
+    groupReport << setprecision(2);
     for (unsigned int j = 0; j < samples.size(); j++){
         for(int i=0;i<group->samples.size();++i){
             if(samples[j]->sampleName==group->samples[i]->sampleName){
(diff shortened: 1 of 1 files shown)


src/core/libmaven/csvreports.cpp
```diff
@@ -365,9 +366,21 @@ void CSVReports::writePeakInfo(PeakGroup* group) {
         if(group->label !='g') return;
     } else if (selectionFlag == 3) {
         if(group->label !='b') return;
-    } else {
-
     }
+
+    // sort the peaks in the group according to the sample names using a lambda function.
+    // this ensures that the order in which the peaks are written is same across different systems.
+    std::sort(group->peaks.begin(), group->peaks.end(), [](Peak& lhs, Peak& rhs) {
```
shubhra-agrawal: @saifulbkhan 
This method already exists in Peak class. 
`bool Peak::compSampleName(const Peak& a, const Peak& b)`
You can just use it like this:
`sort(group->peaks.begin(), groups->peaks.end(), Peak::compSampleName)`;

saifulbkhan: @shubhra-agrawal 
This seems quite apparent, but by any chance could a peak not have a sample related to it? The [comparison function](https://github.com/ElucidataInc/ElMaven/blob/develop/src/core/libmaven/Peak.cpp#L236) in Peak class does not check for such a case but before writing peak values CSVReports class does [check](https://github.com/ElucidataInc/ElMaven/blob/develop/src/core/libmaven/csvreports.cpp#L375) if the corresponding sample is not `NULL`.

Either way I should be using the Peak class' given comparator instead of my own. Just wanted to make sure that this comparator is not missing an odd edge case.

src/core/libmaven/csvreports.cpp
```diff
@@ -333,6 +333,7 @@ void CSVReports::writeGroupInfo(PeakGroup* group) {
         groupReport << SEP << group->meanMz;
     }
 
+    groupReport << setprecision(2);
```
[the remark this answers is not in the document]
saifulbkhan: Will add.

src/core/libmaven/csvreports.cpp
shubhra-agrawal: @saifulbkhan every peak should have a corresponding sampleName. If it's not there, that in itself tells us something's wrong.

saifulbkhan: @shubhra-agrawal cool. Already made the changes you requested btw.

rish9511:
@saifulbkhan 
The changes look fine to me. 
The only suggestion I have is that  we should make ofstream compatible with our custom data types, such  as PeakGroups and Peaks. Atm both `CsvReports` and `JsonReports` are writing similar reports and hence there is some degree of code duplication. 
I'll show some pseudo code to better explain what I mean:


Some header file overriding default ofstreams `<<` operator
```cpp
ofstream& operator<<(ofstream& fstream, Peak& peakObj)
{
    fstream << peakObj.rt  << peakObj.mz ;
    return fstream;
}
```

CsvReports.cpp/JsonReports will then change to
```cpp
void writePeakInfo(PeakObj peakObj)
{
   ofstream report;
   report << peakObj; 
}
```

Might help us  with code duplication problem and we also get to hide all the ugliness in a separate header 
file

src/core/libmaven/csvreports.cpp
```diff
@@ -365,9 +368,12 @@ void CSVReports::writePeakInfo(PeakGroup* group) {
         if(group->label !='g') return;
     } else if (selectionFlag == 3) {
         if(group->label !='b') return;
-    } else {
-
     }
+
+    // sort the peaks in the group according to the sample names using a lambda function.
```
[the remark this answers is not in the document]
saifulbkhan: Changed the comment.

Issue #882, opened 2018-10-03
Labels missing in CSV export if label flags are changed after peak detection


shubhra-agrawal:
**Describe the bug**
CSV export has missing labels based on which labels are off at the time of export irrespective of the labels turned on during peak detection

**To Reproduce**
1. Switch on all labels for isotope detection
2. Find Peaks with report isotopes
3. Switch off C13 label in Options
4. Export CSV for the table created in step 2
5. CSV does not have any C13 children

**Expected behavior**
CSV should export all labels that are present in the peak table. Alternatively, this could be an export option. El-MAVEN could ask users if there are any labels they do not wish to export.

**System (please complete the following information):**
 - OS: Ubuntu

saifulbkhan:
The default behavior has been changed to export all child groups of every labelled group. This will be available in v0.5. Later we can add an option to export only those labels that the user wants.

PR #883, opened 2018-10-03
Fix missing labels in CSV exports after changed label settings

saifulbkhan:
Once a peak detection operation has been completed and a table generated, the isotope labels detected should be exported out as they are shown in the peak table (which is also how they exist
internally as subgroups of each group). With this commit it is ensured that even if isotope detection options are changed, all isotope children detected are exported in the CSV report from a peak table containing labelled data.

Issue: #882

### DIFF
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -131,112 +131,57 @@ void CSVReports::insertPeakInformationIntoCSVFile(PeakGroup* group) {
 
 }
 
-void CSVReports::insertGroupInformationIntoCSVFile (PeakGroup* group) {
-
+void CSVReports::insertGroupInformationIntoCSVFile(PeakGroup* group)
+{
     if(group->compound == NULL || group->childCount() == 0) {
-
         writeGroupInfo(group);
-
-    }
-
-    else {
-
-        string formula = group->compound->formula;
-        int charge = getMavenParameters()->getCharge(group->compound);
-        bool C13Flag = getMavenParameters()->C13Labeled_BPE;
-        bool N15Flag = getMavenParameters()->N15Labeled_BPE;
-        bool S34Flag = getMavenParameters()->S34Labeled_BPE;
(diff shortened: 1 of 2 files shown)


src/core/libmaven/csvreports.cpp
```diff
@@ -131,112 +131,62 @@ void CSVReports::insertPeakInformationIntoCSVFile(PeakGroup* group) {
...
-       * Commented out in Maven776 - Kiran
-      if (counter == 0) {
-          //output zero-intensity peaks
-          insertIsotpesNotFoundInSamples(group, isotopeName);
-      }
-      */
+void CSVReports::insertIsotopes(PeakGroup* group, bool userSelectedIsotopesOnly)
+{
+    if (userSelectedIsotopesOnly) {
+        insertUserSelectedIsotopes(group);
```
rish9511: The default value of `userSelectedIsotopesOnly` is **false** and no where in the code I see it being called as **true**.
And if that's the case, `insertUserSelectedIsotopes(group)` should not even exist.
Is this the case or did I miss something ?

rish9511: From #882, it seems in future we might ask users to choose which labels to export.
If that's the reason, can we comment out the code related to this ? and enable it once we have the functionality. Won't be compiled unnecessarily
Not sure what to do with this
@shubhra-agrawal  
What do you think ?

saifulbkhan: So the changes made will by default export all the isotope labels that were detected for a group. @shubhra-agrawal suggested that there should still be an option to allow the user to export only those labels which they have selected in the "Options->Isotope Detection" settings (and filter the others out, even if they were detected), as it has been doing until now. Providing that function is the purpose of this method. Since most of the code for this method was already present, I am just refactoring and leaving it here until we have a UI allowing us to expose this on the front-end (hopefully with the UX overhaul we will have this as one of the options in our export tab).

PR #886, opened 2018-10-05
Fix PollyPhi

rish9511:
PollyPhi works only if the data is exported with C13 label.
This commit makes sure that in default settings only c13 label
is switched on


src/gui/mzroll/forms/settingsform.ui
```diff
@@ -1443,6 +1449,7 @@ p, li { white-space: pre-wrap; }
  </tabstops>
  <resources>
   <include location="../mzroll.qrc"/>
+  <include location="../mzroll.qrc"/>
```
shubhra-agrawal: @rish9511 extra include statement?

rish9511: @shubhra-agrawal 
Missed it somehow
Have pushed the necessary changes

PR #893, opened 2018-10-12
Use QMessageBox for all DB warning messages

saifulbkhan:
`QMessageBoxResize` was being used for warnings generated after loading compounds DB file. This subclass of message box prohibits scrolling and was not suitable for showing a list of entries that might be present in the detailed test message. Therefore the original `QMessageBox` class will now be used instead since it does not have such restrictions.


saifulbkhan:
@shubhra-agrawal Its being used by a [method](https://github.com/ElucidataInc/ElMaven/blob/develop/src/gui/mzroll/mainwindow.cpp#L1847) in `mainwindow.cpp`. I was not sure of the context in which this was written and that is why chose to let it be (there must be a strong reason to write a separate class for the message box).

PR #894, opened 2018-10-12
[Fix] Check for empty line in compound db

shubhra-agrawal:
Stray newline characters from editing csv files in general editors are very common. Empty lines in the compound db led to unexpected crashes.

This crash has now been resolved.

### DIFF
--- a/src/core/libmaven/databases.cpp
+++ b/src/core/libmaven/databases.cpp
@@ -20,9 +20,13 @@ int Databases::loadCompoundCSVFile(string filename) {
     while (getline(myfile,line)) {
         //This is used to write commands
         if (!line.empty() && line[0] == '#') continue;
-        
+
         //trim spaces on the left
-        line.erase(line.find_last_not_of(" \n\r\t")+1);
+        size_t found = line.find_last_not_of(" \n\r\t");
+        if (found != string::npos)
+            line.erase(found+1);
+        else continue;
+        
         lineCount++;
 
         vector<string> fields;
--- a/bin/methods/compoundlist.csv
+++ b/bin/methods/compoundlist.csv
@@ -1,9 +1,12 @@
 mz,rt,charge,name,formula,id
+
 588.07523818,14.14,-1,ADP-D-glucose,C16H25N5O15P2,C00498
 0,15.5,0,acetoacetyl-CoA,C25H40N7O18P3S,
 101.03767,15.46,-2,acetyl-CoA,,
 507.99277,15.46,-1,dGTP,,
+
 ,1,-1,YourFavoriteCompound1,,
 120.0433,2,-2,YourFavoriteCompound2,,
 ,,0,YourFavoriteCompound3,C4H9O4R,
 120.0433,,,YourFavoriteCompound4,,
+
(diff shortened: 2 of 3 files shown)


src/core/libmaven/databases.cpp
```diff
@@ -20,9 +20,13 @@ int Databases::loadCompoundCSVFile(string filename) {
     while (getline(myfile,line)) {
         //This is used to write commands
         if (!line.empty() && line[0] == '#') continue;
-        
+
         //trim spaces on the left
-        line.erase(line.find_last_not_of(" \n\r\t")+1);
+        size_t found = line.find_last_not_of(" \n\r\t");
```
rish9511: This can be simplied to:
`line.erase(line.find_last_not_of("\r\n\t")+1);`

That's also how it is done in GUI:
https://github.com/ElucidataInc/ElMaven/blob/92b75ba3bb2f43d1bd2a4032150ae3fb149a0162/src/gui/mzroll/database.cpp#L471

Also, for me, on Windows the GUI crashes for the attached DB file but the CLI does not
Can you test this on Windows once ?

[KNOWNS.zip](https://github.com/ElucidataInc/ElMaven/files/2475163/KNOWNS.zip)

shubhra-agrawal: @rish9511 'found' stores whether there are any other characters in this line apart from " \n\t\r" or not. If this is an empty line, don't send it to csvParser in mzUtils::get_fields() and move to the next line. get_fields is where the app crashes.

I forgot to make this same change for the GUI that uses database.cpp which is probably why GUI is crashing for this file. While I'm at it, I'll also add a test for databases.cpp. Thanks for reminding me!

shubhra-agrawal: @rish9511 'found' stores whether there are any other characters in this line apart from " \n\t\r" or not. If this is an empty line, don't send it to csvParser in mzUtils::get_fields() and move to the next line. get_fields is where the app crashes.

I forgot to make this same change for the GUI that uses database.cpp which is probably why GUI is crashing for this file. While I'm at it, I'll also add a test for databases.cpp. Thanks for reminding me!

shubhra-agrawal: @rish9511 I've pushed the required changes. Let me know if this works for you now.

shubhra-agrawal:
@saifulbkhan the expected behavior in GUI is that the empty lines will be skipped. I tested it with the file @rish9511 attached here and bin/compoundlist.csv on Ubuntu. Did not receive any error while loading. I did not test it for an empty file though, is that what you tested with?

saifulbkhan:
> I tested it with the file @rish9511 attached here and bin/compoundlist.csv on Ubuntu. Did not receive any error while loading. I did not test it for an empty file though, is that what you tested with?

Yes I tested it with a file with a file with just a single newline character. Files with regular contents and a few newlines inserted here and there seem to work fine. But a file that is completely empty or has only whitespace elements gives a warning like "Trouble Loading File". That is all I was pointing out. Again it is not very related to the issue you are trying to solve.

src/core/libmaven/databases.cpp
rish9511: 👍

PR #898, opened 2018-10-16
[Fix] Exit peakdetector CLI when compound db is empty

shubhra-agrawal:
Execution of the CLI command should stop in case of an empty compound db during targeted analysis


shubhra-agrawal:
@saifulbkhan Pushed another commit with refactored code. Please take a look.

PR #900, opened 2018-10-18
[Fix] Resolve crash in merging tables

shubhra-agrawal:
Merging the bookmarked groups to a peak table led to the app crashing if any peak table had been deleted during the session.
The merging code was working on the assumption that the table id and the peakList indices will remain in sync.
Deleting a peak table led to an index out of range error.
This has been resolved by searching for the correct Id

Steps to reproduce the crash on develop:
- Create and Delete any peak table
- Bookmark a peak
- Create another peak table
- Merge bookmark table with this new table
- App crashes

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -2636,9 +2636,9 @@ void BookmarkTableDockWidget::mergeGroupsIntoPeakTable(QAction *action)
 
     int initialSize = peakTable->allgroups.size();
     int finalSize = allgroups.size() + initialSize;
-    for (unsigned int i = 0; i < allgroups.size(); i++) {
-        allgroups[i].groupId = ++initialSize;
-        peakTable->allgroups.append(allgroups[i]);
+    for (auto group : allgroups) {
+        group.groupId = ++initialSize;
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -2602,37 +2602,53 @@ void BookmarkTableDockWidget::showMsgBox(bool check, int tableNo) {
...
-  showAllGroups();
+    //return if peak table not found
+    if (!peakTable) {
+        showMsgBox(false, j);
+        return;
+    }
+
+    int initialSize = peakTable->allgroups.size();
+    int finalSize = allgroups.size() + initialSize;
+    for (unsigned int i = 0; i < allgroups.size(); i++) {
```
saifulbkhan: We can still use a range based for loop I think:
```C++
    for (auto group : allgroups) {
        group.groupId = ++initialSize;
        peakTable->allgroups.append(group);
    }
```
The above works for me without any issues. If you can, please use this. Otherwise its not a big deal.

shubhra-agrawal: Didn't notice. Changing it

Issue #583, opened 2018-02-05
New baseline Algorithm


sahil21:
There are a couple of challenges with current baseline algorithm:
1. The baseline calculated by El-MAVEN is often 'inaccurate'
2. Baseline correction is currently done based on a global setting. We can't set separate baseline for individual compounds/groups (similar to #556)

sahil21:
Key points from the meeting with @chubukov 
- Noise and Baseline should be dealt with the same algorithm
- Validating baseline algorithm is hard - Use simulated data or check visually.

Some of the baseline algorithm we are trying
1. **airPLS algorithm** - https://code.google.com/archive/p/airpls/
2. **IRLS algorithm** - https://cran.r-project.org/web/packages/baseline/baseline.pdf
3. **peakutils baseline** -  http://pythonhosted.org/PeakUtils/index.html

sahil21:
@chubukov We have made a prototype depicting 3 baseline algorithms i.e **airPLS**, **IRLS** and **peakutils baseline**. You can test out the prototype here - [Baseline Prototype - Polly Dash App ](https://polly.elucidata.io/polly-dash?polly-notebook-name=baseline.zip&polly-project-id=52)

About Baseline prototype:
- There are 4 sliders in the dash app (1 each for airPLS and peakutils. There are 2 sliders for IRLS. More information around these algorithms is mentioned in the previous comment)
- You can select multiple algorithms to compare the results
- We have added 5 example EICs to test out the baseline algorithms (Select EICs using drop-down)

Let us know your thoughts around these baseline algorithms. 

PS. We have used [Polly](polly.elucidata.io) just to host the [dash app](https://plot.ly/products/dash/) for easy sharing of analysis. The baseline code will be part of this open source repository.

chubukov:
@sahil21 Super. Really great idea. 

The only algorithm that seems to me to give anything reasonable is IRLS. I got what looks to me like pretty decent behavior at lambda1=4 and lambda2=7. 

Was that your experience as well? Were you able to get anything reasonable out of the other two? If so, at what parameters?

Once we have it narrowed to just a few options to test, let's run it on a much wider variety of EICs?

sahil21:
@chubukov I also found IRLS reasonably better than other two (airPLS & peakutils). 

[...]

PR #911, opened 2018-11-09
AsLS Smoothing algorithm for baseline correction

saifulbkhan:
This set of commits adds a new baseline correction algorithm, incorporates this into the GUI and adds (basic) test cases for the algorithm. Please refer to the commit messages and issue #583 for more context.


tests/MavenTests/testEIC.h
```diff
@@ -17,11 +17,9 @@ class TestEIC : public QObject {
 
     public:
         TestEIC();
+
     private:
-        const char* loadFile;
-        QStringList files;
-        QStringList files_ms2;
-        mzSample* _goodSample;
+        Samples* samples;
```
rish9511: IMO `Common` should be initialized once and should be accessible to everyone. 
It can be made global or we can use a singleton, there's a slight difference between them
Otherwise anyone who needs it will have to store a pointer of it, all the samples will have to be loaded again. I will have to do the same for Alignment 

What do you think ?

saifulbkhan: How are you thinking of circulating the global (or singleton) `Samples` instance? I initially sought to do exactly what you wrote, but then I would have to declare `Samples` in `main.cpp` (for MavenTests module) or make `main.cpp` dependent on `common.h` and then pass the instance to each test upon instantiation. This seemed somewhat unclean to me. Like you said, loading the samples once for a test is not a costly operation, compared to loading them for each function (as it was before). I'm open to any good suggestions.

rish9511: We can have a file called `maventests.h`  with the following definition

```
...

#include "common.h"

namespace MavenTests 
{
    Samples* samples = new Samples;
    ....
    ....
}

....
```

And then in `testEic.cpp`

```
....
#include "maventests.h"


void TestEIC::testgetEICms2() {
    mzSample* sample1 = MavenTests::samples->ms2Samples[0];
    ....
} 

....
```

Do you know/think of other things that can be part of `MavenTests` ? If yes, then we can have something like this ?

rish9511: Maybe the database is also loaded every time 
Just making a guess

shubhra-agrawal: @rish9511 Why the decision to create a separate `maventests.h` file? The common modules like samples and databases can be initialized in `common.h` as well. If it has the common functions for different test files, then the common modules like samples and databses should be part of it as well. Correct me if I'm missing something.

rish9511: > If it has the common functions for different test files, then the common modules like samples and databses should be part of it as well

Correct, this is also  what I realized later. The class `Common` can be used for all of this and maybe re-named to something better(MavenTests, TestUtils  etc)

tests/MavenTests/common.h
```diff
@@ -14,6 +14,39 @@ extern Databases DBS;
...
+
+    Samples() {
+        smallSample = new mzSample();
+        smallSample->loadSample("bin/methods/testsample_1.mzxml");
+
+        auto sample1 = new mzSample();
+        auto sample2 = new mzSample();
+        sample1->loadSample("bin/methods/testsample_2.mzxml");
+        sample2->loadSample("bin/methods/testsample_3.mzxml");
+        regularSamples.push_back(sample1);
```
shubhra-agrawal: Would suggest using `ms1Samples` instead

tests/MavenTests/testEIC.cpp
```diff
@@ -1,12 +1,7 @@
 #include "testEIC.h"
 
 TestEIC::TestEIC() {
-    loadFile = "bin/methods/testsample_1.mzxml";
-    loadGoodSample = "bin/methods/testsample_2.mzxml";
-
-    files << "bin/methods/testsample_2.mzxml" << "bin/methods/testsample_3.mzxml";
-    files_ms2 << "bin/methods/ms2test1.mzML" << "bin/methods/ms2test2.mzML";
-
+    samples = new Samples();
```
shubhra-agrawal: Could you change `Samples` to `TestSamples` or something? `new Samples();` is too close to `new mzSample();` imo.

tests/MavenTests/testEIC.h
shubhra-agrawal: Great. I vote for TestUtils. I'm not sure if this change should be made in this particular PR, though. @saifulbkhan your call.

rish9511: I guess we will have to skip this one for now. 
@saifulbkhan  Do you think we will have time to refactor the `common` class ?

src/core/libmaven/EIC.cpp
```diff
@@ -278,6 +285,15 @@ void EIC::computeThresholdBaseline(int smoothingWindow, int dropTopX)
     //cerr << "eic no noise obs = " << eic_noNoiseObs << "\n";
 }
 
+void EIC::computeBaseline()
+{
+    if (_baselineMode == BaselineMode::Threshold)
```
rish9511: I always prefer Switch case  over if/else if/else where ever it's possible. , especially in the case of Enums

src/core/libmaven/EIC.cpp
```diff
@@ -146,25 +148,27 @@ void EIC::computeBaseLine(int smoothingWindow, int dropTopX)
...
-        computeThresholdBaseline(smoothingWindow, dropTopX);
-    else if (baselineMode == BaselineMode::AsLSSmoothing)
-        computeAsLSBaseline(5.0f, 0.05f);
+    return true;
 }
 
-void EIC::computeAsLSBaseline(int lambda, float p, int numIterations)
+void EIC::_computeAsLSBaseline(float lambda, float p, int numIterations)
 {
+    if (!_clearBaseline())
```
rish9511: Can we have this check in `computeBaseLine()` ?  
We won't have to put this check in both the functions then

rish9511: Do you plan to make use of this check in future ? Say, if the baseline could not be cleared, will the user be informed in that case ?

tests/MavenTests/testEIC.cpp
```diff
@@ -64,67 +65,129 @@ void TestEIC::testgetEICms2() {
...
+
+    // TODO: for lack of better floating point comparators to test the
+    // consistency of the output, we can at least test whether the operation
+    // completes successfully
+    e->computeBaseline();
+
+    // test whether all values are greater than zero, otherwise EIC widget
+    // behaves erratically
+    auto allGreaterThanZero = true;
+    for (unsigned int i = 0; i < e->intensity.size(); ++i)
```
rish9511: Can be converted to this:
```
if(!std::all_of(e->intensity.begin(), e->intensity.end(), [](float intensity) { return intensity >= 0.0f; }
    allGreaterThanZero = false;

```

tests/MavenTests/testEIC.cpp
```diff
@@ -64,69 +65,132 @@ void TestEIC::testgetEICms2() {
...
+        if (e->baseline[i] < 0.0f)
+            allGreaterThanZero = false;
```
rish9511: Can be converted to this:

```
if(!std::all_of(e->intensity.begin(), e->intensity.end(), [](float intensity) { return intensity >= 0.0f; }
    allGreaterThanZero = false;

```

src/core/libmaven/PeakDetector.cpp
```diff
@@ -82,8 +83,16 @@ vector<EIC*> PeakDetector::pullEICs(mzSlice* slice,
...
-                        e->setBaselineDropTopX(baseline_dropTopX);
+
+                        // set appropriate baseline parameters
+                        e->setBaselineMode(baselineMode);
+                        if (baselineMode == EIC::BaselineMode::AsLSSmoothing) {
+                            e->setAsLSSmoothness(firstBaselineParameter);
+                            e->setAsLSAsymmetry(secondBaselineParameter);
+                        } else {
+                            e->setBaselineSmoothingWindow(firstBaselineParameter);
+                            e->setBaselineDropTopX(secondBaselineParameter);
```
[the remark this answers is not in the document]
rish9511: The same set of logic seems to be replicated else where. I believe, if I have not missed anything, we can get rid of this logic elsewhere also. Correct me if I am wrong

src/core/libmaven/EIC.cpp
saifulbkhan: Yes this can be included in `computeBaseline()` method itself.

As for you second comment, I think a simple debug message should be fine in such a case?

src/core/libmaven/EIC.cpp
```diff
@@ -124,29 +128,120 @@ EIC *EIC::eicMerge(const vector<EIC *> &eics)
...
+    return true;
+}
+
+void EIC::_computeAsLSBaseline(float lambda, float p, int numIterations)
+{
+    if (!_clearBaseline())
+        return;
+
+    using namespace Eigen;
+    unsigned int n = static_cast<unsigned int>(intensity.size());
```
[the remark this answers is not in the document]
saifulbkhan: Sure.

src/core/libmaven/EIC.cpp
```diff
@@ -124,29 +128,120 @@ EIC *EIC::eicMerge(const vector<EIC *> &eics)
...
+    // clip negative values from the baseline vector
+    std::transform(begin(clippedBaseline),
+                   end(clippedBaseline),
+                   begin(clippedBaseline),
+                   [](float f) { return f < 0.0f ? 0.0f : f; });
+
+    // since baseline right now is an array of float, the data from clipped
+    // vector is copied to this array
+    std::copy(begin(clippedBaseline), end(clippedBaseline), baseline);
+}
```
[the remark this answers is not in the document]
saifulbkhan: I do not think the reader (for this piece of code) should have to go through each step and affirm that the return type on each step is correct, or even be aware of it. Most of these variables are some type derived from templates of Eigen library's `Matrix` or `Vector` meta types. Putting them within this already complicated algorithm will in fact decrease the readability in my opinion. Also I believe `auto` was meant for being used as a [duck typing](https://en.wikipedia.org/wiki/Duck_typing) alternative to hide precisely this kind of type complexity.

> I also feel some of the variable declared above can be named better.

Apart from `w` (for weights), I think the rest of the variables are fine. I myself am usually very adamant about verbosity of variables, but since this algorithm was derived from the original author's pseudo-code implementation, the variables have been kept as such to allow easy identification of all elements involved. Besides its a common practice (in mathematical computing) to use single letter lower-case symbols for vectors and upper-case symbols for Matrices. Additionally, the formula `A·x = b` is extremely commonly used for representing a system of linear equations, and hence the name of the symbols.

As for the comments, I think, a few months down the line developers (including me) will have an easier time debugging this where each step for this algorithm is slightly explained. Please allow this exception for such functions implementing scientific papers.

src/core/libmaven/EIC.cpp
```diff
@@ -124,29 +128,120 @@ EIC *EIC::eicMerge(const vector<EIC *> &eics)
...
+
+Eigen::SparseMatrix<float> EIC::_diff(Eigen::SparseMatrix<float> mat)
+{
+    Eigen::SparseMatrix<float> E1 = mat.block(0, 0, mat.rows() - 1, mat.cols());
+    Eigen::SparseMatrix<float> E2 = mat.block(1, 0, mat.rows() - 1, mat.cols());
+    return E2 - E1;
```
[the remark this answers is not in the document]
saifulbkhan: Since this is only ever used by the `_computeAsLSBaseline` method, why not just put this piece of code within the method as a lambda function. I don't see any reason why something so unrelated should be left hanging around as a member of our EIC class. What do you think?

src/core/libmaven/EIC.cpp
```diff
@@ -124,29 +128,120 @@ EIC *EIC::eicMerge(const vector<EIC *> &eics)
...
+}
+
+void EIC::_computeThresholdBaseline(int smoothingWindow, int dropTopX)
+{
+    if (!_clearBaseline())
+        return;
+
+    unsigned int n = static_cast<unsigned int>(intensity.size());
```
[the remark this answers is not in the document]
saifulbkhan: This clearly does not make sense, since, sizes of collections are `unsigned int` by default.

src/core/libmaven/EIC.cpp
saifulbkhan: Sure.

src/core/libmaven/EIC.h
```diff
@@ -183,6 +189,18 @@ class EIC
...
+     * @brief Set smoothness (λ) to be used for default AsLS baseline estimation.
+     * @param s smoothness (will be mutated to 10^s when actually used)
+     */
+    void setAsLSSmoothness(int s) { _aslsSmoothness = s; }
+
+    /**
+     * @brief Set asymmetry (p) to be used for default AsLS baseline estimation.
+     * @param a asymmetry value (will be divided by 100 when actually used).
+     */
+    void setAsLSAsymmetry(int a) { _aslsAsymmetry = a; }
```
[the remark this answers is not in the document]
saifulbkhan: Yes it should.

src/core/libmaven/EIC.h
```diff
@@ -265,10 +283,74 @@ class EIC
...
+     * @return Whether the baseline should be processed further or not.
+     */
+    bool _clearBaseline();
+
+    /**
+     * @brief Computes a baseline using naive thresholding method.
+     * @param smoothingWindow is the size of window used for 1D guassian smoothing.
+     * @param dropTopX percent of the highest intensities will be truncated.
+     */
+    void _computeThresholdBaseline(int smoothingWindow, int dropTopX);
```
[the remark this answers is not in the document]
saifulbkhan: Yes.

src/core/libmaven/EIC.h
```diff
@@ -265,10 +283,74 @@ class EIC
...
+     *
+     * @param lambda for smoothness. Typical values of lambda for MS data range
+     * from 10^3 to 10^9, depending on dataset. The exponent value should be
+     * passed here as integer, i.e. lambda should be within 3 to 9.
+     * @param p for asymmetry. Values between 0.01 to 0.10 work reasonable well
+     * for MS data.
+     * @param numIterations for the number of iterations that should be
+     * performed (since this is an iterative optimization algorithm).
+     */
+    void _computeAsLSBaseline(float lambda, float p, int numIterations=10);
```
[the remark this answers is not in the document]
saifulbkhan: Okay.

src/core/libmaven/EIC.h
```diff
@@ -265,10 +283,74 @@ class EIC
...
+
+    /**
+     * @brief Calculate and return approximate derivative of a sparse matrix.
+     * @param mat instance of Eigen::SparseMatrix containing double values.
+     * @return Difference matrix of the mat sparse matrix.
+     */
+    Eigen::SparseMatrix<float> _diff(Eigen::SparseMatrix<float> mat);
```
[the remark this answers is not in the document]
saifulbkhan: Planning to move this as a lambda within `_computeAsLSBaseline`, unless we have a strong reason not to do so.

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
saifulbkhan: I think I will have to try out these snippets before I can comment. While what you suggest reduces code quantity (when setting parameters each time like above), the current method makes more sense semantically - as in what would we do if the baseline parameter for a new algorithm takes in three arguments. I hope we do not need another such baseline implementation, but pointing this out just for consideration.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -342,8 +360,9 @@ void PeakDetector::processSlices(vector<mzSlice *> &slices, string setName)
                         mavenParameters->eic_smoothingAlgorithm,
                         mavenParameters->amuQ1,
                         mavenParameters->amuQ3,
-                        mavenParameters->baseline_smoothingWindow,
-                        mavenParameters->baseline_dropTopX,
+                        baselineMode,
+                        firstBaselineParam,
+                        secondBaselineParam,
```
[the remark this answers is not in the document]
saifulbkhan: Again I need to try out the snippets before commenting on your ideas.

tests/MavenTests/common.h
saifulbkhan: Alright.

tests/MavenTests/common.h
```diff
@@ -14,6 +14,39 @@ extern Databases DBS;
...
+        regularSamples.push_back(sample2);
+
+        auto ms2Sample1 = new mzSample();
+        auto ms2Sample2 = new mzSample();
```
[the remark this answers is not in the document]
saifulbkhan: I think the code is more readable as it is. Calling methods on indexed elements does not seem very clean.

tests/MavenTests/testEIC.cpp
saifulbkhan: I can do that.

tests/MavenTests/testEIC.cpp
saifulbkhan: Will do.

tests/MavenTests/testEIC.cpp
```diff
@@ -64,67 +65,129 @@ void TestEIC::testgetEICms2() {
...
+    QVERIFY(fabs(*maxVal) <= 1e-10f);
+
+    // deallocate
+    delete e;
+}
+
+void TestEIC::testcomputeBaselineEmptyEIC()
+{
+    // create an empty EIC, i.e., its intensity vector is empty
+    EIC* e = new EIC();
```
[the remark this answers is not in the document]
saifulbkhan: Oh I should.

tests/MavenTests/testEIC.h
saifulbkhan: I cannot guarantee I will have enough time for refactoring classes anymore. But since `common` is a relatively small class, I could give it a try.

src/core/libmaven/EIC.cpp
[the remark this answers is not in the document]
rish9511: > why not just put this piece of code within the method as a lambda function.

Yes,we can do that or maybe declare it inline. Whatever seems appropriate to you

src/gui/mzroll/forms/settingsform.ui
```diff
@@ -425,53 +396,259 @@
...
+                 <property name="toolTip">
+                  <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Smoothness (&lt;span style=&quot; font-weight:600;&quot;&gt;λ&lt;/span&gt;) for flexibility of the baseline. This value is represented as a linear scale for log of &lt;span style=&quot; font-weight:600;&quot;&gt;λ&lt;/span&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+                 </property>
+                 <property name="text">
+                  <string/>
+                 </property>
+                </widget>
+               </item>
+               <item row="0" column="1">
+                <widget class="QSlider" name="smoothnessSlider">
```
shubhra-agrawal: Can we have some default values for both sliders? If there is a default value, it should also be displayed in the value label. Currently, it does not show any selected value or range for either of the sliders. 
image

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
shubhra-agrawal: > what would we do if the baseline parameter for a new algorithm takes in three arguments. I hope we do not need another such baseline implementation, but pointing this out just for consideration.

I agree. Just because these 2 algorithms happened to have 2 parameters, does not mean we should take advantage of it. 

Considering we're going to move towards metabolite specific settings soon, it might be a good idea to think of structs to bundle different sets of parameters together? All EIC settings like smoothing algo, window, baseline, baselineDiff etc can be stored in one struct and passed to this function. What do you guys think?

src/core/libmaven/PeakDetector.h
```diff
@@ -136,12 +136,21 @@ class PeakDetector {
 	 */
 	vector<mzSlice*> processCompounds(vector<Compound*> set, string setName);
 
-	static vector<EIC*> pullEICs(mzSlice* slice, std::vector<mzSample*>&samples,
-			int peakDetect, int smoothingWindow, int smoothingAlgorithm,
-			float amuQ1, float amuQ3, int baselineSmoothingWindow,
```
shubhra-agrawal: int peakDetect is not being used in pullEICs

src/gui/mzroll/forms/settingsform.ui
```diff
@@ -1449,6 +1617,7 @@ p, li { white-space: pre-wrap; }
  </tabstops>
  <resources>
   <include location="../mzroll.qrc"/>
```
shubhra-agrawal: got added twice.

src/gui/mzroll/settingsform.cpp
```diff
@@ -499,3 +525,15 @@ void SettingsForm::setStringValue(QString key, QString value) {
...
 
+void SettingsForm::_smoothnessChanged(int value)
+{
+    smoothnessValue->setText(QString::number(value));
+    recomputeEIC();
+}
+
+void SettingsForm::_asymmetryChanged(int value)
+{
+    double actualAsym = static_cast<double>(value) / 100.0;
```
shubhra-agrawal: @saifulbkhan Why the conversion to actualAsym? You can set the range and step size of the slider to give you the actual number you want. Was there a reason for converting it yourself?

src/gui/mzroll/forms/settingsform.ui
saifulbkhan: I am assigning these default values (same as in EIC.cpp): smoothness - 5, asymmetry - 3. Their respective labels will also have these as default values and change on any further `valueChanged` signal.

src/core/libmaven/PeakDetector.h
saifulbkhan: Yes, this PR was just minding its own business. I can remove it as an argument but we will eventually be coming back to these methods for more refactoring anyway.

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
saifulbkhan: Using a struct for passing a bunch of parameters around is not a bad idea. But we should carefully outline and define what these structures will be. Otherwise we might end up with another `MavenParameters` sort of thing. `MavenParameters` is not so bad though, and for now an instance of it can be passed to `pullEICs` to reduce a fair amount of arguments.

src/gui/mzroll/forms/settingsform.ui
saifulbkhan: Removing this.

src/gui/mzroll/settingsform.cpp
saifulbkhan: Unfortunately, sliders can only have integers for values and steps. That is why this complication.

tests/MavenTests/common.cpp
```diff
@@ -1,62 +1,72 @@
 #include "common.h"
```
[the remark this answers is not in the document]
saifulbkhan: Okay.

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
shubhra-agrawal: I'm suggesting a struct for EIC parameters within MavenParameters. Instead of sending the full MavenParameters object to functions, we can pass that one struct.

src/gui/mzroll/settingsform.cpp
shubhra-agrawal: Oh yes. My bad.

tests/MavenTests/testEIC.cpp
```diff
@@ -30,130 +23,221 @@ void TestEIC::cleanup() {
...
     e2 = mzsample_2->getEIC(195,0,69,1,"",0.5,0.5);
     QVERIFY(e2->rt.size() == 305);
     QVERIFY(e2->scannum[e2->scannum.size()-1] == 3041);
     QVERIFY(e2->maxIntensity == 10600);
 
-    mzsample_2->loadSample(files_ms2.at(0).toLatin1().data());
     EIC* e3 = NULL;
     e3 = mzsample_2->getEIC(195,0,69,1,"",2,2);
     QVERIFY(e3->rt.size() == 305);
-    QVERIFY(e3->scannum[e3->scannum.size()-1] == 6080);
```
shubhra-agrawal: The scan number changed to half the previous value? why? looks like it's the same file and settings.

saifulbkhan: The value being checked should have been half all this time. The same mzSample object (namely `mzsample_2`) was being used to load the same sample (having 3040 number of scans) file twice. Not sure why this is not handled in `mzSample` class but loading another sample through the same object simply appended the new scans (size: 3040) from second load into the existing scans (size: 3040) vector. Maybe this is intentional, but I fail to see the use case.

src/gui/mzroll/forms/settingsform.ui
shubhra-agrawal: The Asymmetry label shows '3' at the start not '0.03'
The default algorithm is AsLss at the moment. I think it might be due to QtCreator. Before saving the file, change back the tab to Threshold and the main tab to Instrumentation tab. It saves the state of the file.

tests/MavenTests/testEIC.cpp
shubhra-agrawal: ahhh. got it. No, it's not intentional.

src/gui/mzroll/forms/settingsform.ui
saifulbkhan: It was actually the absence of new baseline settings in "default_settings.xml" and the generated "settings.h" file.

src/core/libmaven/EIC.cpp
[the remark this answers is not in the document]
saifulbkhan: `intensitySize` would be too large a variable name to use for size declaration and indexing.

src/core/libmaven/EIC.cpp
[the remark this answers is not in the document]
saifulbkhan: This was actually to avoid a warning of implicit conversion.

src/core/libmaven/PeakDetector.h
```diff
@@ -138,17 +138,7 @@ class PeakDetector {
...
-                                 int smoothingAlgorithm,
-                                 float amuQ1,
-                                 float amuQ3,
-                                 EIC::BaselineMode baselineMode,
-                                 int firstBaselineParameter,
-                                 int secondBaselineParameter,
-                                 double minSignalBaselineDifference,
-                                 int eicType,
-                                 string filterline);
+                                 MavenParameters* mp);
```
shubhra-agrawal: I have concerns around sending a reference to MavenParameters. The object is shared with multiple classes which can update the parameters. A lot of our UI currently blocks any change in baseline parameters while the algo is in process, so we can work on this in the next release but just pointing it out for future ref.

rish9511: If we  want MavenParameters to be Immutable inside, then we should use a `const reference`.

PR #914, opened 2018-11-14
Enable OpenMp on MacOS

rish9511:
QMake by default uses Clang that comes bundled with XCode. That version
of Clang does not have OpenMp support. In order to use Clang which comes
with OpenMp support we need to use:
1. LLVM's clang(https://clang.llvm.org/)
2. Version >= 3.8

1. Removes all '#ifndef' directives that stops Mac from
using OpenMp
2. Adds proper linker and compiler flags that are required to
use the right Clang vesion

### DIFF
--- a/src/core/libmaven/libmaven.pro
+++ b/src/core/libmaven/libmaven.pro
@@ -13,8 +13,7 @@ QMAKE_CXXFLAGS +=  -std=c++11
 QMAKE_CXXFLAGS += -DOMP_PARALLEL
 linux: QMAKE_CXXFLAGS += -Ofast -ffast-math
 win32: QMAKE_CXXFLAGS += -Ofast -ffast-math
-!macx: QMAKE_CXXFLAGS += -fopenmp
-!macx: LIBS += -fopenmp
+QMAKE_CXXFLAGS += -fopenmp
 
 TARGET = maven
 
@@ -27,14 +26,16 @@ INCLUDEPATH +=  $$top_srcdir/3rdparty/pugixml/src/ \
                 $$top_srcdir/3rdparty/libcsvparser \
                 $$top_srcdir/3rdparty/libdate/ \
                 $$top_srcdir/3rdparty/ErrorHandling \
-                $$top_srcdir/3rdparty/obiwarp
+                $$top_srcdir/3rdparty/obiwarp \
 
 QMAKE_LFLAGS += -L$$top_builddir/libs
+
 LIBS += -lobiwarp
 
 macx{
 
-    INCLUDEPATH += /usr/local/include/
+    INCLUDEPATH += /usr/local/include/ \
+                   /usr/local/Cellar/llvm/6.0.1/lib/clang/6.0.1/include/
     QMAKE_LFLAGS += -L/usr/local/lib/
     LIBS +=  -lboost_signals
 }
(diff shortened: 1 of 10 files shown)


src/gui/mzroll/mzroll.pro
```diff
@@ -12,10 +12,9 @@ QMAKE_CXXFLAGS +=  -std=c++11
 QMAKE_CXXFLAGS += -DOMP_PARALLEL
 linux: QMAKE_CXXFLAGS += -Ofast -ffast-math
 win32: QMAKE_CXXFLAGS += -Ofast -ffast-math
-!macx: QMAKE_CXXFLAGS += -fopenmp
+QMAKE_CXXFLAGS += -fopenmp
 !macx: LIBS += -fopenmp
```
shubhra-agrawal: @rish9511 Don't need this condition anymore, do we?

src/gui/mzroll/mzroll.pro
```diff
@@ -48,11 +47,13 @@ win32 {
 }
 
 mac {
-    INCLUDEPATH  += $$top_srcdir/3rdparty/google-breakpad/src/
-    QMAKE_LFLAGS += -L$$top_builddir/libs/
+    QMAKE_CXXFLAGS += -fopenmp
```
shubhra-agrawal: Added this flag already on line 15

shubhra-agrawal: breakpad and pthread is also common to all 3 platforms. Can't that be moved out of the individual blocks?

tests/MavenTests/MavenTests.pro
```diff
@@ -28,7 +28,11 @@ LIBS += -lmaven -lpugixml -lneural -lcsvparser -lpls -lErrorHandling -lLogger -l
 !macx: LIBS += -fopenmp
 
 macx {
-LIBS -= -lnetcdf -lcdfread
+    QMAKE_CXXFLAGS += -fopenmp
```
shubhra-agrawal: Here as well. openmp flag can be added on line 18

src/gui/mzroll/mzroll.pro
rish9511: > Added this flag already on line 15

This has been removed  in the next set of commits. 

> breakpad and pthread is also common to all 3 platforms

Breakpad is but  I doubt pthread is. Not really sure why pthread is added in the Mac section

> Can't that be moved out of the individual blocks?

It can be. I plan to fix/refactor our pro files in the coming time, do you want me to fix this in this PR?

shubhra-agrawal: Not exactly related to this PR so you can do it later if you want. (unless it's the bucket list kind of 'later' :P )

src/core/libmaven/libmaven.pro
```diff
@@ -13,8 +13,7 @@ QMAKE_CXXFLAGS +=  -std=c++11
 QMAKE_CXXFLAGS += -DOMP_PARALLEL
 linux: QMAKE_CXXFLAGS += -Ofast -ffast-math
 win32: QMAKE_CXXFLAGS += -Ofast -ffast-math
-!macx: QMAKE_CXXFLAGS += -fopenmp
-!macx: LIBS += -fopenmp
```
shubhra-agrawal: Adding openMP to LIBS for all platform here but not in mzroll. Why is that? just want to understand

rish9511: LIBS flag is used by the linker and not by the compiler, and all the linker flags have to be added in the pro files responsible for creating the binary(mzroll.pro, mavenTests.pro, peakdetector.pro). Hence, this flag is not needed here. 

Hope I answered your question

shubhra-agrawal: Got it. Thanks!

tests/MavenTests/MavenTests.pro
rish9511: Done

shubhra-agrawal:
@saifulbkhan Benchmarking can be done as part of testing. It's there in the testing doc I think.

PR #916, opened 2018-11-14
Alignment ui fixes

rish9511:
This PR brings the following the changes:

1. Club different alignment buttons under popup button
2. Rename different alignment widgets according to their functionality
3. Allow alignment widgets to be closed from the widget itself


src/gui/mzroll/forms/alignmentdialog.ui
```diff
@@ -324,7 +324,7 @@
         </item>
         <item>
          <property name="text">
-          <string>Loess Fit</string>
+          <string>Loess Fit  [Deprecated]</string>
```
shubhra-agrawal: Would suggest moving Loess Fit to be the third option. Also, Either change the "OBI-Warp (beta)" to "OBI-Warp [beta]" or change this to "(deprecated)".

src/gui/mzroll/samplertwidget.cpp
```diff
@@ -32,16 +32,17 @@ AlignmentPolyVizDockWidget::AlignmentPolyVizDockWidget(MainWindow *mw) :
...
 
-AlignmentPolyVizDockWidget::~AlignmentPolyVizDockWidget()
+SampleRtWidget::~SampleRtWidget()
 {
     delete ui;
 }
 
-void AlignmentPolyVizDockWidget::plotGraph() {
-
-    if (!_mw->alignmentPolyVizDockWidget->isVisible()) return;
```
shubhra-agrawal: This check was added to prevent ui freeze in case of checking/unchecking samples when there are 100 samples loaded in El-MAVEN. The plots are being updated even when the widget is closed. Will this case be handled in this PR?

src/gui/mzroll/alignmentvizallgroupswidget.cpp
```diff
@@ -1,9 +1,27 @@
 #include "alignmentvizallgroupswidget.h"
+#include <QDockWidget>
+#include <QToolBar>
 
 using namespace std;
 
-AlignmentVizAllGroupsWidget::AlignmentVizAllGroupsWidget(MainWindow* mw) {
-    this->_mw = mw;
+AlignmentVizAllGroupsWidget::AlignmentVizAllGroupsWidget(MainWindow* mw, QDockWidget* dockWidget):
```
shubhra-agrawal: This plot is not visible till alignment is run, even though there is no dependence on the algorithm. Is the plotting function just being called after alignment? Should be called as soon as the widget opens, ideally.

shubhra-agrawal: I think the Sample rt widget has the same problem but I haven't looked at the code to know if it is independent of the algorithm. I'm guessing it is. In which case, it should be plotted even without alignment.

src/gui/mzroll/samplertwidget.cpp
rish9511: Thanks for pointing this out. 
I'll put it back

src/gui/mzroll/forms/alignmentdialog.ui
rish9511: >Would suggest moving Loess Fit to be the third option.

Yes, makes sense

>Also, Either change the "OBI-Warp (beta)" to "OBI-Warp [beta]" or change this to "(deprecated)".

Sure

src/gui/mzroll/mainwindow.cpp
```diff
@@ -320,18 +319,24 @@ using namespace mzUtils;
...
 	isotopePlotDockWidget->setWidget(customPlot);
 
 	//treemap	 = 	  new TreeMap(this);
 	//peaksPanel	= new TreeDockWidget(this,"Group Information", 1);
 	spectraDockWidget = createDockWidget("Spectra", spectraWidget);
-	alignmentVizDockWidget = createDockWidget("AlignmentVisualization", alignmentVizPlot);
-	alignmentVizAllGroupsDockWidget = createDockWidget("AlignmentVisualizationForAllGroups", alignmentVizAllGroupsPlot);
-	pathwayDockWidget = createDockWidget("PathwayViewer", pathwayWidget);
+
+    groupRtDockWidget = createDockWidget("Per Group Alignment Visualization", groupRtVizPlot);
```
shubhra-agrawal: Please change the widget names (on the UI) to "Per Group Deviation", "All Groups Deviation" and "Sample Deviation"

src/gui/mzroll/forms/alignmentdialog.ui
rish9511: Done

src/gui/mzroll/samplertwidget.cpp
rish9511: Done

src/gui/mzroll/mainwindow.cpp
rish9511: Done

rish9511:
@saifulbkhan 
> can we have dock widget titles for all alignment widgets?

Thanks for pointing it out. This has been fixed with the latest commit.

PR #917, opened 2018-11-14
Functionality test for OBI-WARP

rish9511:


tests/MavenTests/testMzAligner.cpp
```diff
@@ -44,15 +35,65 @@ void TestMzAligner::testDoAlignment() {
...
+    mavenparameters->samples = samples->alignmentSamples;
+    mavenparameters->clsf = new ClassifierNeuralNet;
+    mavenparameters->clsf->loadModel("bin/default.model");
+
+    ObiParams params("cor", false, 2.0, 1.0, 0.20, 3.40, 0.0, 20.0, false, 0.60);
+    Aligner aligner;
+    aligner.alignWithObiWarp(mavenparameters->samples, &params);
+
+    DBS.loadCompoundCSVFile("bin/methods/KNOWNS.csv");
+    vector<Compound*> compounds = DBS.getCopoundsSubset("KNOWNS");
```
shubhra-agrawal: I think it's high time we correct this typo "getCompoundSubset" :P

tests/MavenTests/testMzAligner.cpp
```diff
@@ -44,15 +35,65 @@ void TestMzAligner::testDoAlignment() {
...
+                    sampleOriginalRt.push_back(peak.getScan()->originalRt);
+                    sampleNewRt.push_back(peak.getScan()->rt);
+                }
+            }
+            for(int rtCount=0; rtCount<sampleOriginalRt.size(); rtCount++) {
+                 float newDiff = std::abs(refSampleOriginalRt - sampleNewRt[rtCount]);
+                 float originalDiff = std::abs(refSampleOriginalRt - sampleOriginalRt[rtCount]);
+                 if(newDiff <= originalDiff)
+                     QVERIFY(true);
+                 else if((int)(newDiff) <= (int)(originalDiff))
```
shubhra-agrawal: Would suggest only keeping the int comparison if there's a binary true/false scenario.

tests/MavenTests/testMzAligner.cpp
```diff
@@ -44,15 +35,65 @@ void TestMzAligner::testDoAlignment() {
...
+    peakDetector.processSlices(slices, "compounds");
 
-    vector<mzSample*> samplesToLoad;
-    for (int i = 0; i <  files.size(); ++i) {
-        mzSample* sample = new mzSample();
-        sample->loadSample(files.at(i).toLatin1().data());
-        samplesToLoad.push_back(sample);
+    for(PeakGroup grp: mavenparameters->allgroups) {
+
+        if(grp.getName() == "NAD+" || grp.getName() == "methionine" || grp.getName() == "dTMP") {
```
shubhra-agrawal: The test condition looks good but I have concerns around the number of cases that will be covered using these 3 metabolites. Ideally, I would expect the following cases to be tested for overfitting and underfitting:
- Low and high intensity peaks in a group
- All high or low intensity peaks in a group
- Low resolution peaks

If we have such groups in these samples, then it's a small change. If not, let's think about changing the samples later.

src/core/libmaven/mzAligner.cpp
```diff
@@ -483,7 +484,7 @@ void Aligner::alignSampleRts(mzSample* sample, vector<float> &mzPoints,ObiWarp&
     }
 }
 
-void Aligner::alignWithObiWarp(vector<mzSample*> samples,  ObiParams* obiParams,int referenceSampleIndex){
+void Aligner::alignWithObiWarp(vector<mzSample*> samples,  ObiParams* obiParams, int referenceSampleIndex) {
```
saifulbkhan: Just as a reminder, we decided that our opening braces for functions, methods and namespaces will be on the next line. Space-separation for all other opening braces.

tests/MavenTests/testMzAligner.cpp
```diff
@@ -44,15 +35,65 @@ void TestMzAligner::testDoAlignment() {
...
+                else {
+                    sampleOriginalRt.push_back(peak.getScan()->originalRt);
+                    sampleNewRt.push_back(peak.getScan()->rt);
+                }
+            }
+            for(int rtCount=0; rtCount<sampleOriginalRt.size(); rtCount++) {
+                 float newDiff = std::abs(refSampleOriginalRt - sampleNewRt[rtCount]);
+                 float originalDiff = std::abs(refSampleOriginalRt - sampleOriginalRt[rtCount]);
+                 if(newDiff <= originalDiff)
+                     QVERIFY(true);
```
saifulbkhan: Most assertion checkers can take in any expression that eventually evaluates to a single boolean, and if thats the case with `QVERIFY` then I think this set of checks can be be simplified to:
```C++
QVERIFY(newDiff <= originalDiff
        || (int)(newDiff) <= (int)(originalDiff))
```
I have not tried this myself though.

tests/MavenTests/testMzAligner.cpp
rish9511: Done

tests/MavenTests/testMzAligner.cpp
rish9511: Done

tests/MavenTests/testMzAligner.cpp
rish9511: For unknown reasons, while running  MavenTests  the number of groups found by `processSlices` is way too less compared to the the groups found in GUI. If its not a too much of a problem, can we do this in the next release ?

shubhra-agrawal: Could you check if this is the case in older releases as well? We can write tests for alignment after the release but if something in peak detection has changed in the core from the previous release, we should fix it. We have clients using CLI regularly.

rish9511: This has nothing to do with the CLI. My guess is that I am not setting the parameters correctly but I am not sure

shubhra-agrawal: Cool. Leave this for now, then

PR #919, opened 2018-11-16
Fix Alignment dialog's progress bar

rish9511:


src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -558,9 +560,9 @@ void BackgroundPeakUpdate::processSlices(vector<mzSlice*>&slices,
         writeCSVRep(setName);
 }
 
-void BackgroundPeakUpdate::qtSlot(const string& progressText, unsigned int completed_slices, int total_slices)
+void BackgroundPeakUpdate::qtSlot(const string& progressText, unsigned int progress, int totalSteps)
```
shubhra-agrawal: Could do with a better name.

shubhra-agrawal: Could do with a better name.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -319,6 +319,8 @@ void BackgroundPeakUpdate::alignWithObiWarp(){
         Q_EMIT(updateProgressBar("Aligning Samples", 0, 0));
```
shubhra-agrawal: Not sure if we need this. You could display 1% to indicate that some process has started before the boost signals start working. But the oscillating progress bar is not ideal.

shubhra-agrawal:
Have suggested some minor changes. But overall, it looks good!

PR #920, opened 2018-11-16
Allow users to choose reference sample

rish9511:


src/core/libmaven/mzAligner.cpp
```diff
@@ -484,23 +485,25 @@ void Aligner::alignSampleRts(mzSample* sample, vector<float> &mzPoints,ObiWarp&
...
 
-    if(referenceSampleIndex < 0){
-        /**
-         * currently reference sample is choosen randomly,
-         * TODO: give user options to choose reference sample and pass index of
-         * that sample as referenceSampleIndex
-         */
+void Aligner::setRefSample(mzSample* sample)
+{
+    if(sample != nullptr)
```
shubhra-agrawal: No need to print the sample name, no? This could be an inline function, in that case.

src/gui/mzroll/alignmentdialog.cpp
```diff
@@ -93,6 +103,13 @@ void AlignmentDialog::intialSetup() {
 	algoChanged();
 	minIntensity->setValue(_mw->mavenParameters->minIntensity);
 	maxIntensity->setValue(_mw->mavenParameters->maxIntensity);
+
+    samplesBox->clear();
+    samplesBox->addItem("Select Randomly",QVariant(QVariant::fromValue(static_cast<void*>(nullptr))));
+    for(auto sample: _mw->samples) {
+        if(sample->isSelected)
+            samplesBox->addItem(sample->sampleName.c_str(), QVariant(QVariant::fromValue(static_cast<void*>(sample))));
```
saifulbkhan: Can we do a sort-by-name on these samples once before adding them to the drop-down list? On my system the samples are appearing in the list in no particular order.

rish9511: I agree that the samples should be displayed in some order but I don't think this is the right place to sort them. The order of samples is messed up in other parts as well. 
Ideally `_mw->samples` should be stored in sorted order or in the way user orders them. This ordering should be displayed in the `ProjectDockWidget` and others widget should be synced with it.

PR #921, opened 2018-11-26
New "projectDB" module

saifulbkhan:
This module contains the schema for (upcoming) SQLite-based project saving feature and the following classes:
- `Connection` (vaguely representing connection to a SQLite database).
- `Cursor` (vaguely representing an executable query).
- `ProjectDatabase` (main utility class which will be used by El-MAVEN to save/load data).

Since this is a sub-module meant for longer term use in this project, comprehensive tests would make it easy to trust and maintain the interfaces provided by it. The following tests have been planned but are yet to be written:

- [ ] Functional tests for primary save/load methods of `ProjectDatabase` class.
- [ ] Unit tests for all methods of all classes.

Reviewers are requested to only review this patch for design and completeness of the classes/module. The correctness of the same should be verified by tests, which will be gradually added to this PR or as a separate PR.

### DIFF
--- a/src/core/libmaven/Compound.h
+++ b/src/core/libmaven/Compound.h
@@ -58,6 +58,10 @@ class Compound{
         string pubchem_id;      /**@param  -  pubchem_id -    PubChem id*/
         string hmdb_id;         /**@param  -  hmdb_id-    Human Metabolome Database id */
         string alias;       /**@param   -  alias name of compound   */
+
+        // TODO: from MAVEN (upstream), find out what this is
+        string smileString;
+
         /**
         *@param -  srmId will hold filterLine string from mzxml file which represent type of
         *mass spec and ionization is used and other info as well
@@ -73,6 +77,17 @@ class Compound{
         float precursorMz;	/**@param  -  QQQ parent ion  mz value   */
         float productMz;    /**@param  -  QQQ child ion   mz value */
         float collisionEnergy; /**@param  -   QQQ collision energy of this compound   */
+        float logP; // TODO: find out what value this is.
+
+        // TODO: from MAVEN (upstream), find out what this is
+        bool virtualFragmentation;
+
+        // TODO: from MAVEN (upstream) decoy compound?
+        bool isDecoy;
+
+        // TODO: from MAVEN (upstream). Can this be derived somehow.
+        // Also maybe use an enum.
+        int ionizationMode;
 
         string db;			/**@param -   name of database for example KEGG, ECOCYC.. etc..    */
 
--- a/src/core/libmaven/Fragment.h
+++ b/src/core/libmaven/Fragment.h
@@ -2,7 +2,7 @@
 #define FRAGMENT_H
 #include <vector>
 #include "Peptide.hpp"
-#include "Scan.h"
+
 class PeptideRecord;
 class Scan;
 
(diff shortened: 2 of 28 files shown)


chubukov:
Hi @saifulbkhan . We currently also save our output in a structured database, and have been talking about direct writing to the  from Maven for a while.  Is there any documentation you've put together around what functionality you're going to enable with SQLite project saving? Perhaps we can also take advantage of it (perhaps with minor modifications/additions).

saifulbkhan:
Hi @chubukov,

For this release most of our implementation is around covering more or less the exact schema (with minor naming changes) from upstream [MAVEN](https://github.com/eugenemel/maven) project. But we do plan to eventually extend it to enable complete session restores and overall as a better output format than mzroll.

Here's a gist of what we are going to store in our "emDB" SQLite project:
1. **Sample** (metadata) - We store sample names and their current file paths for samples loaded in a session (if the file at these paths no longer exist in a later session when reloading project, the user will be able to provide them through the UI). Apart from this cohort set, selection status and other session specific information is also stored.
2. **Scan** data - we are not using scan information for saving or restoring our sessions for now, but I think it might be of use for downstream analysis (we guess that to be the reason because of save methods but no load ones in upstream MAVEN's DB project code).
3. **Peak groups** (and their **Peaks**) - Most non-derivable information for peak groups (from all peak tables and bookmark table) and their peaks will be saved so as to recreate the data structure in a new El-MAVEN session. A few calls to group statistics methods need to be performed (after loading groups and their peaks) so as to fill in derivable information about the group.
4. **Compounds** - Again compound information enough to recreate the data structure is stored in the database. But only compounds associated to peak groups are stored. Some specific scenarios need to be covered around missing database(s), duplicity of compounds, etc. to fully achieve session restore.
5. **Alignment data** - Original retention time and updated retention time for each scan in each sample is stored in the database. This seems like an overkill considering Eugene has been able to store every 200th scan and [perform](https://github.com/eugenemel/maven/blob/master/src/maven/projectDB.cpp#L522) some form of [segmented alignment](https://github.com/eugenemel/maven_core/blob/15ee8205dded519b852df0ab1d8ce8fd8146a867/libmaven/mzAligner.cpp#L358) algorithm. Still I think, our way is fair enough for a first start.

For a complete set of attributes capable of currently being stored please have a look at the [schema](https://github.com/ElucidataInc/ElMaven/blob/7364ca0b3fabd1516c0e7abcd17b62049f89bc68/src/projectDB/schema.h) attached to this PR. Some of them are simply initialized to default values for properties that we do not support yet but were present in MAVEN's schema. We do plan to go over them soon and figure out their benefits.

Hope this helps.

src/projectDB/connection.h
```diff
@@ -0,0 +1,124 @@
...
+     * @brief Obtain the absolute path to the connected database.
+     * @return Path as a string.
+     */
+    std::string dbPath();
+
+private:
+    /**
+     * @brief Stores absolute path for the connected SQLite database file.
+     */
+    std::string _dbPath;
```
rish9511: Since this is a private member and you are providing a public function to allow others to access it, this should be declared as a `const`. We don't want it to be modified by others right ?

src/core/libmaven/Compound.h
```diff
@@ -58,6 +58,10 @@ class Compound{
         string pubchem_id;      /**@param  -  pubchem_id -    PubChem id*/
         string hmdb_id;         /**@param  -  hmdb_id-    Human Metabolome Database id */
         string alias;       /**@param   -  alias name of compound   */
+
+        // TODO: from MAVEN (upstream), find out what this is
+        string smileString;
```
shubhra-agrawal: smiles is a text notation to store a metabolite's structure. This information is probably present in a .MSP file because it has a lot of compound specific information. Just fyi

src/core/libmaven/Fragment.h
```diff
@@ -95,4 +95,92 @@ class Scan;
         bool operator<(const Fragment* b) const;
         bool operator==(const Fragment* b) const;
     };
+
+    // TODO: from MAVEN (upstream). find our what this is.
+    struct FragmentationMatchScore {
```
shubhra-agrawal: This is related to the Database feature we're going to add. When we match the fragmentation pattern from the MSP file and our samples, there will be a score calculated based on the extent of match. This is meant to store that information.

src/core/libmaven/Scan.h
```diff
@@ -29,6 +29,17 @@ class Scan
     */
     inline unsigned int nobs() const { return mz.size(); }
```
shubhra-agrawal: Is the m/z array sorted? or is that an assumption we're making

src/core/libmaven/mzMassCalculator.h
```diff
@@ -37,6 +38,9 @@ class MassCalculator {
     public:
         enum IonizationType { ESI=0, EI=1};
         static IonizationType ionizationType;
```
shubhra-agrawal: El-MAVEN does not do adduct calculation yet. Not sure if we need this piece of code for now. You can leave it in for this release but make a note to remove this later?

src/projectDB/projectDB.pro
```diff
@@ -0,0 +1,25 @@
...
+QMAKE_LFLAGS += -L$$top_builddir/libs/
+
+LIBS += -lmaven \
+        -fopenmp \
+        -lboost_filesystem \
+        -lsqlite3
+
+macx {
+    LIBS -= -fopenmp
+}
```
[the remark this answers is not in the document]
saifulbkhan: This has been fixed in the PR #922. Thanks for the suggestions.

src/projectDB/connection.h
saifulbkhan: There is no setter for this variable, so I don't think there is a way of modifying this from anywhere outside the class.

rish9511: > there is no setter for this variable, so I don't think there is a way of modifying this from anywhere outside the class. 

Technically yes, no other class will be able to modify the variable. However, what I meant was,we never actually want  any other class  to change the Database path . So, even the copy should be immutable.

src/core/libmaven/Scan.h
saifulbkhan: An assumption, which, now that I think of it, we should not be making. Returning the `std::min` and `std::max` of `mz` vector is definitely the right way of doing this.

src/projectDB/connection.h
rish9511: Unless you know  how to implement this correctly, lets leave this as it is for now. 
I don't understand how to implement it  correctly. Sorry for the trouble

src/core/libmaven/mzUtils.h
```diff
@@ -39,7 +39,9 @@
 #define FLOATROUND(x, sd) (int(x * sd + 0.5) / sd)
 
 //TODO: Kiran-Sahil, Added while merging spectrahitstable
-#define EMASS 		 0.00054857971 //(6.022141×10^23 ⋅ 9.10938×10^-28 )
```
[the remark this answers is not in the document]
saifulbkhan: Will remove these three (re)definitions in that case.

src/core/libmaven/mzMassCalculator.h
saifulbkhan: I will add a todo in the code itself.

PR #922, opened 2018-11-26
[Ftr] SQLite based project

saifulbkhan:
This PR includes implementation of a new SQLite-based project feature. Users will be able to save their progress for a session in a new **"emDB"** format and load these files to continue where they left off from the saved session. This offers (hopefully a more stable) alternative to "mzroll" files that El-MAVEN (and MAVEN) have used to save project data. The implementation depends on the `projectDB` module proposed in PR #921. Ideally the user should be able to restore their progress from a previous session as completely as possible using these "emDB" files. It is worth mentioning that this feature was inspired (and a lot of code was borrowed) from the upstream branch of [MAVEN](https://github.com/eugenemel/maven) (latest v8.0.9).

### DIFF
--- a/mzroll.pri
+++ b/mzroll.pri
@@ -35,6 +35,8 @@ unix: {
     LIBS +=  -lboost_signals -lErrorHandling -lobiwarp
 }
 
+LIBS += -lboost_system -lboost_filesystem -lsqlite3
+
 #INSTALL_LIBDIR = $$(INSTALL_LIBDIR)
 #unix {
 #  !mac {
--- a/mzroll.pri
+++ b/mzroll.pri
@@ -35,8 +35,6 @@ unix: {
     LIBS +=  -lboost_signals -lErrorHandling -lobiwarp
 }
 
-LIBS += -lboost_system -lboost_filesystem -lsqlite3
-
 #INSTALL_LIBDIR = $$(INSTALL_LIBDIR)
 #unix {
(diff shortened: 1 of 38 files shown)


rish9511:
IMO, a lot more can be done to make this PR better. The changes I have in my mind are mostly related to refactoring. These are the following places which required refactoring imo

1. There should be no `autosave` class. Atm, when we are saving the project it's `autosave` that starts the background thread  and then uses `mzFileIO` for all the saving work.  To me it does not make sense to have and additional class(`autosave`) whose job is just to start a thread and nothing else. We already have `mzFileIO` for that, why not just use it ?

2. Function `SaveProjectSqlite` should not be part of class `ProjectDockWidget`.  I think the reason why it's placed there is because there are other similar functions in class `ProjectDockWidget`.  Functionality of `ProjectDockWidget` is to show the loaded samples etc. Although, i agree that the function `SaveProjectSqlite` is not itself responsible for saving anything but still I believe it should not be part of `ProjectDockWidget`.   
I remember having a discussion with respect to this, where we discussed having an IO class for reading and writing work. So, maybe we can use the `mzFileIO` class and have a subclass that does the saving and loading work 


Why I think this is important is because we already have a lot of unstructured code and this is going to add more to it.  

This is  for future reference though, I know we don't have enough time to make all these changes and hence it has to be skipped for now.

PR #943, opened 2018-12-19
Autosave session progress in emDB without prompts

saifulbkhan:
Whenever El-MAVEN crashes, the user risks losing their progress for that session. This calls for a need to have a somewhat fail-safe autosaving mechanism that tries to minimize this loss in the event of an abrupt application exit. A timestamped autosave file will now be generated the first time a user bookmarks or marks a peak group as good or bad. This file will be used to constantly save any further curation done by the user in that session. Finally a prompt will appear at the end of a session asking the user if they want to save it as a named emDB file.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -93,7 +93,7 @@ void MainWindow::printvalue() {
         const QString timestamp = nowTime.toString(QLatin1String("yyyyMMdd-hhmmsszzz"));
         QString AutosavePath = samplePath + timestamp + ".mzroll";
         this->_currentProjectName = AutosavePath;
-        this->autosaveEnabled = true;
+        this->timestampFileExists = true;
         this->saveProject();
         unsigned int countCrashState = 0;
         settings->beginWriteArray("crashTables");
@@ -827,7 +827,7 @@ QString MainWindow::_newAutosaveFile()
 AutoSave::AutoSave(MainWindow* mw)
 {
     _mainwindow = mw;
-    _mainwindow->autosaveEnabled = false;
+    _mainwindow->timestampFileExists = false;
     saveTablesOnly = false;
 }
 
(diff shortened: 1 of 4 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -892,26 +880,36 @@ void MainWindow::_setProjectFilenameFromProjectDockWidget()
...
-            if (this->fileName.isEmpty()) {
-                this->doAutosave = false;
-                return;
-            }
-            autosave->saveProjectWorker();
+    if (this->peaksMarked == 1) {
+        if (!this->autosaveEnabled)
+            this->_currentProjectName = this->_newAutosaveFile();
+
+        if (this->_currentProjectName.isEmpty()) {
```
shubhra-agrawal: When would this be true?

saifulbkhan: It seems it will never be but I try to not remove existing checks like these because they might have been put to cover some not-so-apparent edge case. I will remove it if you still feel its unnecessary.

shubhra-agrawal: Oh wait. I thought this check was inside the previous `if` block and hence will never be true. My bad.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -892,26 +880,36 @@ void MainWindow::_setProjectFilenameFromProjectDockWidget()
...
-        this->doAutosave = this->askAutosave();
-        if (this->doAutosave) {
-            _setProjectFilenameIfEmpty();
-            if (this->fileName.isEmpty()) {
-                this->doAutosave = false;
-                return;
-            }
-            autosave->saveProjectWorker();
+    if (this->peaksMarked == 1) {
+        if (!this->autosaveEnabled)
```
shubhra-agrawal: 'autosaveEnabled': Please change the variable name to something more explanatory.

shubhra-agrawal: 'autosaveEnabled': Please change the variable name to something more explanatory.

src/gui/mzroll/mainwindow.h
```diff
@@ -284,11 +284,9 @@ Q_OBJECT
 	//TODO: Sahil - Kiran, removed while merging mainwindow
 	// bool isSampleFileType(QString filename);
 	// bool isProjectFileType(QString filename);
-	bool askAutosave();
     void saveProject(bool explicitSave=false);
     void saveProjectForFilename(bool tablesOnly=false);
-    bool doAutosave;
-	int askAutosaveMain;
+    bool autosaveEnabled;
```
shubhra-agrawal: Function documentation is required.

PR #948, opened 2018-12-28
Add GA triggers

shubhra-agrawal:
Following triggers have been added for Google Analytics:
- user clicks on Polly upload button, which app was selected
- user performs alignment, which algorithm was selected
- user exports data intentionally, which format, UI point of origin
- user creates a new peak table, or adds the first group to a bookmark table
- user successfully loads a custom compound DB, MS or MRM
- user merges bookmarked peaks into a peak table, successful or not

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -3151,7 +3151,8 @@ void MainWindow::Align() {
 
 	BackgroundPeakUpdate* workerThread;
 
-    if(alignmentDialog->alignAlgo->currentIndex() == 1){
+    if (alignmentDialog->alignAlgo->currentIndex() == 1) {
+		analytics->hitEvent("Alignment", "Obi-Warp");
 		workerThread = newWorkerThread("alignWithObiWarp");
 		workerThread->setMavenParameters(mavenParameters);
 		alignmentDialog->setWorkerThread(workerThread);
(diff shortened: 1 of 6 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -1203,6 +1203,7 @@ TableDockWidget* MainWindow::addPeaksTable(QString title) {
         customTableId = std::atoi(idString.c_str());
     }
     TableDockWidget* panel = new PeakTableDockWidget(this, customTableId);
+	analytics->hitEvent("New Table", "Peak Table");
```
saifulbkhan: There's another method used to add peak tables going by the name `MainWindow::createPeakTable`. Not sure if we should put GA on that one because its only used when loading mzroll tables saved from previous sessions. Just wanted to point this out.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -396,8 +396,6 @@ void TableDockWidget::deleteAll() {
 
 void TableDockWidget::showAllGroups() {
   treeWidget->clear();
-  _mainwindow->getAnalytics()->hitEvent("PeaksTable", "ShowAllGroups",
```
saifulbkhan: Extremely unhelpful trigger. Good job removing this.

PR #949, opened 2018-12-28
Analytics overhaul v0.6.1

saifulbkhan:


src/gui/mzroll/mainwindow.cpp
```diff
@@ -1985,56 +1993,83 @@ bool MainWindow::loadMetaInformation(QString filename) {
...
+                 ++it) {
+                notFoundColumns += "\n" + *it;
+            }
+            msgBox.setDetailedText(QString::fromStdString(notFoundColumns));
+            msgBox.setWindowFlags(msgBox.windowFlags()
+                                  & ~Qt::WindowCloseButtonHint);
+        }
+        analytics->hitEvent("Load Compound DB",
+                            "Column Error",
+                            DB.notFoundColumns.size());
```
shubhra-agrawal: the number of unknown columns might not be of use when analysing this data. Maybe the value can be whether the db was successfully loaded or not. The `if` block is for successfull load but some additional/unknown columns were found. The `else` block is for error in upload because of missing/unknown columns.

shubhra-agrawal: Yeah, row error does not need a value.

PR #951, opened 2019-01-03
Fix certain emDB autosave issues

saifulbkhan:
This PR fixes the following issues pertaining to autosaving using emDB project format:
1. Fix explicit save operation using timestamp project name.
2. Fix threaded save not setting last saved project.
3. Fix no save before new session load.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -918,7 +918,14 @@ void MainWindow::explicitSave()
 
 void MainWindow::threadSave(QString filename)
 {
+    // a non-temporary file is being used to save session data,
+    // all autosave states should be discarded
+    if (timestampFileExists) {
+        resetAutosave();
+    }
+
     _currentProjectName = filename;
+    _latestUserProjectName = filename;
     autosave->saveProjectWorker();
 }
 
@@ -935,7 +942,7 @@ void MainWindow::saveProject(bool explicitSave)
             _setProjectFilenameFromProjectDockWidget();
 
         // if no projects were saved or opened
-        if (_loadedProjectName.isEmpty()) {
+        if (_latestUserProjectName.isEmpty()) {
             QString message = "Would you like to save your data for this "
                               "session as a project?";
             QMessageBox::StandardButton reply;
@@ -981,7 +988,7 @@ void MainWindow::saveProject(bool explicitSave)
                 _setProjectFilenameIfEmpty();
                 analytics->hitEvent("Project Save", "emDB In New File");
             } else if (msgBox.clickedButton() == saveButton) {
-                _currentProjectName = _loadedProjectName;
+                _currentProjectName = _latestUserProjectName;
                 analytics->hitEvent("Project Save", "emDB In Current File");
             } else {
                 return;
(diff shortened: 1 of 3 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -908,6 +908,11 @@ void MainWindow::autosaveProject()
 
 void MainWindow::explicitSave()
 {
+    // the user is ordering an explicit save, get rid of autosave states
```
shubhra-agrawal: @saifulbkhan If there is a crash during the explicit save process, we might lose all data. Would be safer to delete timestamp files once the explicit save is done successfully. Thoughts?

saifulbkhan: Good catch. There will always be a "code-distance" between scrapping of timestamp file and saving of a new file (running in the save thread) that can fail and leave the user with neither of the files on their system. But we can minimize this for an apparent fail-safe process. Will make some changes and push soon.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -988,7 +1000,8 @@ void MainWindow::saveProject(bool explicitSave)
         this->autosave->saveProjectWorker();
     } else if (explicitSave) {
         _setProjectFilenameFromProjectDockWidget();
-        if (_loadedProjectName.isEmpty()) {
+        cerr << _currentProjectName.toStdString() << endl;
```
saifulbkhan: This should not be here.

src/gui/mzroll/mainwindow.cpp
saifulbkhan: It turns out this here is more or less the optimum place where we should reset before explicit save.

shubhra-agrawal: Is that because the currentProjectName is updated during explicitSave?

saifulbkhan: Yes. And if you call the `resetAutosave` method after initiating a threaded save, the behavior would be undefined since `_currentProjectName` is used by the worker thread.

shubhra-agrawal: Got it. Doesn't make sense to handle that in this PR then.

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -623,6 +623,13 @@ void ProjectDockWidget::setLastSavedProject(QString filename)
     _lastSave = std::chrono::system_clock::now();
 }
 
+bool ProjectDockWidget::_userSQLiteProjectIsCurrent()
```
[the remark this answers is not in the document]
saifulbkhan: That does sound more intuitive, even readability-wise.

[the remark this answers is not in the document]
saifulbkhan: Eh... but that is not what the method checks for though (i.e. if there's unsaved changes or not). Maybe `_currentProjectIsTemporary` and change the second condition to its boolean opposite?

```C++
bool ProjectDockWidget::_currentProjectIsTemporary()
{
    auto sqliteProjectIsOpen = _mainwindow->fileLoader->sqliteProjectIsOpen();
    auto sqliteProjectIsTemporary = _mainwindow->timestampFileExists;
    return userProjectIsCurrent && sqliteProjectIsOpen;
}
```

I come up with the most horrible names.

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -659,7 +666,7 @@ void ProjectDockWidget::saveSQLiteProject(QString filename)
 
 void ProjectDockWidget::saveSQLiteProject()
 {
-    if (_mainwindow->fileLoader->sqliteProjectIsOpen()) {
+    if (!_currentProjectIsTemporary()) {
```
shubhra-agrawal: if `sqliteProjectisOpen` is false, then it'll try to `getLastOpenedProject()`. Is that what you intended?

PR #954, opened 2019-01-07
Disable alignment for samples containing MS2 scans

saifulbkhan:
OBI-Warp alignment takes too long for MS2 scans and appears to have frozen the application. Moreover, we have not confirmed if it actually works for such datasets. This commit disables alignment while any samples loaded contain MS/MS scans.

### DIFF
--- a/src/gui/mzroll/alignmentdialog.cpp
+++ b/src/gui/mzroll/alignmentdialog.cpp
@@ -200,6 +200,22 @@ void AlignmentDialog::algoChanged()
     samplesBox->setVisible(obiWarp);
     refSampleLabel->setVisible(obiWarp);
 
+    auto samples = _mw->getSamples();
+    auto ms2SamplesDetected = false;
+    for (const auto sample : samples) {
+        if (sample->ms2ScanCount() > 0) {
+            showInfo("Obi-warp does not work with MS2 scans at the moment.\nWe "
+                     "will inform you once this functionality has been added.");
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/alignmentdialog.cpp
```diff
@@ -200,6 +200,22 @@ void AlignmentDialog::algoChanged()
...
+        if (sample->ms2ScanCount() > 0) {
+            showInfo("Obi-warp does not work with MS2 scans at the moment.\nWe "
+                     "will inform you once this functionality has been added.");
+            ms2SamplesDetected = true;
+            alignButton->setDisabled(true);
+            break;
+        }
+    }
+    if (!ms2SamplesDetected || !obiWarp) {
+        alignButton->setDisabled(false);
```
shubhra-agrawal: Correct me if I'm wrong but this message will appear no matter which algorithm is selected, right? And the alignment button will be disabled for all algorithms even though the message is about Obi-warp specifically. The button will also stay disabled even if the user changes the algorithm they have selected.

saifulbkhan: Yes but it will be corrected (and button will be enabled) once control reaches this later check (`!ms2SamplesDetected || !obiWarp`). The user should not notice the correction at all.

shubhra-agrawal: Got it. Missed the check for obiwarp, sorry!

PR #959, opened 2019-01-11
[Fix] Show error message for obi-warp instead of exiting

shubhra-agrawal:
The app exits abruptly if a  sanity check in obi-warp (3rd party) fails. This leads to frequent crashes during alignment. This is non-deterministic and alignment tends to work if the process is re-started.
Until the reason for this process being non-deterministic is found, an error message has been put in place that will cancel the alignment if the check fails and ask the user to try again. This is just a temporary solution to prevent data loss.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -1117,12 +1117,11 @@ void MainWindow::savePeakTableAsMzRoll(TableDockWidget* peaksTable,
     }
 }
 
-void MainWindow::showAlignmetErrorDialog(QString errorMessage)
+void MainWindow::showAlignmentErrorDialog(QString errorMessage)
 {
-    QErrorMessage alignmentErrorDialog(this);
-    alignmentErrorDialog.setWindowTitle("Alignment Error");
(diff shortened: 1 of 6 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -1117,12 +1117,11 @@ void MainWindow::savePeakTableAsMzRoll(TableDockWidget* peaksTable,
...
-void MainWindow::showAlignmetErrorDialog(QString errorMessage)
+void MainWindow::showAlignmentErrorDialog(QString errorMessage)
 {
-    QErrorMessage alignmentErrorDialog(this);
-    alignmentErrorDialog.setWindowTitle("Alignment Error");
-    alignmentErrorDialog.showMessage(errorMessage);
-    alignmentErrorDialog.exec();
+    QMessageBox alignmentError;
+    alignmentError.setText(errorMessage);
+    alignmentError.exec();
```
saifulbkhan: Let's use `alignmentError.open()` method for executing all modal dialogs (wherever possible) like we last agreed.

Issue #840, opened 2018-08-17
Decide on a style guide and source code formatting


saifulbkhan:
Different parts of ElMaven's codebase follow different styles of formatting. C++ is *not* a simple programming language, although there have been efforts to make it so in the recent (2011, 2014 and 2017) releases. Since it provides such a broad set of features and tries to cover different programming paradigms (procedural, OOP and even functional), we should probably decide on a concrete set of code formatting and style guidelines if we want to avoid having an ugly and inconsistent codebase, as it continues to grow in complexity.

From what I understand, the newer parts were written closely following along the Clang [(LLVM)](http://llvm.org/docs/CodingStandards.html) guidelines. Using a clang formatting plugin on VS Code (which itself uses the `clang-format` executable underneath), works well for the most part. But unless configured it formats using its own default style (two spaces for indentation, 80 character line length, breaking function params/args only when it exceeds 80, etc.).

While I have no issues with this particular style, it is worth noting that `clang-format` can be [configured](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) with a ".clang-format" config file present in the source directory (or parent directory). Some popular styles (other than LLVM) used by the C++ community and supported by `clang-format` (out-of-the-box) are:
- [Google C++ Style Guide](http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml)
- [Chromium's Style Guide](http://www.chromium.org/developers/coding-style)
- [Mozilla's Sttyle Guide](https://developer.mozilla.org/en-US/docs/Developer_Guide/Coding_Style)
- [Webkit's Style Guide](http://www.webkit.org/coding/coding-style.html)

We need not strictly follow either of these guides, but it is advisable to do so and there should be an appropriate reason if we chose to deviate from either of the well-defined standards. Please use this thread to have a discussion about which particular style & format we ourselves would like to see enforced in ElMaven.

On a side note, we can also maybe agree upon using a code linter while we are discussing formatting (open to suggestions on this).

saifulbkhan:
We now have a [style guide](https://github.com/ElucidataInc/ElMaven/blob/develop/STYLE_GUIDE.md) on develop (tagged under v0.7.0).

PR #966, opened 2019-01-31
Add a C++ code style guide for the project

saifulbkhan:
Added a style-guide (markdown) file and a file specifying styles that can be used with the `clang-format` tool for formatting C++ source files.

Issue: #840


STYLE_GUIDE.md
```diff
@@ -0,0 +1,213 @@
...
+
+// Correct
+if (someCondition) {
+    doSomething();
+} else {
+    doSomeOtherThing();
+    return -1;
+}
+```
+
```
saifulbkhan: Add a comment around whether having `if` condition on the same line as the statement within is acceptable or not.

PR #972, opened 2019-02-05
ProjectDB versioning

saifulbkhan:
This patch implements the following:
1. Add versioning utilities to "projectDB" module:
    - As the format of the database for storing information changes with subsequent releases, there might be cases where the table structure might need to be modified while keeping the data intact. A set of functions and data structures have been added in the module for allowing easy upgrades requiring minimal effort with each release involving an upgrade.
    - Upgrades can be made by writing SQL scripts that transform the data from one DB format to the next. For now, only backwards compatibility is guaranteed by these utilities, because only upgrades can be performed on database files from previous versions. Older releases of the library will have no way of knowing how to downgrade a newer version of the database to an older one that the older version of the application would be able to use.

2. Use versioning in `ProjectDatabase` class:
    - All ProjectDatabase instances will now be able to perform checks to evaluate whether they need a database upgrade and will automatically attempt to do so.
    - Since database upgrades can potentially change the database to be incompatible with older release in which it was first created, a backup is also saved of the original DB before upgrading.
    - For development builds, an extra check is made to ensure that the latest DB format available is used.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -3892,27 +3892,9 @@ void MainWindow::markGroup(PeakGroup* group, char label) {
 	//getPlotWidget()->scene()->update();
 }
 
-int MainWindow::versionCheck() {
-
-//define a MACRO for converting DEFINES to strings.. jfc!
-#define xstr(s) str(s)
-#define str(s) #s
-
-	QString hostname = "http://genomics-pubs.princeton.edu";
-	QString path = "/mzroll/vercheck.php?";
-	QString os = "os=" + QString(xstr(PLATFORM));
-	//QString ver = "ver=" + QString::number(EL_MAVEN_VERSION);
-	QString ver = "ver=620";
-	QString query = os + "&" + ver;
-
-	logWidget->append("Latest version check: " + hostname + path + query);
-
-	QDownloader* downloader = new QDownloader(this);
-	downloader->getPage(hostname + path + query);
-	connect(downloader, SIGNAL(downloadResult(QString)), logWidget,
-			SLOT(append(QString)));
-
-	return 0;
+QString MainWindow::appVersion() {
+        auto version = STR(EL_MAVEN_VERSION);
+        return QString(version);
 }
 
 void MainWindow::toggleIsotopicBarPlot(bool show)
--- a/src/projectDB/projectdatabase.cpp
+++ b/src/projectDB/projectdatabase.cpp
@@ -6,11 +6,45 @@
 #include "connection.h"
 #include "cursor.h"
 #include "mzSample.h"
+#include "projectversioning.h"
 #include "schema.h"
 
-ProjectDatabase::ProjectDatabase(const string& dbFilename)
+ProjectDatabase::ProjectDatabase(const string& dbFilename,
+                                 const string& version)
 {
     _connection = new Connection(dbFilename);
+
+    // figure out whether this database needs upgrade
+    using namespace ProjectVersioning;
+    auto aheadBy = 0;
+    auto versionInfo = extractVersionInfoFromTag(version, &aheadBy);
+    auto appVersion = Version(versionInfo.first);
+    auto requiredDbVersion = getDbVersionForApp(appVersion);
+    auto currentDbVersion = this->version();
+
+    // if build is ahead by non-zero number of commits, then its a dev build
+    if (aheadBy > 0) {
+        requiredDbVersion = getLatestDbVersion();
+    }
+
+    if (currentDbVersion != requiredDbVersion && !this->isEmpty()) {
+        // upgrade needed, close existing connection
+        delete _connection;
+
+        string upgradeScript = generateUpgradeScript(currentDbVersion,
+                                                     requiredDbVersion);
+        cout << "Debug: Upgrading database from v"
+             << currentDbVersion
+             << " to v"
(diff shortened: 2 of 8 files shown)


src/projectDB/projectversioning.h
```diff
@@ -0,0 +1,208 @@
...
+ * major, minor and patch levels (ref: htttps://semver.org). The most obvious
+ * reason for the existence of this struct is to allow comparison between two
+ * Version objects.
+ */
+struct Version {
+
+    /**
+     * @brief Create a version object from a version string of the form
+     * "MAJOR.MINOR.PATCH". If the string is not of this form, the returned
+     * Version object will most likely not get an incorrect representation than
```
shubhra-agrawal: Do you mean "..will most likely get an incorrect representation" ?

saifulbkhan: Yes I meant that. Sorry about that. Corrected it.

src/projectDB/projectversioning.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+    return false;
+}
+
+int Version::_compare(const Version& other) const
+{
+    int majorDiff = _major - other.major();
+    int minorDiff = _minor - other.minor();
+    int patchDiff = _patch - other.patch();
+    return majorDiff + minorDiff + patchDiff;
```
saifulbkhan: This method has a basic flaw. Comparing version `0.6.1` and `0.7.0` will yield 0 (indicating they are equal when they are not), because `minorDiff` will be -1 and `patchDiff` will be 1, cancelling each other out. Will make this correction

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3892,27 +3892,9 @@ void MainWindow::markGroup(PeakGroup* group, char label) {
...
-
-	logWidget->append("Latest version check: " + hostname + path + query);
-
-	QDownloader* downloader = new QDownloader(this);
-	downloader->getPage(hostname + path + query);
-	connect(downloader, SIGNAL(downloadResult(QString)), logWidget,
-			SLOT(append(QString)));
-
-	return 0;
+QString MainWindow::appVersion() {
```
rish9511: Can we please make this `static` ? There are  places where we need to get the app version without actually requiring the entire mainwindow instance

src/projectDB/projectdatabase.cpp
```diff
@@ -6,11 +6,45 @@
...
+        delete _connection;
+
+        string upgradeScript = generateUpgradeScript(currentDbVersion,
+                                                     requiredDbVersion);
+        cout << "Debug: Upgrading database from v"
+             << currentDbVersion
+             << " to v"
+             << requiredDbVersion
+             << endl;
+        upgradeDatabase(dbFilename, upgradeScript, version);
```
rish9511: Call only if required. There's no check here or in `upgradeDatabase`

src/gui/mzroll/mainwindow.cpp
saifulbkhan: Yes I can do that.

src/projectDB/projectdatabase.cpp
saifulbkhan: If control has reached this point, then it definitely does need an upgrade (and its a logical error if it does not). The check on [line 30](https://github.com/ElucidataInc/ElMaven/pull/972/files#diff-b65d4507eb26f772cc8ef1fef3e8c8b9R30) is the only check necessary.

rish9511: I was talking about the case where the `upgradeScript` might be empty.  From your point it seems as if, `generateUpgradeScript` will never return an empty string

src/projectDB/projectversioning.cpp
```diff
@@ -0,0 +1,257 @@
...
+        int dbVersion = entry.first;
+        string upgradeScript = entry.second;
+
+        // relying on keys of map being sorted for generating the correct script
+        if (dbVersion >= fromVersion && dbVersion < toVersion)
+            mergedUpgradeScript += upgradeScript + "\n\n";
+    }
+
+    return mergedUpgradeScript;
+}
```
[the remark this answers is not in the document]
rish9511: Not sure what's a good way to implement this. 
0->2 can contain a list of upgrade scripts from 0-1 and 1-2, so we won't have to manually write them again in such cases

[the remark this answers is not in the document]
rish9511: Anyway, I am resolving this for now. Not an immediate requirement

src/projectDB/projectdatabase.cpp
saifulbkhan: It might. When there is only a single table being added in the entire release. In that case an upgrade you are correct. An unnecessary upgrade should not be performed. I'll add a check in `upgradeDatabase` function, because we should still back-up user's original database and that piece of code should still be executed. Let me know if you disagree.

src/projectDB/projectversioning.cpp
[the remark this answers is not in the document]
saifulbkhan: But then instead of having a list of such scripts, I am using the `generateUpgradeScript` method to create it when needed. That way we avoid having maintaining such lists.

src/projectDB/projectdatabase.cpp
rish9511: Makes sense, thanks!

Issue #965, opened 2019-01-29
Repair code for reading MSP files


saifulbkhan:
**Description**
MSP (NIST's Mass spectral transfer) files can be used to read in MS/MS metabolites along with mz and intensity profiles of fragments arising from them. El-MAVEN has a function to allow reading these files but it does not read in as many fields as available and even when loaded is not used best used by the application (at least its not very evident how this information is used) for analyzing PRM datasets. Moreover, MSP files can be large in size and it blocks the GUI for the entire duration of load.

**Solution to the issues involving MSP load**
- [x] Add checks for more fields commonly available in MSP files.
- [x] Check for end of data stream and make sure the last compound in file is not missed.
- [x] Run load operation in a different thread from main thread to avoid UI freeze.
- [x] Add a status-bar elements informing the user about the load progress.
- [x] Do not save MSP filenames to be loaded automatically in the next session.
- [x] Figure out a way to load multiple compounds with the same name and ID but having different fragmentation profiles.
- [x] Show a prompt to the user about mismatched polarity of selected library and  loaded dataset.

shubhra-agrawal:
These changes have been merged to develop and will be part of v0.7 release.

Issue #842, opened 2018-08-17
default to last used folder for loading compound list 


chubukov:
When you click "load compound file" it looks like the default location is always the installation folder. It would be nice if it defaulted to the last location that you loaded a compound file from (similar to the way loading data files works).

shubhra-agrawal:
@chubukov This has been merged to develop and will be part of v0.7 release.

PR #973, opened 2019-02-05
Revamp code around loading NIST library files

saifulbkhan:
This PR includes the following changes:

1. Add checks for more fields available in MSP files.
2. Prevent last compound in MSP from getting lost.
3. Save last opened directory for compounds DB.
4. Disallow selecting multiple compound files while loading.
5. Load compound databases in a separate thread.
6. Show progress bar when loading NIST library files.
7. Add criteria for saving NIST library files as "lastDatabaseFile".
8. Do not show "Precursor Mz" information for compounds having fragmentation information.
9. Preserve all fragmentation records from MSP files.
10. Add methods to obtain compound type and check whether a database is an NIST library.

Issue: #965 
Issue: #842

### DIFF
--- a/src/gui/mzroll/database.cpp
+++ b/src/gui/mzroll/database.cpp
@@ -99,6 +99,7 @@ bool Database::addCompound(Compound* newCompound)
         compoundAdded = true;
     } else {
         bool matched = false;
+        bool newFragmentPattern = false;
         for (int i = 0; i < compoundsDB.size(); i++) {
             Compound* currentCompound = compoundsDB[i];
             bool sameID = currentCompound->id == newCompound->id;
@@ -141,20 +142,21 @@ bool Database::addCompound(Compound* newCompound)
                 if (equalMzs && equalIntensities && equalIonTypes) {
                     matched = true;
                 } else {
-                    // same compound but different fragmentation spectra, change
-                    // its name according to the number of compounds with the
-                    // same ID
-                    int loadOrder = prmIdCount.at(newCompound->id);
-                    newCompound->name = newCompound->name
-                                        + " ("
-                                        + to_string(loadOrder)
-                                        + ")";
-                    prmIdCount[newCompound->id] = ++loadOrder;
-                    break;
+                    newFragmentPattern = true;
                 }
             }
         }
         if (!matched) {
+            // existing compound but different fragmentation spectra, change
+            // its name according to the number of compounds with the same ID
+            if (newFragmentPattern) {
+                int loadOrder = prmIdCount.at(newCompound->id);
+                newCompound->name = newCompound->name
+                                    + " ("
+                                    + to_string(loadOrder)
+                                    + ")";
+                prmIdCount[newCompound->id] = ++loadOrder;
+            }
             compoundsDB.push_back(newCompound);
             compoundAdded = true;
         }
--- a/src/gui/mzroll/database.cpp
+++ b/src/gui/mzroll/database.cpp
@@ -426,6 +426,176 @@ map<string,int> Database::getDatabaseNames() {
 	return dbnames;
 }
 
+int Database::loadNISTLibrary(QString filepath,
+                              bsignal::signal<void (string, int, int)>* signal)
+{
+    QString filename = QFileInfo(filepath).fileName();
+    if (signal)
+        (*signal)("Preprocessing database " + filename.toStdString(), 0, 0);
+
+    qDebug() << "Counting number of lines in NIST Libary file…" << filepath;
+    ifstream file(filepath.toStdString());
+    file.unsetf(ios_base::skipws); // do not skip newlines
+    unsigned lineCount = std::count(istream_iterator<char>(file),
+                                    istream_iterator<char>(),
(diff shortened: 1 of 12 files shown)


src/gui/mzroll/mzfileio.cpp
```diff
@@ -223,112 +223,169 @@ PK$PEAK: m/z int. rel.int.
     } while (!line.isNull());
     return compoundCount;
 }
-//TODO: Shouldnot be here
-int mzFileIO::loadNISTLibrary(QString fileName) {
-    qDebug() << "Loading Nist Libary: " << fileName;
-    QFile data(fileName);
 
+int mzFileIO::loadNISTLibrary(QString filepath)
```
shubhra-agrawal: mzFileIO is not available to the core and therefore not to the CLI. In the future, we would like to add support for MS/MS data in CLI as well. Why was the decision made to add it here and not in one of the database files? I understand we will have to merge database.cpp and databases.cpp to prevent code duplication, but there are no such plans for mzFileIO as of now.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1862,7 +1862,15 @@ void MainWindow::_postCompoundsDBLoadActions(QString filename,
...
         analytics->hitEvent("Load Compound DB", "Successful Load", msLevel);
 
-        settings->setValue("lastDatabaseFile", filename);
+        // do not save NIST library files for automatic load when starting next
+        // session, unless they are less than 2Mb in size
+        QFileInfo fileInfo(filename);
+        bool isMSPFile = filename.endsWith("msp", Qt::CaseInsensitive);
+        bool isSPTXTFile = filename.endsWith("sptxt", Qt::CaseInsensitive);
+        bool notNISTFile = !(isMSPFile || isSPTXTFile);
+        bool smallerThan5Mb = fileInfo.size() < 2000000;
```
rish9511: This should be `smallerThan2Mb`

rish9511:
Have pointed out a small typo otherwise  looks fine to me mostly

src/gui/mzroll/mainwindow.cpp
saifulbkhan: My bad. Fixing it.

src/gui/mzroll/database.h
```diff
@@ -74,11 +102,12 @@ class Database {
 	deque<Molecule2D*> coordinatesDB;
 
 	map<string, Compound*> compoundIdMap;
-	map<string, Reaction*> reactionIdMap;
+        map<string, Reaction*> reactionIdMap;
 	map<string, Pathway*> pathwayIdMap;
 	map<string, Molecule2D*> coordinatesMap;
     vector<string> notFoundColumns;
     vector<string> invalidRows;
+    map<string, int> prmIdCount;
```
shubhra-agrawal: idCount is also applicable in case of MRM compound lists where different fragments of a precursor will have the same ID. Would suggest changing the name to something more generic. `idCount` works.

src/gui/mzroll/mzfileio.cpp
```diff
@@ -223,112 +223,149 @@ PK$PEAK: m/z int. rel.int.
...
+                                               .toDouble();
+        } else if (line.startsWith("EXACTMASS:", Qt::CaseInsensitive)) {
+            currentCompound->mass = line.mid(10, line.length())
+                                        .simplified()
+                                        .toDouble();
+        } else if (line.startsWith("ADDUCT:", Qt::CaseInsensitive)) {
+            currentCompound->adductString = line.mid(8, line.length())
+                                                .simplified()
+                                                .toStdString();
+        } else if (line.startsWith("FORMULA:", Qt::CaseInsensitive)) {
```
shubhra-agrawal: Check if formulae in the form of ".H2O" or "(H2O)" are handled somewhere. If not, then remove '.' and '()' from the QString as well.

src/gui/mzroll/database.cpp
```diff
@@ -84,36 +84,86 @@ void Database::loadCategories() {
...
+        return false;
+
     bool compoundAdded = false;
 
-    //new compound id .. insert into compound list
-    if (!compoundIdMap.count(c->id)) {
-        compoundIdMap[c->id] = c;
-        compoundsDB.push_back(c);
+    // lambda that checks if a compound contains PRM information
+    auto isPRM = [](const Compound* cpd) -> bool {
```
shubhra-agrawal: A compound cannot be 'PRM'. Could you replace `isPRM` with `hasSpectralInfo` or `hasFragmentationInfo` to make it more explicit? You can suggest other names as well. These are just off the top of my head

saifulbkhan: I agree. This should have been renamed to something more apt. But since then I've removed this piece of code entirely (we agreed to not check for any categories and simply append a counter after a compound's name). You may want to refresh your page.

shubhra-agrawal: I'm going commit-by-commit. Forgot about that change, sorry. I'll remove the comment

Issue #969, opened 2019-01-31
Save settings for a session in emDB


saifulbkhan:
The "emDB" format has been introduced as a way to restore user sessions as completely as possible, but at the moment only is [able to restore user data](https://github.com/ElucidataInc/ElMaven/wiki/Session-restore-in-El-MAVEN). It cannot be used to truly restore a session until it can load back the settings that were used for analysis in that session.

Some preparatory work and refactoring might be needed, before settings can be saved in emDB files.

**Here is the current plan to achieve this**
- [x] Add versioning to `projectDB` module.
- [x] Set up automatic upgrades when creating a `ProjectDatabase` instance.
- [x] Bump database version, since the DB schema will have to be changed to accommodate settings table.
- [x] Write migration script(s) for updating/downgrading a given DB instance to one that can be used in current release. We want to be completely backwards compatible with previous versions of emDB.
- [x] Create a table for storing settings from options, peak filtering and alignment dialogs of the application.
- [x] Add methods in `projectDB` module to write/read settings to/from database. Write tests for these methods as well.
- [x] Implement methods to load settings before loading other data-related information from emDB.
- [x] Possibly add UI elements to notify user about save/load operation.
- [x] Warn user when they attempt to load a project file created in a future version of El-MAVEN.

shubhra-agrawal:
These changes have been merged to develop and will be part of the v0.7.0 release.

PR #977, opened 2019-02-12
Save user settings in emDB project files

saifulbkhan:
This patch contains the following changes in context to issue #969:
- Added "_user_settings_" table to schema and increased database's `user_version` variable to 1.
- Added methods in `ProjectDatabase` module to write/read settings to/from database. 
- Implement methods in `mzFileIO` and `Controller` classes to save and load settings whenever requested through signals.
- Added UI updates when loading settings.


src/gui/mzroll/controller.cpp
```diff
@@ -125,6 +138,69 @@ void Controller::updateMavenParameters(const QString& key,  const QVariant& valu
...
+        var = v.value<QTabWidget*>()->currentIndex();
+
+    if (QString(v.typeName()).contains("QLineEdit"))
+        var = v.value<QLineEdit*>()->text().toStdString();
+
+    // specifically for massCutoffType
+    if (QString(v.typeName()).contains("QString"))
+        var = v.value<QString*>()->toStdString();
+
+    _mw->fileLoader->insertSettingForSave(k.toStdString(), var);
```
shubhra-agrawal: `insertSettingForSave` can also be done within `_syncMpWithUi`, right? Just trying to understand.

saifulbkhan: I think `_updateSettingsForSave` exists separately, even though it performs the same checks as `_syncMpWithUi`, because it is called every time the UI is explicitly changed (such as when startup settings are loaded or when `MavenParameters` is reset). `_syncMpWithUi` instead is a function that is only called to synchronize `MavenParameters` with the UI. This way they are semantically different. We can with some more modifications make the sync function perform insertions, but the call graph would get increasing ugly.

rish9511:
I tested this and its seems to be working fine 
Code also looks fine 

One thing that I want to point out is, we should not be making formatting changes, i.e correcting indentation, naming conventions etc in the same commit where a certain feature or a bug is being fixed. Such changes, imo, should be done in  a separate commit.  Having such  changes in the same commit just  makes it harder for the reviewer to follow what the commit was intended to do

I understand a lot of codebase needs to be ported to our new conventions  but not sure  if having such 
changes in the same commit is right way to go about it. What do you guys think ?

saifulbkhan:
I totally agree with this sentiment. I have a hard time with this myself. Apologies for doing this in the PR (especially as long as this one). Maybe I should hold myself back when making changes to such pieces.

For future commits, I suggest we should format the portion we wish to change as a first commit and then make the intended changes as the second commit.

rish9511:
> For future commits, I suggest we should format the portion we wish to change as a first commit and then make the intended changes as the second commit.

Ya, that's what came to my head. Let's do this from next time

saifulbkhan:
I think I had covered the settings pretty exhaustively from the two dialogs. We may add new settings for PRM data soon and alignment data when the alignment dialog becomes a persistent object. But apart from that, I don't think there are any existing variables being saved in `lastRun.xml` that I might want to change in this PR.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2953,40 +2962,44 @@ void MainWindow::createToolBars() {
...
+    quantType = new QComboBox(hBox);
+    quantType->addItem("AreaTop");
+    quantType->addItem("Area");
     quantType->addItem("Height");
-    quantType->addItem("AreaNotCorrected"); //TODO: Sahil-Kiran, Added while merging mainwindow
+    quantType->addItem("AreaNotCorrected");
     quantType->addItem("AreaTopNotCorrected");
     quantType->addItem("Retention Time");
-	quantType->addItem("Quality");
-	quantType->setToolTip("Peak Quntitation Type");
```
shubhra-agrawal: Since this part of the code is being touched, could you please correct the typo? :P

saifulbkhan: Missed this. Correcting.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -244,11 +244,11 @@ using namespace mzUtils;
 
 
 	if (QFile::exists(clsfModelFilename)) {
-		settings->setValue("clsfModelFilename", clsfModelFilename);
+                settings->setValue("peakClassifierFile", clsfModelFilename);
```
shubhra-agrawal: Minor nitpick. Indentation is off here. Feel free to ignore.

src/projectDB/projectdatabase.cpp
```diff
@@ -527,6 +539,225 @@ void ProjectDatabase::saveScans(const vector<mzSample*>& sampleSet)
...
+                      , :obi_warp_score                   \
+                      , :obi_warp_response                \
+                      , :obi_warp_bin_size                \
+                      , :obi_warp_gap_init                \
+                      , :obi_warp_gap_extend              \
+                      , :obi_warp_factor_diag             \
+                      , :obi_warp_factor_gap              \
+                      , :obi_warp_no_standard_normal      \
+                      , :obi_warp_local                   \
+                      , :main_window_compound_db_selected \
```
shubhra-agrawal: Column name in schema.h is "main_window_selected_db_name". Please make it consistent to prevent confusion in the future.

saifulbkhan: Will correct. This doesn't just cause confusion, I'd be surprised if the db name was being saved at all.

saifulbkhan:
Fixed the reported issues.

Issue #970, opened 2019-01-31
Group quality filter gives unreliable results in automated peak detection


shubhra-agrawal:
**Describe the bug**
Getting groups in Peak Table with a lower maxQuality than the quality threshold defined in Peaks dialog.

**To Reproduce**
1. Load samples
2. Go to _Peaks_ dialog -> _Group Filtering_ tab
3. Set _Minimum Quality_ to 0.2 (or any other value)
4. Click _Find Peaks_
5. Sort peak table by _Max Quality_ column (low-high)
6. See groups with _Max Quality_ less than 0.2

**Expected behavior**
Only groups with maxQuality greater than 0.2 should be filtered out.

**Screenshots**

Minimum Quality set by user
![image](https://user-images.githubusercontent.com/28925846/52049366-9b640180-2573-11e9-8ef7-40587d0f236c.png)

Resulting Peak Table
![image](https://user-images.githubusercontent.com/28925846/52049379-a323a600-2573-11e9-889b-143e2b8e8771.png)

**System (please complete the following information):**
 - OS: Ubuntu
 - Version v0.3.2 and after

**Other information**
If I set increasingly higher _Min Quality_ threshold for every Peak Table, the number of groups decrease in every iteration but there are always some groups with lower _Max Quality_ than desirable.

[Edit] Peak has a _groupOverlap_ parameter that is used to calculate the quality score. This parameter is updated AFTER quality calculation and grouping. Group filtering is done based on this quality score.
When a Peak table is created, quality scores are re-calculated with the now populated _groupOverlap_ parameter, some of these groups end up with a lower quality score.

Proposed solution: Re-calculate peak qualities after grouping so that low-quality groups can be filtered out. Peak table should not be updating group qualities, regardless.

shubhra-agrawal:
This fix has been merged to develop and will be a part of v0.7.0. Closing the thread.

PR #978, opened 2019-02-13
Fix filtering of groups based on peak quality

saifulbkhan:
El-MAVEN's group filtering methods filtered based on peak quality quantile before it was actually updated by the classifier. The classifier was being used to assign a peak quality when groups had
already been filtered and were being added into the peak table. 

This commit makes the necessary changes such that peak qualities (assigned using the classifier) and group qualities (assigned based on each peak quality within the group) are being considered when the actual group filtering process happens. Hence, no groups with all bad peaks (based on user set quantile) will end up in the peak table from automated peak detections.

Issue: #970

### DIFF
--- a/src/core/libmaven/groupFiltering.cpp
+++ b/src/core/libmaven/groupFiltering.cpp
@@ -46,7 +46,12 @@ bool GroupFiltering::filterByMS1(PeakGroup &peakgroup)
 
     peakgroup.groupStatistics();
 
-    if (_mavenParameters->clsf->hasModel() && peakgroup.goodPeakCount < _mavenParameters->minGoodGroupCount)
+    if (_mavenParameters->clsf->hasModel()) {
+        _mavenParameters->clsf->classify(&peakgroup);
+        peakgroup.updateQuality();
+    }
+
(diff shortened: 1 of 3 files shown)


src/core/libmaven/groupFiltering.cpp
```diff
@@ -40,7 +40,12 @@ bool GroupFiltering::filter(PeakGroup &peakgroup)
 
     peakgroup.groupStatistics();
 
-    if (_mavenParameters->clsf->hasModel() && peakgroup.goodPeakCount < _mavenParameters->minGoodGroupCount)
+    if (_mavenParameters->clsf->hasModel()) {
+        _mavenParameters->clsf->classify(&peakgroup);
+        peakgroup.updateQuality();
+    }
+
+    if (peakgroup.goodPeakCount < _mavenParameters->minGoodGroupCount)
```
shubhra-agrawal: If there is no model file, all groups will be filtered out. This condition should be moved to the previous `if` block

saifulbkhan: Fixed

PR #979, opened 2019-02-13
Make OBI-WARP work for PRM data

rish9511:
To make PRM data work with OBI-WARP, it's required that we filter out MS2 scans from the data. This is what this PR implements. 

Apart from that, a new type called `SampleType` has been added to `mzSample`. The type can be **MS** for MS1 data, **SRM**, **MRM**, **PRM** for MS2 data. This is mainly done to add more readability to the code. I am aware that `ms1Counts()` and  `ms2Counts()` exists and can achieve the same functionality, but types just make the code more readable imo

### DIFF
--- a/src/core/libmaven/mzAligner.cpp
+++ b/src/core/libmaven/mzAligner.cpp
@@ -520,11 +520,13 @@ bool Aligner::alignWithObiWarp(vector<mzSample*> samples,
     float binSize = obiParams->binSize;
     float minMzRange = 1e9;
     float maxMzRange = 0;
-    for (int j = 0; j < refSample->scans.size(); ++j) {
-        if(refSample->scans[j]->mslevel  == 1 ){
-            for (int k = 0; k < refSample->scans[j]->mz.size(); ++k) {
-                minMzRange = min(minMzRange, refSample->scans[j]->mz[k]);
-                maxMzRange = max(maxMzRange, refSample->scans[j]->mz[k]);
+
+    for(const auto scan: refSample->scans) {
+        // PRM/DDA data have both mslevel 1 and mslevel 2 scans. We only want to align mslevel 1 scans
+        if(scan->mslevel == 1) {
+            for(const auto mz: scan->mz) {
+                minMzRange = min(minMzRange, mz);
+                maxMzRange = max(maxMzRange, mz);
             }
         }
     }
--- a/src/core/libmaven/mzAligner.cpp
+++ b/src/core/libmaven/mzAligner.cpp
@@ -452,26 +452,31 @@ bool Aligner::alignSampleRts(mzSample* sample,
                              bool setAsReference,
                              const MavenParameters* mp)
 {
-    vector<float> rtPoints(sample->scans.size());
-    vector<vector<float> > mxn(sample->scans.size());
-    for (int j = 0; j < sample->scans.size(); ++j) {
-        if (mp->stop) return (true);
-        rtPoints[j] = sample->scans[j]->originalRt;
-        mxn[j] = vector<float> (mzPoints.size());
-    }
+    vector<float> rtPoints;
+    vector<vector<float> > mxn;
 
-    for (int j = 0 ; j < sample->scans.size(); ++j) {
-        for (int k = 0; k < sample->scans[j]->mz.size(); ++k) {
-            if (mp->stop) return (true);
-            if (sample->scans[j]->mz[k] < mzPoints.front()
-                || sample->scans[j]->mz[k] > mzPoints.back())
-                continue;
-            int index = upper_bound(mzPoints.begin(), mzPoints.end(), sample->scans[j]->mz[k])
-                            - mzPoints.begin() -1;
-
-            mxn[j][index] = max(mxn[j][index], sample->scans[j]->intensity[k]);
+    int mxnCount = 0;
+    for(auto scan: sample->scans) {
+        if(mp->stop) return (true);
+        if(scan->mslevel == 1 ) {
+            rtPoints.push_back(scan->originalRt);
+            mxn.push_back(vector<float> (mzPoints.size()));
         }
     }
+
+    for(auto scan: sample->scans) {
+        if(scan->mslevel == 1) {
+            mxnCount++;
+            for(int indice = 0; indice <  scan->mz.size(); indice++) {
+                if (mp->stop) return (true);
+                if (scan->mz[indice] < mzPoints.front() || scan->mz[indice] > mzPoints.back())
+                    continue;
+                int index = upper_bound(mzPoints.begin(), mzPoints.end(), scan->mz[indice]) - mzPoints.begin() -1;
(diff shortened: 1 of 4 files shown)


src/core/libmaven/mzAligner.cpp
```diff
@@ -452,26 +452,31 @@ bool Aligner::alignSampleRts(mzSample* sample,
...
-                || sample->scans[j]->mz[k] > mzPoints.back())
-                continue;
-            int index = upper_bound(mzPoints.begin(), mzPoints.end(), sample->scans[j]->mz[k])
-                            - mzPoints.begin() -1;
-
-            mxn[j][index] = max(mxn[j][index], sample->scans[j]->intensity[k]);
+    for(auto scan: sample->scans) {
+        if(scan->mslevel == 1) {
+            mxnCount++;
+            for(int indice = 0; indice <  scan->mz.size(); indice++) {
```
saifulbkhan: Did you mean `index` instead of `indice`? Also, we've agreed to single letter counter variables in our style guide, so we can just have this as `i`, considering `index` name is being used by a variable declared within this for-loop block.

src/core/libmaven/mzAligner.cpp
```diff
@@ -515,9 +521,11 @@ bool Aligner::alignWithObiWarp(vector<mzSample*> samples,
...
-        for (int k = 0; k < refSample->scans[j]->mz.size(); ++k) {
-            minMzRange = min(minMzRange, refSample->scans[j]->mz[k]);
-            maxMzRange = max(maxMzRange, refSample->scans[j]->mz[k]);
+        if(refSample->scans[j]->mslevel  == 1 ){
+            for (int k = 0; k < refSample->scans[j]->mz.size(); ++k) {
+                minMzRange = min(minMzRange, refSample->scans[j]->mz[k]);
+                maxMzRange = max(maxMzRange, refSample->scans[j]->mz[k]);
+            }
         }
     }
```
saifulbkhan: Although this won't result in any increased efficiency in this case, the whole for-block might be more readable this way:
```C++
    for (const auto scans : refSample->scans) {
        if(scan->mslevel > 1 )
            continue;
        for (const auto mz : scan->mz) {
            minMzRange = min(minMzRange, mz);
            maxMzRange = max(maxMzRange, mz);
        }
    }
```
It has one less level of indentation and uses range based loops. I understand this is not part of the change you were trying to introduce, so you need not implement it this way. This is just a suggestion.

src/core/libmaven/mzSample.h
```diff
@@ -217,6 +217,15 @@ class mzSample
     * @brief Load sample (supported formats: .mzxml, .mzml, .mzdata, .mzcsv and .cdf)
     * @param filename Sample file name
     */
+
+    enum SampleType: int {
```
saifulbkhan: We agreed on using `enum class` over `enum`. Let's do that here.

src/core/libmaven/mzSample.h
```diff
@@ -217,6 +217,15 @@ class mzSample
...
+        NONE = 0,
+        MS,
+        SRM,
+        MRM,
+        PRM
+    };
```
saifulbkhan: Lets change `PRM` type to `DDA` (Data Dependent Acquisition) for PRM samples, and add a `DIA` (Data Independent Acquisition) for swath samples here. Suggested by @shubhra-agrawal.

src/core/libmaven/mzSample.h
```diff
@@ -679,6 +688,7 @@ class mzSample
 
     vector<float> getIntensityDistribution(int mslevel);
 
+    SampleType sType;
```
saifulbkhan: I understand `sType` is set only once when data has been loaded from a sample file. If that is the case, then this should be a private member of `mzSample` class and should be made read-only by providing a getter method - `getSampleType()` or something like that.

src/gui/mzroll/alignmentdialog.cpp
```diff
@@ -201,17 +201,17 @@ void AlignmentDialog::algoChanged()
     refSampleLabel->setVisible(obiWarp);
 
     auto samples = _mw->getSamples();
-    auto ms2SamplesDetected = false;
+    auto srmMrmData = false;
```
saifulbkhan: This can be simply `mrmData`, since SRM is only a special case of MRM.

rish9511:
@saifulbkhan 

> I also would like to suggest that we disable the "Align" button once alignment process starts. Not only does it not make sense to have it active, clicking it actually triggers another alignment process, even though one is already in progress. Not sure if it should be a part of this PR or not. Your call


I am aware about this. Since we are short on time I am not  going to making  it part of this PR
I have created an [issue](https://github.com/ElucidataInc/ElMaven/issues/993) for this. Let's try to make this a part of next release.


Apart from this I have fixed the issues you pointed out. Please have a look

src/core/libmaven/mzSample.h
```diff
@@ -217,6 +217,16 @@ class mzSample
     * @brief Load sample (supported formats: .mzxml, .mzml, .mzdata, .mzcsv and .cdf)
     * @param filename Sample file name
     */
+
+    enum class SampleType: int {
+        NONE = 0,
+        MS,
+        SRM,
```
shubhra-agrawal: There's no need for separate `SRM` and `MRM` types imo. Just calling it MRM is fine.

src/gui/mzroll/alignmentdialog.cpp
```diff
@@ -201,17 +201,17 @@ void AlignmentDialog::algoChanged()
     refSampleLabel->setVisible(obiWarp);
 
     auto samples = _mw->getSamples();
-    auto ms2SamplesDetected = false;
+    auto mrmData = false;
     for (const auto sample : samples) {
-        if (sample->ms2ScanCount() > 0) {
-            showInfo("Obi-warp does not work with MS2 scans at the moment.\nWe "
+        if (sample->getSampleType() == mzSample::SampleType::SRM || sample->getSampleType() == mzSample::SampleType::MRM ) {
```
shubhra-agrawal: This can be made pretty simple. Just check if there are MS1 scans. If yes, alignment can be done on the MS1 scans. If not, a message pops up saying "There are no MS1 scans for alignment". No need for sample types etc.

rish9511: The types were introduced to make the code more readable.  Checking for `SRM/MRM` types makes it clear that there are certain types of sample that are not supported by OBI-WARP, just checking for `ms1Scans`, which I believe is a more technical term, will leave the readers in a more ambiguous state. Let me know if you think otherwise

rish9511:
Have pushed the required changes. Please have a look

src/gui/mzroll/alignmentdialog.cpp
shubhra-agrawal: Thanks for the change. We can discuss the error message with the product team and come up with a simpler message for the users if required!

Issue #918, opened 2018-11-14
Crash after changing PPM cutoff setting


lparsons:
**Describe the bug**
ElMaven crashes (sometimes, but not infrequently) when the PPM is changed and a peak is clicked.

**To Reproduce**
1. Open some sample data
2. Click on a peak
3. Change the `ppm` cutoff
4. Click on the peak again 
5. ElMaven crashes and offers to send a report

**Expected behavior**
ElMaven to show the updated peak data

**System (please complete the following information):**
 - Ubuntu v0.5.0, Windows v0.4.0

shubhra-agrawal:
@lparsons Thanks for reporting the steps! I can see the crash on my system. Some further observations:
- Only happens when I decrease the ppm (in my limited testing)
- Only happens when the isotope widget is open 

It's probably some corner case from the isotope widget considering the crash only happens on clicking on the peak (which is when the isoWidget is updated). IsoWidget is also a sample-centric widget so if decreasing the ppm negatively affects the selected peak, it might not know how to deal with the suddenly missing peak.

shubhra-agrawal:
@lparsons This has been fixed and merged to develop. Will be part of the v0.7.0 release.

PR #983, opened 2019-02-18
Fix crash when changing mass resolution

saifulbkhan:
Whenever isotope widget is open and peaks are selected, the widget is supposed to be updated with isotopologue information of the selected peak. However, the method that is called on peak selection
sets the current group after setting the sample (of the selected peak) which causes a slot to be called which uses current group. The slot may or may not be called before the group is set and this was causing irregular crashes. This call order has been fixed with this commit.

At the same time, the worker threads used for pulling isotopes have been corrected with updating the correct MavenParameters object, to further minimize the possibility of crashes.

Issue: #918

### DIFF
--- a/src/gui/mzroll/isotopeswidget.cpp
+++ b/src/gui/mzroll/isotopeswidget.cpp
@@ -69,15 +69,15 @@ void IsotopeWidget::clearWidget()
 
 void IsotopeWidget::peakSelected(Peak *peak, PeakGroup *group)
 {
-	clearWidget();
-	if (!(peak && group && group->compound))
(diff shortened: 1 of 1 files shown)


rish9511:
The patch works but I am not sure why

I was able to get hold of the [backtrace](https://pastebin.com/zv94aYhW) while I was on develop
I have only added the relevant threads info

The call stack in backtrace  and the place where you have fixed the bug  are completely different. That's the reason why I am unsure as to why your patch works.  


If this works for @shubhra-agrawal as well, we can merge it then.

PR #985, opened 2019-02-21
Use clang provided by https://clang.llvm.org

rish9511:
This commit makes sure that our code never uses the default clang compiler that comes with XCode

In case the user has not installed clang from brew, the compilation process will exit informing the user
about the missing compiler. 

For now, we are checking if the version of clang is 6.  This is done because  we don't want to be changing the versions of compilers often. So, this check will have to be changed whenever we change our compiler version.Also, these kind of checks will also be added for Qt/gcc versions in future

**EDIT**
For now, the solution is to make a common pri file that defines `QMAKE_CC` and `QMAKE_CXX`
and include that file in all the pro files.  I am sure qmake provides a way to define such things on a global level but at the moment I am unable to find a way to do this

### DIFF
--- a/.qmake.conf
+++ b/.qmake.conf
@@ -1,3 +1,4 @@
 top_srcdir=$$PWD
 top_builddir=$$top_srcdir/../build
 mzroll_pri=$$top_srcdir/mzroll.pri
+mac_compiler=$$top_srcdir/mac-compiler.pri
--- a/3rdparty/3rdparty.pro
+++ b/3rdparty/3rdparty.pro
@@ -1,3 +1,5 @@
+include($$mac_compiler)
+
 TEMPLATE = subdirs
 CONFIG += ordered qt thread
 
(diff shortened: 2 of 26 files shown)


saifulbkhan:
It turns out that

rish9511:
> But please before merging can you add the instructions in README as to what LDFLAGS and CPPFLAGS

Sure, will do that 



> ld: library not found for -lomp

Make sure you have `libomp.dylib`. On my system  it's present in `/usr/local/opt/llvm@6/lib/ `
And `LDFLAGS` var should be correctly set in `.bash_profile` or whatever file your terminal reads after starting up 

This is how it's set on my system:
 `LDFLAGS="-L/usr/local/opt/llvm@6/lib -Wl,-rpath,/usr/local/opt/llvm@6/lib"`

rish9511:
@saifulbkhan 

Have changed our build scripts to make use of the LDFLAGS variable for now. Can't seem to figure out other way for this.  Please check if this works for you now 

Also, i have not added the checks in `mzroll.pri` intentionally. It's included in a lot of other files where linking does not happen, and therefore will lead to a lot of unnecessary warnings if LDFLAGS is not set

PR #986, opened 2019-02-25
UI for MSMS fragmentation matching

shubhra-agrawal:
This PR implements the new UI for the upcoming MSMS feature. This includes:

- Match Fragmentation section in Peaks dialog (only interactive when MSMS data is loaded, compound search is selected, NIST library is loaded)
- Set default values for the parameters
- Store parameters in settings file
- Set "Report isotopes" checked by default (only interactive when compound search is selected)


src/core/libmaven/mavenparameters.cpp
```diff
@@ -33,14 +33,18 @@ MavenParameters::MavenParameters(string settingsPath):lastUsedSettingsPath(setti
...
+        minFragMatchScore = 0;
+        minFragMatch = 3;
+        
+        
         /*
         * Whenever we create an instance of this class, massCutoffType must be set for massCutoffMerge
         */
         massCutoffMerge = new MassCutoff();
-        massCutoffMerge->setMassCutoff(30);
+        massCutoffMerge->setMassCutoffAndType(30, "ppm");
```
saifulbkhan: This not being set was often the cause of some unexpected crashes. Good work setting it in the constructor itself.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -252,7 +252,7 @@ void  MavenParameters::setPeakDetectionSettings(const char* key, const char* val
         eicMaxGroups = atof(value);
 
     if(strcmp(key, "matchFragmentationOptions") == 0 )
-        matchFragmentation = atof(value);
+        matchFragmentationFlag = atof(value);
```
saifulbkhan: Please also add a check and assign `minFragMatchScore` for the correct key.

PR #989, opened 2019-02-27
Add option in EPI dialog to export to QuantFit

saifulbkhan:
QuantFit is a service that allows quantification of metabolites as well as that of labelled fragments of metabolites. With this addition users will directly be able to send intensity data to QuantFit through El-MAVEN. They will be provided with a link that will take them to an unploader interface where they can provide any required metadata for proceeding with calibration.

With this patch, users will also be notified with an email (which contains URL to analysis) about their upload to Polly.


src/pollyCLI/pollyintegration.cpp
```diff
@@ -326,19 +366,50 @@ QString PollyIntegration::createProjectOnPolly(QString projectname) {
...
+    QString runId = parseId(resultAndError.at(0));
+    return runId;
+}
+
+QString PollyIntegration::redirectionUiEndpoint(QString componentId,
+                                                QString runId)
+{
+    QString command = "getEndpointForRuns";
+    QStringList arguments = QStringList() << credFile;
+    QList<QByteArray> resultAndError = runQtProcess(command, arguments);
```
rish9511: Can we shift it to either of the two ? I prefer the latter one 
```
QStringList arguments(credFile);
QList<QByteArray> resultAndError = runQtProcess(command, arguments);
```
or 

```
QList<QByteArray> resultAndError = runQtProcess(command, QStringList(credFile));
```

saifulbkhan: Will do.

bin/index.js
```diff
@@ -248,6 +249,119 @@ module.exports.createWorkflowRequest = function (token_filename, project_id) {
     });
 }
 
+module.exports.createRunRequest = function (token_filename, component_id, project_id) {
+    if (has_id_token(token_filename)) {
+        public_token_header = read_id_token(token_filename);
+    }
+    var payload = {
+        "component_details":{
+            "component_name": "calibration",
```
shubhra-agrawal: @saifulbkhan How about sending the`component_name` as an argument instead of the `component_id`? The id can be fetched using the name anyway. Plus this will make this function more flexible so you could just send "intomix" as the component name when we need to integrate that app into El-MAVEN.

Just a suggestion, does not need to be done right now.

bin/index.js
```diff
@@ -248,6 +249,119 @@ module.exports.createWorkflowRequest = function (token_filename, project_id) {
...
+            "component_id": component_id
+        },
+        "project_id": project_id,
+        "name": "Polly™ QuantFit Untitled"
```
shubhra-agrawal: We should also be able to fetch the app name using a component id or similar info. Again, can be left for later.

Issue #976, opened 2019-02-08
No peak values when CSV exported from emDB loaded session.


saifulbkhan:
**Describe the bug**
CSV reports exported from a session that was reloaded from an emDB file, do not have intensity values (of selected quantitation type) of peaks from each of the samples.

**To Reproduce**
1. Curate peaks and save project as an emDB file.
2. Quit El-MAVEN
3. Start new El-MAVEN session.
3. Load emDB project file.
4. Export peak table as a CSV report.
5. CSV report has missing peak values.

**Expected behavior**
For each sample, the CSV file should contain the correct peak quantity values of the selected type.

**System (please complete the following information):**
 - OS: MacOS 10.14
 - Version: v0.6.1

shubhra-agrawal:
Will be part of v0.7.0

PR #990, opened 2019-03-04
Save sample IDs associated with groups in emDB

saifulbkhan:
Groups have a vector of mzSamples which is referred to (for intensities) when they are exported to CSV. These samples were not being stored in emDB format which led to missing intensity values in CSV files exported from emDB-restored peak tables. This behavior has been corrected.

Issue: #976

### DIFF
--- a/src/projectDB/projectversioning.cpp
+++ b/src/projectDB/projectversioning.cpp
@@ -42,6 +42,7 @@ map<int, string> dbVersionUpgradeScripts = {
         "ALTER TABLE peakgroups ADD COLUMN fragmentation_spearman_rank_corr REAL;"
         "ALTER TABLE peakgroups ADD COLUMN fragmentation_tic_matched REAL;"
         "ALTER TABLE peakgroups ADD COLUMN fragmentation_num_matches REAL;"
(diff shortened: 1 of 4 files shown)


saifulbkhan:
I do not think we need to do this, considering that this issue comes up only when using `emDB` format which was itself introduced as a feature in 0.6. Also, this patch changes schema with the upgrade scripts being marked as a 0.7 feature.

src/projectDB/projectversioning.cpp
```diff
@@ -25,7 +25,12 @@ map<Version, int> appDbVersionMap = {
  * format to the next. And DO NOT assume that any of the tables that should be
  * transformed already exist.
  */
-map<int, string> dbVersionUpgradeScripts = {};
+map<int, string> dbVersionUpgradeScripts = {
+    {
+        0, "ALTER TABLE compounds ADD COLUMN note TEXT;"
```
shubhra-agrawal: Why add `note` column in this branch? Is this branch supposed to be based on some other one?

saifulbkhan: Sorry about this. I was running into an issue with project-versioning when I was working with this and PR #980. Must have staged this change during that mess. Will correct.

Issue #962, opened 2019-01-17
Interchanged values saved in emDB peakgroups table


saifulbkhan:
**Describe the bug**
emDB file is saving `expectedRtDiff` attribute in place of `expectedAbundance` and vice-versa. This bug was caught while writing tests for the `projectDB` module.

The major issue that this causes is that `expectedRtDiff` values exported in a CSV file from an emDB loaded table will have its `expectedAbundance` values in that column instead of the correct values.

Thankfully, isotope plot widget calculates abundance values everytime before displaying so this bug should not effect the plot.

**Expected behavior**
The obvious expected behavior is to save group properties in their correct column.

shubhra-agrawal:
Done. Will be part of v0.7.0

PR #991, opened 2019-03-04
Fix expected RT diff and abundance swapped in emDB

saifulbkhan:
Because of the positioning of expected RT difference and expected abundance variables in insertion statement, the wrong values were being bound and saved in (and read from) SQLite databases. This
has been fixed.

Issue: #962


rish9511:
Same as #990 ,should we patch this up for 0.6 series ?
Asking because CSV exports were working fine before and we broke it in 0.6.1

saifulbkhan:
CSV exports themselves are not broken, only when the user exports one from an `emDB` restored session. And again this is a patch dependent on the existence of project versioning ability (marked as a 0.7 feature) and I would like to avoid having a different schema for the same minor version (0.6) being distributed in patch releases.

PR #992, opened 2019-03-05
Group classification user survey

rish9511:
The following things have been implemented in this PR

- New ML model is now being used to mark Peak groups as good and bad. The peaks get uploaded to our cloud platform, which marks the  Peak groups as good and bad(Done by @Ky5595 )

- The information, whether the Peak group is marked as good and bad, is being in stored in `markedGoodByCloudModel` and `markedBadByCloudMode` variables. This along with the name of isotope and compound is being uploaded to PollyPhi in a separate csv file

- Video demonstrating the PollyPhi workflow has been added in the codebase. A custom video player  has been added for now to display the video. In future we might shift to `QWebEngine`

P.S The size of the binary after including both videos, in release mode, is 72 MB. Without the videos it's 7-10 MB


3rdparty/libsvm/tools/README
```diff
@@ -0,0 +1,210 @@
+This directory includes some useful codes:
```
saifulbkhan: @Ky5595 Are we using these Python scripts? If not, then we should not include them in our repo (the license allows modifications to the original source code in any form).

bin/group.weights
```diff
@@ -0,0 +1,9 @@
+nnlib network weights data file version 1.0
```
shubhra-agrawal: @Ky5595 Too much clutter in the bin folder. Could you group some of your files and use a sub-folder?

src/gui/mzroll/videoplayer.cpp
```diff
@@ -0,0 +1,162 @@
...
+    footerLayout->addWidget(m_closeButton);
+    footerLayout->itemAt(1)->setAlignment(Qt::AlignRight);
+
+    layout->addLayout(footerLayout);
+
+
+    connect(m_playButton, &QAbstractButton::clicked,
+            this, &VideoPlayer::play);
+    connect(m_positionSlider, &QAbstractSlider::sliderMoved,
+            this, static_cast<void (VideoPlayer::*)(int)>(&VideoPlayer::setPosition));
```
saifulbkhan: This is a peculiar cast. I have never seen it. I could not find such an example just by searching around. What is it doing? (just curious :)

src/gui/mzroll/videoplayer.cpp
```diff
@@ -0,0 +1,162 @@
...
+    m_title->setContentsMargins(0,0,0,10);
+    m_title->setAlignment(Qt::AlignCenter);
+    layout->addWidget(m_title);
+
+    m_message = new QLabel(this);
+    m_message->setWordWrap(true);
+    m_message->setAlignment(Qt::AlignCenter);
+    m_message->setContentsMargins(0,0,0,20);
+    m_message->setText("Classify groups as Good and Bad quickly and with accuracy with the new Machine Learning \
+algorithm. Save time classifying data and spend more time analyzing it. ");
```
saifulbkhan: Maybe we should not be setting such message strings in the constructor itself. We might need this player later for other demonstrations as well. But it need not be changed right now.

src/gui/mzroll/videoplayer.h
```diff
@@ -0,0 +1,53 @@
...
+    QMediaPlayer* m_mediaPlayer;
+    QAbstractButton *m_playButton;
+    QWidget* m_hboxWidget;
+    QWidget* m_vboxWidget;
+    QCheckBox* m_checkBox;
+    QPushButton* m_closeButton;
+    QVideoWidget* m_vidWidget;
+    QSlider *m_positionSlider;
+    QLabel *m_errorLabel;
+    QSettings* m_settings;
```
saifulbkhan: I am not sure I have seen this way of declaring private members (with the `m_` prefix). We have agreed to having private member (methods, slots and variable) names starting with underscores only.

src/gui/mzroll/videoplayer.cpp
rish9511: Ya, this should be used as a generic video player. Will change this if we add another video

src/gui/mzroll/videoplayer.h
rish9511: My bad, will change this

src/gui/mzroll/videoplayer.cpp
rish9511: In case of overloaded functions we need to let the object of that class know which overloaded function to call. I am also unsure what exactly the syntax means but broadly speaking , in this case it's placing the address of the function which takes integer as one the arguments . More [here](https://wiki.qt.io/New_Signal_Slot_Syntax#Overload)

src/pollyCLI/pollyintegration.cpp
```diff
@@ -4,7 +4,8 @@ PollyIntegration::PollyIntegration(): nodePath(""), jsPath("")
 {
     credFile = QStandardPaths::writableLocation(QStandardPaths::QStandardPaths::GenericConfigLocation) + QDir::separator() + "cred_file";
 
-    // nodePath = QStandardPaths::findExecutable("node");
+    // It's important to look for node in the system first, as it might not always be present in the bin dir.
+     nodePath = QStandardPaths::findExecutable("node");
```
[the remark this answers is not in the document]
rish9511: > is this not the same as one of the checks that are performed below within the platform-specific #ifdef macros?

No. If we provide `QStandardPaths::findExecutable` with our custom paths, then it only looks for the executable in those custom paths and not in system paths.  

Our repo has `node` binaries in `bin` folder which I feel is very unnecessary.  Without this check, EPI will not work if node is not found in the local repo.  I just wanted to make sure that in such case we should make use of  `node` present in the system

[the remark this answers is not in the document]
saifulbkhan: Cool got it.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -422,10 +435,21 @@ void PeakGroup::reduce() { // make sure there is only one peak per sample
 void PeakGroup::updateQuality() {
     maxQuality=0;
     goodPeakCount=0;
-    for(unsigned int i=0; i< peaks.size(); i++) {
-        if(peaks[i].quality > maxQuality) maxQuality = peaks[i].quality;
-        if(peaks[i].quality > minQuality) goodPeakCount++; //Sabu
+    float highestIntensity=0;
```
shubhra-agrawal: `highestIntensity` is not being used anywhere. We can remove this.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -679,7 +746,9 @@ void TableDockWidget::prepareDataForPolly(QString writableTempDir,
 }
 
 void TableDockWidget::exportJsonToPolly(QString writableTempDir,
-                                        QString jsonfileName) {
+                                        QString jsonfileName,
+                                        bool uploadToPolly)
```
shubhra-agrawal: `uploadToPolly` is misleading, no? this should be `uploadtoPollyPhi` if we're only sending for fluxomics workflow, or `true` every time because it's `exportJsonToPolly`. @rish9511

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -323,6 +360,26 @@ void TableDockWidget::addRow(PeakGroup *group, QTreeWidgetItem *root) {
       item->setText(15, QString::number(group->changeFoldRatio, 'f', 2));
       item->setText(16, QString::number(group->changePValue, 'e', 4));
     }
+    item->setText(15,QString::number(group->avgPeakQuality,'f',2));
```
shubhra-agrawal: @Ky5595 Column number 15 and 16 are being overwritten here. Please resolve this.

src/gui/mzroll/mzroll.pro
```diff
@@ -6,6 +6,8 @@ include($$mzroll_pri)
 DESTDIR = $$top_srcdir/bin/
 #TEMPLATE = app
 
+QT += multimedia multimediawidgets
```
shubhra-agrawal: @rish9511 Please update the README for Ubuntu installation.

saifulbkhan:
@rish9511 FYI: the video dialog instead of being a dialog box, opens in a new tab in the El-MAVEN window on MacOS (10.14). Pressing `Cmd + Q` causes the original El-MAVEN window to hide and then pressing the same combo again makes it appear but in the switched tab.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -695,12 +764,117 @@ void TableDockWidget::exportJsonToPolly(QString writableTempDir,
...
+  //   }
+  //   if (grp.label =='g' & grp.markedGoodByCloudModel == 1){
+  //     tp+=1;
+  //   }
+  //   if (grp.label =='b' & grp.markedBadByCloudModel == 1){
+  //     tn+=1;
+  //   }
+  // }
+
+  for (int i = 0; i < allgroups.size(); i++) {
```
shubhra-agrawal: We have talked about this offline but putting it here so it's trackable.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -980,17 +1171,17 @@ void TableDockWidget::showConsensusSpectra() {
   }
 }
 
-void TableDockWidget::markGroupGood()
-{
+void TableDockWidget::markGroupGood() {
+  qDebug()<<"inside markGroupGood()";
```
shubhra-agrawal: Please remove the debug statements

src/gui/mzroll/tabledockwidget.cpp
rish9511: > this should be uploadtoPollyPhi if we're only sending for fluxomics workflow or true every time because it's exportJsonToPolly.

I think you meant  `pollyPhi` workflow. It can't be true every time since we also upload data to fluxomics. 

> uploadToPolly is misleading, no? 

Yes, to some extent. If we are uploading to pollyPhi  i.e if `uploadToPolly` is `true` then we add some extra content(data labelled by ML model) to json files. Maybe we can rename it to `addExtraData` ?

shubhra-agrawal: Sorry if I wasn't clear. I meant that if we are only sending the extra data for PollyPhi workflow, the variable can be renamed to `uploadtoPollyPhi` and set true when PollyPhi is selected. Or we can keep it as `uploadToPolly` and add the extra data in all Polly JSONs.

I'm in favour of changing the variable name to either `uploadToPollyPhi` or `addMLInfo`. What do you think?

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -853,11 +963,28 @@ void TableDockWidget::setGroupLabel(char label) {
       QVariant v = item->data(0, Qt::UserRole);
       PeakGroup *group = v.value<PeakGroup *>();
       if (group != NULL) {
+        if (!(group->label=='g'||group->label=='b')){
+          numberOfGroupsMarked+=1;
+          if (numberOfGroupsMarked == 1) curationTimer.start();
```
shubhra-agrawal: the timer was being used to calculate the time saved by using the ML model. This was presented on the toast you just removed. We can safely remove this timer now.

shubhra-agrawal: Along with the tenPeaksMarked signal.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -853,11 +963,28 @@ void TableDockWidget::setGroupLabel(char label) {
...
+          group->setLabel(label);
+          subsetPeakGroups.push_back(*group);
+        }
         group->setLabel(label);
+        if (numberOfGroupsMarked ==10){
+          if (_mainwindow->sessionCount==1){
+            // This code is used to show the initial inferences from the cloud model.
+            Q_EMIT(tenPeaksMarked(curationTimer.elapsed()));
+          }
+          _mainwindow->sessionCount+=1;
```
shubhra-agrawal: sessionCount is not required either. It was used to ensure the "accuracy and time of ML model" toast is only shown once per session.

src/gui/mzroll/tabledockwidget.cpp
rish9511: >  Or we can keep it as uploadToPolly and add the extra data in all Polly JSONs

They don't want this data in FirstView

> I'm in favour of changing the variable name to either uploadToPollyPhi or addMLInfo. What do you think?

addMLInfo sounds better

src/pollyCLI/pollyintegration.cpp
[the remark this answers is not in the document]
shubhra-agrawal: @rish9511 What if the system node is an older/otherwise incompatible version? Shouldn't we look in the bin first and then the system?

src/gui/mzroll/tabledockwidget.cpp
shubhra-agrawal: @Ky5595 Please take care of this.

PR #996, opened 2019-03-14
Add UI for MS2 events

shubhra-agrawal:
- Adds markers on the EIC x-axis to mark the RTs where a fragmentation event happened. 
- Populates the MS2 Events list (or fragPanel) with all the MS2 scans corresponding to a mass and time range
- Calculates purity of an MS2 event for display in the MS2 Events list
- Adds the isolation window variable to be used in fragmentation widget later


src/gui/mzroll/eicwidget.cpp
```diff
@@ -1994,35 +1984,39 @@ void EicWidget::timerEvent(QTimerEvent * event) {
...
-        		QColor color = QColor::fromRgbF( sample->color[0], sample->color[1], sample->color[2], 1 );
-                EicPoint* p  = new EicPoint(toX(scan->rt), toY(10), NULL, getMainWindow());
+        		QColor color = QColor::fromRgbF(sample->color[0],
+												sample->color[1],
+												sample->color[2],
+												1);
+                EicPoint* p  = new EicPoint(toX(scan->rt),
+											toY(10),
+											NULL,
+											getMainWindow());
```
rish9511: The indentation here and [above](https://github.com/ElucidataInc/ElMaven/pull/996/commits/251f6f3d27efad5c7e9226a8e8663e3c02db3b5f#diff-999b8a9c0e6dc83ba91ee104baffd2c2R2012) looks a bit weird

src/gui/mzroll/eicwidget.cpp
```diff
@@ -907,31 +905,23 @@ void EicWidget::replot(PeakGroup* group) {
...
-    if(_showMergedEIC) addMergedEIC();
-    if(_focusLineRt >0) setFocusLine(_focusLineRt);  //qDebug() << "\tsetFocusLine msec=" << timerX.elapsed();
-    if(_showNotes)	 getNotes(eicParameters->_slice.mzmin,eicParameters->_slice.mzmax);	//get notes that fall withing this mzrange
-    if(_showMS2Events && eicParameters->_slice.mz>0) { addMS2Events(eicParameters->_slice.mzmin, eicParameters->_slice.mzmax); }
+    if (_showCubicSpline) addCubicSpline();
+    if (_showBaseline) addBaseLine();
+	if (_showTicLine || _showBicLine) addTicLine();
+    if (_showMergedEIC) addMergedEIC();
+    if (_focusLineRt >0) setFocusLine(_focusLineRt);
+    if (_showNotes)	getNotes(eicParameters->_slice.mzmin,eicParameters->_slice.mzmax);	//get notes that fall withing this mzrange
```
rish9511: The function calls should be done on the next line, no?  Our style guide does not mention whether the calls should be made in the next line or the same one but I prefer defining anything after the conditional statement on the next line,

e.g.
```
    ...

    if(_showCubicSpline)
        addCubicSpline();

    ...
```

src/gui/mzroll/eicwidget.cpp
```diff
@@ -1994,35 +1984,39 @@ void EicWidget::timerEvent(QTimerEvent * event) {
...
-        for (unsigned int j=0; j < sample->scans.size(); j++ ) {
-            Scan* scan = sample->scans[j];
 
-            if (scan->mslevel > 1 && scan->precursorMz >= mzmin && scan->precursorMz <= mzmax) {
+    //clear MS2 events list
+	mw->fragPanel->clearTree();
+    
+	int count = 0;
+    for (auto sample : samples) {
+        for (auto scan : sample->scans) {
```
rish9511: We need to use `auto&` to prevent copying .
Forgot to mention this in the previous PRs

src/gui/mzroll/eicwidget.cpp
shubhra-agrawal: EicWidget is tab-indented so it looks weird in different editors. I can make them space-indented in the ui_fragSpectra branch though. To prevent merge conflicts with further branches.

src/gui/mzroll/eicwidget.cpp
shubhra-agrawal: Agreed. I'll push the changes in this PR.
Also, please add this to the style guide if @saifulbkhan agrees?

src/gui/mzroll/eicwidget.cpp
shubhra-agrawal: auto const & seems more suitable then. We're not modifying the elements.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3514,17 +3517,21 @@ void MainWindow::showFragmentationScans(float pmz) {
...
 	fragPanel->clearTree();
-	for (unsigned int i = 0; i < samples.size(); i++) {
-		for (unsigned int j = 0; j < samples[i]->scans.size(); j++) {
-			if (samples[i]->scans[j]->mslevel > 1
-					&& massCutoffDist(samples[i]->scans[j]->precursorMz, pmz,massCutoff) < massCutoff->getMassCutoff()) {
-				fragPanel->addScanItem(samples[i]->scans[j]);
+	for (auto sample : samples) {
+		for (auto scan : sample->scans) {
+			if (scan->mslevel < 2) continue;
+			if (massCutoffDist(scan->precursorMz,
```
rish9511: This is indented with tabs. Please switch to spaces

src/gui/mzroll/eicwidget.cpp
rish9511: `const auto&`

shubhra-agrawal: Is that a convention? both `auto const &` and `const auto&` work the same, right?

src/gui/mzroll/mainwindow.cpp
shubhra-agrawal: Done.

src/gui/mzroll/eicwidget.cpp
saifulbkhan: I agree. I will add it to the style guide.

src/gui/mzroll/eicwidget.cpp
saifulbkhan: `const auto&` is pretty much the convention. I use `auto&` almost always (unless its a pointer variable, like in this case), but `const` is not always possible (not sure if it is in this case or now).

src/core/libmaven/mzSample.cpp
```diff
@@ -495,6 +495,11 @@ void mzSample::parseMzMLSpectrumList(const xml_node &spectrumList)
 		if (string2float(precursorMzStr) > 0)
 			precursorMz = string2float(precursorMzStr);
 
+		string precursorIsolationStr = isolationWindow["isolation window lower offset"];
+		float precursorIsolationWindow = 1.0;
```
saifulbkhan: Just like earlier, let us start declaring float with the proper suffix (i.e., `float precursorIsolationWindow = 1.0f;`).

src/core/libmaven/Scan.cpp
```diff
@@ -539,7 +537,72 @@ vector<int> Scan::assignCharges(MassCutoff *massCutoffTolr) {
...
+	//calculate total intensity in isolated segment
+	double totalInt = 0;
+	for (mzPoint& point: isolatedSegment)
+        totalInt += point.y;
+
+	if (totalInt > 0) {
+		return (targetInt / totalInt);
+	} else {
+		return 0;
+	}
```
saifulbkhan: We should stick to using spaces for indentation. This new method body uses tabs and spaces mixed.

src/core/libmaven/Scan.cpp
```diff
@@ -539,7 +537,72 @@ vector<int> Scan::assignCharges(MassCutoff *massCutoffTolr) {
...
+	//extract isolated window
+    vector<mzPoint> isolatedSegment = this->getIsolatedRegion(this->isolationWindow);
+	if (isolatedSegment.size() == 0) return 0;
+
+	//get last full scan
+	Scan* lastFullScan = this->getLastFullScan();
+	if (!lastFullScan) return 0;
+
+	//locate intensity of isolated mass
+    MassCutoff* massCutoff = new MassCutoff();
```
saifulbkhan: When done with `massCutoff` lets deallocate it? Although it probably does not take up a lot of space, this method might be called multiple number of times, allocating a new `MasCutoff` everytime. We can do small things like this to reduce (minimal as it may be) the amount of memory we consume during runtime.

Also just as a reminder, all of us collaborators should keep an eye out for leaks like this in each other's PRs, to avoid making the situation worse from newly added code. Pretty sure I am guilty in this regard as well.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3513,30 +3513,29 @@ void MainWindow::setClipboardToGroup(PeakGroup* group) {
...
-	}
+    if (samples.size() <= 0)
+        return;
+    fragPanel->clearTree();
+    for (auto sample : samples) {
+        for (auto scan : sample->scans) {
+	        if (scan->mslevel < 2) continue;
+	        if (massCutoffDist(scan->precursorMz,
+                               pmz,
+                               massCutoff) > massCutoff->getMassCutoff()) {
```
saifulbkhan: Again, `massCutoff` will probably not be needed after this point. Lets free the allocated memory.

shubhra-agrawal: This one is a local pointer to a global variable, new memory has not been allocated.

src/core/libmaven/Scan.cpp
shubhra-agrawal: Thanks for pointing this out. Pushing the change!

src/core/libmaven/Scan.cpp
shubhra-agrawal: Done.

PR #998, opened 2019-03-14
UI/UX changes for spectra and fragSpectra widgets

shubhra-agrawal:
- Calculate and display average precursor purity and mean group Rt in fragmentationSpectra widget
- Open fragmentationSpectra widget on clicking a group in PeakTable (only if it has fragmentation info)
- Update both spectra widgets on clicking a peak top
- Open spectra widget on clicking an item in MS2 Events list
- Update spectra widget on clicking MS2 markers


shubhra-agrawal:
@rish9511 
< The spectra widget does not change on selecting a new group from a Peak table, however the Fragmentation spectra widget changes. Is this intentional ?
Introduced in this PR

Yes, it is. The spectra widget displays the raw data in a scan as opposed to the Fragmentation spectra widget which displays the group fragmentation. Updating the group does not update the spectra widget at the moment. Maybe when we add a sample drop-down to the spectra widget, we can have it update that as well.

< Just after uploading samples(tested with PRM samples) there are some extra lines on the EIC widget.
Zooming in and out makes them go. This was there in #997 as well.

As we discussed offline, those dashed lines are the baseline. The EIC is very zoomed in so the high intensity peak corresponding to this high baseline is not visible on the screen.

src/gui/mzroll/spectrawidget.cpp
```diff
@@ -394,33 +396,79 @@ void SpectraWidget::setTitle()
...
-        _titleText += tr("CE:<b>%1</b>  ").arg(QString::number(_currentScan->collisionEnergy,'f',0));
+        _titleText += tr("<b>CE:</b> %1  ").arg(QString::number(_currentScan->collisionEnergy,'f',0));
     }
 
     if (_currentScan->productMz) {    
-       _titleText += tr("Prod m/z:<b>%1</b>  ").arg(QString::number(_currentScan->productMz,'f',3));
+       _titleText += tr("<b>Prod m/z:</b> %1  ").arg(QString::number(_currentScan->productMz,'f',3));
+    }
+
+    if (_currentScan->precursorMz > 0 && _currentScan->sample) {
```
rish9511: Why is the second check required ?

src/core/libmaven/Fragment.cpp
```diff
@@ -201,6 +205,34 @@ void Fragment::buildConsensus(float productPpmTolr)
     this->consensus = consensusFrag; 
 }
 
+float Fragment::consensusRt()
+{
+    if (brothers.size() == 0) 
+		return this->rt;
```
rish9511: `this` is not required when using inside member functions. Code will start to look messy if you want to use a lot member variables.  Although, we do not have this in our guidelines, not sure if we should have it, but I generally avoid using `this` pointer, unless some sort of  `type casting` is required

shubhra-agrawal: Yes, missed that. Lots of `this` usage in MAVEN. Will remove that.

src/gui/mzroll/spectrawidget.cpp
shubhra-agrawal: purity is not a pre-calculated value. It's calculated from the sample. If `_currentScan` does not have an associated sample (this can happen in case of fragmentation widget), purity cannot be calculated so it's checking for that possibility.

shubhra-agrawal: Can remove the sample check. PrecursorMz check is there to determine if it's MS2 or not but the sample one isn't really necessary.

src/gui/mzroll/spectrawidget.cpp
```diff
@@ -615,30 +617,35 @@ void SpectraWidget::drawAnnotations() {
...
-    if ( fabs(_spectralHit.precursorMz - _currentScan->precursorMz) < 0.5 ) {
-        drawSpectralHit(_spectralHit);
-        if (_currentGroup) setGroupTitle();
-    } else {
-        //TODO: either remove the check or inform user on the UI in case of failure
-       qDebug() << " overlaySpectra() skipped: " << _spectralHit.precursorMz << " " << _currentScan->precursorMz;
-    }
+    if (_spectralHit.mzList.size() > 0) {
+        setGroupTitle();
+        if (fabs(_spectralHit.precursorMz - _currentScan->precursorMz) < 0.5)
```
saifulbkhan: Comparing a float against a double

src/core/libmaven/mzSample.cpp
```diff
@@ -963,10 +968,24 @@ void mzSample::parseMzXMLScan(const xml_node &scan, const int& scannum)
 		scanpolarity = getPolarityFromfilterLine(filterLine);
 	}
 
-	//TODO: What is this for this is zero
-	precursorMz = string2float(string(scan.child_value("precursorMz")));
-
-	// string b64String; naman Unused variable: b64String
+    xml_node precursor = mzxml_scan_node.child("precursorMz");
+    if (precursor) {
+        _scan->precursorMz = string2float(mzxml_scan_node.child_value("precursorMz"));
```
saifulbkhan: The variable `_scan` is also not known in this scope.

src/core/libmaven/mzSample.cpp
```diff
@@ -963,10 +968,24 @@ void mzSample::parseMzXMLScan(const xml_node &scan, const int& scannum)
 		scanpolarity = getPolarityFromfilterLine(filterLine);
 	}
 
-	//TODO: What is this for this is zero
-	precursorMz = string2float(string(scan.child_value("precursorMz")));
-
-	// string b64String; naman Unused variable: b64String
+    xml_node precursor = mzxml_scan_node.child("precursorMz");
```
saifulbkhan: The variable `mzxml_scan_node` is not known in this scope. This is one of the issues that fail to compile.

src/gui/mzroll/spectrawidget.cpp
shubhra-agrawal: I'll use the almostEqual method from mzUtils for this. Thanks for pointing out!

src/core/libmaven/mzSample.cpp
shubhra-agrawal: Fixing it. Lost my changes once and reproduced it from memory. Forgot a few crucial things. My bad.

src/core/libmaven/mzSample.cpp
shubhra-agrawal: Same as above. Pushing the changes.

shubhra-agrawal:
> Also, correct me if I am wrong but is this not the branch where deleting a peak table was crashing when one of the spectra widgets were open? If so I do not see any particular commits relating to that change.

That is also happening on develop, right? I'll send it as a separate fix. Opening an issue for it now.

PR #999, opened 2019-03-18
Build tests conditionally

rish9511:
Tests are not required be to deployed on user's system. Our build
scripts at the moment have no way of checking if it's actually
required to build tests or not.

NOTESTS variable will be used to check if tests are to be built or
not

### DIFF
--- a/build.pro
+++ b/build.pro
@@ -24,3 +24,5 @@ include($$mac_compiler)
 TEMPLATE = subdirs
 CONFIG += ordered qt thread
 SUBDIRS += 3rdparty crashhandler src tests/MavenTests 
+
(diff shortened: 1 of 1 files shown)


rish9511:
> Or is this something left for the elmaven-packaging repo to take care of.

packaging-repo does not use `run.sh`.  The command used in the packaging repo for building is `qmake CONFIG+=release CONFIG+=force_debug_info`,  after this PR, it will be changed to `qmake NOTESTS=yes ...`

Is there a reason why you would not want to build tests in release mode ?   
As to what I understand, you want to use this functionality while using `run.sh`.  No good solution in my mind atm apart from using the command line directly

saifulbkhan:
Right. I had forgotten that packaging script did not use `run.sh`. I have no problems with this change then.

PR #1001, opened 2019-03-19
Check dependencies

rish9511:
[pkg-confg](https://www.freedesktop.org/wiki/Software/pkg-config/) is a helper tool used when compiling applications and libraries. We will be using it to check if all the dependencies are installed before compilation process actually starts. 
Currently, the check is just for `libcurl`,  it also  has been implemented in a very immature way(no custom error messages if a particular dependency is not found). In future, we can add checks for other libs as well etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,20 +36,32 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
 ### Windows
 
 - Download [MSYS2](http://www.msys2.org/) installer and follow the installation instructions provided on their website.
-- Open MSYS2 and give the following commands to set up libraries and tool chains for El-MAVEN. Reopen MSYS2 when required:  
-- **For 64 bit**:  
-`pacman --force -Sy`  
-`pacman --force -Syu`  
-`pacman --force -Su`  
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
 `pacman --force -Sy`  
 `pacman --force -Syu`  
 `pacman --force -Su`  
-`pacman --force -Sy base-devel msys2-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-qt5 mingw64/mingw-w64-x86_64-hdf5 mingw64/mingw-w64-x86_64-netcdf mingw64/mingw-w64-x86_64-boost msys/git`  
+`pacman --force -Sy base-devel msys2-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-qt5 mingw64/mingw-w64-x86_64-hdf5 mingw64/mingw-w64-x86_64-netcdf mingw64/mingw-w64-x86_64-boost msys/git mingw-w64-x86_64-curl`
 
 - **For 32 bit**:
```
saifulbkhan: Let's remove this section altogether, since we do not support 32-bit systems anymore.

rish9511:
@saifulbkhan 
Have modified the build instructions for Windows platform. 
Please have a look again

PR #1004, opened 2019-03-25
Crash on peak Table deletion

shubhra-agrawal:
Spectra widget used to store a pointer to the currently selected peakGroup from the peakTable.
On deleting the peakTable, the peakGroup is also deleted and therefore accessing it through the pointer make the app crash.

The peakGroup object will now be copied and stored as the _currentGroup to prevent such crashes.

### DIFF
--- a/src/gui/mzroll/spectrawidget.cpp
+++ b/src/gui/mzroll/spectrawidget.cpp
@@ -5,7 +5,7 @@ SpectraWidget::SpectraWidget(MainWindow* mw) {
     eicparameters = new EICLogic();
    _currentScan = nullptr;
    _avgScan = nullptr;
-   _currentGroup = nullptr;
+   _currentGroup = PeakGroup();
    _spectralHit = SpectralHit();
 
     initPlot();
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/spectrawidget.cpp
```diff
@@ -218,13 +218,15 @@ void SpectraWidget::overlayPeptideFragmentation(QString peptideSeq,MassCutoff *p
 
 void SpectraWidget::overlayPeakGroup(PeakGroup* group)
 {
-    _currentGroup = nullptr;
+    delete _currentGroup;
```
saifulbkhan: `_currentGroup` is a `PeakGroup` on stack now. We cannot `delete` it. This statement causes compilation failure.

shubhra-agrawal: A proper example for why last-minute changes are dangerous :)

I'll remove this.

shubhra-agrawal:
> There is one issue that does not allow me to compile this branch as it is. After and fixing this and building, the crash does not appear anymore. The solution seems cleaner to me.

Should probably do the same in isotopeWidget in the next release.

> One final note - after deleting a peak table, the spectra widgets and events list do not seem to be cleared, however, I do not know whether they should be. Just pointing this out.

I don't see why they 'should' be cleared but I can't make a case for the opposite either. I'll change the UX if the product team is not happy with it.

PR #1010, opened 2019-03-28
Fix binary name

rish9511:


src/gui/mzroll/mzroll.pro
```diff
@@ -26,7 +26,7 @@ DEFINES += "EL_MAVEN_VERSION=$$VERSION"
 DEFINES += "PLATFORM=\"$$QMAKE_HOST.os\""
 DEFINES += EMBEDHTTPSERVER
 
-TARGET = "El-MAVEN $$VERSION"
+TARGET = "El-MAVEN"
```
shubhra-agrawal: Correct me if I'm wrong but `$$VERSION` is the full version name with the commit ID, right? Why are we removing that from the binary?

rish9511: VERSION = v0.6.1-commitID. 
It's redundant. We have it on title bar of the Main window
Most importantly, I have been unable to compile it on my system because of this naming convention. 
Msys shell doesn't like spaces in the binary  name, atleast on my system. 

This will also prevent creation of multiple binaries in the bin folder. If we ever want to use a specific version, we can copy it under a different name 

Does this make sense ?

shubhra-agrawal: Got it. Instead of removing this whole thing, can we add a hyphen or underscore in between so that it does not complain about spaces? I find it helpful to have the version name in the binary name itself. Plus I don't see why having multiple binaries in the bin folder is necessarily a problem.
What do you think?

rish9511: The aim is to keep the name of binary in a standard format. No application, as far as I have seen,  keeps the version along with the binary name. Maybe they just do it while deploying the application, which is what we are going to do if we decide to keep version along with the binary name.   I hope you don't want to keep the version string on user's system 

@saifulbkhan  Thoughts ?

shubhra-agrawal: Valid point. Let's remove it then.

saifulbkhan: I agree with @rish9511. We can have a different binary name (maybe without any spaces and more verbose information if you want) for debug build. The release binary should not have any such information, in my opinion (just like the `Chrome_66` or `Slack_3.3.8` is not what we see in title bar for such consumer applications).

shubhra-agrawal: My main concern is that the version name should be there somewhere. Either in the binary name or the title.  For tagged versions, the version name can stay in the title as it is pretty clean anyway and also helps when users report issues but forget to provide the version information. Not the best reason, I know. But it has helped us a lot in the past. 
For developer builds, it would be nice to have the version information in the binary name as well so a quick look can tell me if I have the correct binary for the branch I want to test/build. but I won't be terribly inconvenienced if it isn't.

rish9511: > But it has helped us a lot in the past. As long as we don't have a proper issue reporting system in the software itself, I feel like we shouldn't remove it from the title.

Title bar , right ? We are not removing it from the there 

> For dev builds

We can do something about this. Dev builds can continue having the version string like it's now.
Is this urgent ?

saifulbkhan: Maybe we should have an "About" dialog, for version information. Almost all applications I've ever used have them.

rish9511: Ya, we decided this some time back
We can do this

shubhra-agrawal: > We can do something about this. Dev builds can continue having the version string like it's now.
Is this urgent ?

No it isn't. Fine for now. 

> Maybe we should have an "About" dialog, for version information. Almost all applications I've ever used have them.

An About dialog is definitely a requirement. In any case, since the title is not being changed in this PR, that's a conversation for later.

PR #1015, opened 2019-04-02
 Make Crash reporter an active application

rish9511:
Windows does not allow an application to interrupt what the user is currently doing in another application. As a result, Crash reporter used to stay in task bar instead of getting displayed.
To fix this, it's important to make the application active before showing it


shubhra-agrawal:
Tested on Windows!

PR #1020, opened 2019-04-03
[Fix] Make CSV exports compatible with Polly apps

shubhra-agrawal:
Polly apps accept a specific format from El-MAVEN. The additional columns for MSMS info in CSV files can cause major errors on Polly apps. These columns will therefore be dropped from the CSV export before sending to Polly.
The local CSV exports will continue to behave as earlier.


shubhra-agrawal:
@saifulbkhan Have tested this using PollyPhi on production Polly as the test environment wasn't up.

PR #1021, opened 2019-04-10
Calculate MS2 score and num of Events for bookmarked groups

shubhra-agrawal:
The 'MS2 Score" and "#MS2 Events" columns were zero for all bookmarked groups (double-click bookmark and shift-drag integrated). These scores are now being calculated during peak grouping and updated in the table.

### DIFF
--- a/src/core/libmaven/eiclogic.h
+++ b/src/core/libmaven/eiclogic.h
@@ -34,14 +34,17 @@ public:
 								int intensityWeight,
 								int deltaRTWeight);
 
-	void groupPeaks(float eic_smoothingWindow,
-					float grouping_maxRtWindow,
-					double minQuality,
-					double distXWeight,
-					double distYWeight,
-					double overlapWeight,
-					bool useOverlap,
-					double minSignalBaselineDifference);
+    void groupPeaks(float eic_smoothingWindow,
+                    Compound* compound,
+                    float grouping_maxRtWindow,
+                    double minQuality,
+                    double distXWeight,
+                    double distYWeight,
+                    double overlapWeight,
+                    bool useOverlap,
+                    double minSignalBaselineDifference,
+                    float productPpmTolerance,
+                    string scoringAlgo);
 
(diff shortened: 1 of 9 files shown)


src/core/libmaven/eiclogic.h
```diff
@@ -34,14 +34,17 @@ class EICLogic {
...
+                    Compound* compound,
+                    float grouping_maxRtWindow,
+                    double minQuality,
+                    double distXWeight,
+                    double distYWeight,
+                    double overlapWeight,
+                    bool useOverlap,
+                    double minSignalBaselineDifference,
+                    float productPpmTolerance,
+                    string scoringAlgo);
```
saifulbkhan: Should have included this file in the previous commit or the part from `EicWidget::groupPeaks` of last commit should have been included in this commit. I was wondering how this method was even compiling with two missing parameters.

src/gui/mzroll/eicwidget.cpp
```diff
@@ -198,6 +198,13 @@ void EicWidget::integrateRegion(float rtmin, float rtmax) {
 	}
 
 	eicParameters->_integratedGroup.groupStatistics();
+    int ms2Events = eicParameters->_integratedGroup.getFragmentationEvents().size();
+    if (ms2Events) {
+		float ppm = getMainWindow()->mavenParameters->fragmentTolerance;
+		string scoringAlgo = getMainWindow()->mavenParameters->scoringAlgo;
+        eicParameters->_integratedGroup.computeFragPattern(ppm);
+		eicParameters->_integratedGroup.matchFragmentation(ppm, scoringAlgo);
+    }
```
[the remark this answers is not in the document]
shubhra-agrawal: `getFragmentationEvents()` does not just fetch the ms2Event count but also calculates it. That's why the check. The calculation should probably happen during peakgrouping though, but manual integration does not use that workflow.

src/core/libmaven/eiclogic.h
shubhra-agrawal: That was a staging error on my part. Tried to break the whole change into different commits. Must have missed that.

Issue #42, opened 2016-11-07
Crash: segmentation fault when loading mzML samples when zlib is enabled


sahil21:
I enabled the zlib in the elMaven code (In mzroll.pri it was commented out) and while uploading the mzML samples it was giving an segmentation fault.

PS: I compiled the libmaven as dynamic library for this specific case as there was some issue when I enable ZLIB in static mode.

saifulbkhan:
El-MAVEN v0.8.0 and later will be able to read (mzML and mzXML) samples containing zlib compressed data.

PR #1024, opened 2019-04-11
Support samples containing zlib compressed data

shubhra-agrawal:
Decompression of zlib encoded data strings from sample files has not been possible for a long time. This issue has already been fixed in the upstream MAVEN project, by using a new base64 decoding
function that returns a string and then simply passing this string to the decompression function in mzUtils (if decompression is needed at all). This bypasses the original implementation where a
lot of raw buffer manipulation was being done before sending the string to the decompression function. The same mechanism will not be used in El-MAVEN.

Also part of this PR are the following related changes:
 - Remove unrequired methods from the base64 namespace.
 - Add checks for compression status in MzML files.
 -  Add ZLIB macro definitions to all platforms.
 - Use a Boost (iostreams) library for safe decompression of binary data.

### DIFF
--- a/mzroll.pri
+++ b/mzroll.pri
@@ -9,13 +9,13 @@ win32 {
     DEFINES += MINGW
     DEFINES += WIN32
     DEFINES += CDFPARSER
-    LIBS += -lcdfread -lnetcdf -lz
+    LIBS += -lcdfread -lnetcdf -lz -lboost_iostreams-mt
 }
 
 unix: {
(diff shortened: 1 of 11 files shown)


saifulbkhan:
Note: this should also fix #42.

shubhra-agrawal:
@saifulbkhan I have checked this with LCMS, PRM, DDA data against MAVEN. This was a mixture of mzXML and mzML files so I think we're safe. Merging this!

PR #1027, opened 2019-04-16
Support for mzrollDB format

saifulbkhan:
MAVEN users have started using a SQLite project format, mzrollDB, from which emDB was itself derived. However, a lot of (mostly naming) changes exist between the two formats. To allow easier transition of MAVEN users to El-MAVEN, mzrollDB format will now be allowed to be read in and a converted emDB file will be generated containing almost all information from the mzrolDB. This emDB file will then be made the session's current project.

Note: alignment data is not being saved at the moment because of difference of save/load algorithm of alignment data between the two applications.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -1107,7 +1107,7 @@ void MainWindow::saveProjectForFilename(bool tablesOnly)
 {
     if (fileLoader->isMzRollProject(_currentProjectName)) {
         _saveAllTablesAsMzRoll();
-    } else if (fileLoader->isSQLiteProject(_currentProjectName)) {
+    } else if (fileLoader->isEmdbProject(_currentProjectName)) {
         if (tablesOnly) {
             auto allTables = getPeakTableList();
             allTables.append(bookmarkedPeaks);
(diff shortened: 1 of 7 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -1707,8 +1707,8 @@ void MainWindow::open()
         this,
         "Select projects, peaks, samples to open:",
         dir,
-        tr("All Known Formats(*.mzroll *.emDB *.mzPeaks *.mzXML *.mzxml "
-           "*.mzdata *.mzData *.mzData.xml *.cdf *.nc *.mzML);;")
+        tr("All Known Formats(*.mzroll *.emDB *mzrollDB *.mzPeaks *.mzXML "
```
shubhra-agrawal: Typo "*.mzrollDB"

saifulbkhan:
@shubhra-agrawal 

MS2 event counts are being populated as they are in the database. A lot of groups having MS2 scores (> 0), have 0 as their MS2 event count (in the database) and that is reflected in El-MAVEN as well.

MaxAreaTopCorrected is not saved in mzrollDB and it is our default quantification type - that is why we do not see any values if the session is loaded from an mzrollDB. If we switch to a different quantitation type it gets populated as it should.

Fixed missing peak tables issue. Please take a look at the latest commit.

PR #1037, opened 2019-04-23
[UI] Add demo calendar link in PollyPhi video

shubhra-agrawal:


shubhra-agrawal:
@saifulbkhan Would have been preferable but there is no URL for it atm.

PR #1038, opened 2019-04-23
Fix peak classification

shubhra-agrawal:
There are two changes in this PR:
1. Parent and C12 parent groups were being classified differently by the new ML model due to a missing quality statistic in C12 parent.
2. El-MAVEN and PollyPhi were using different values to denote 'bad' and 'ambiguous' groups. This has been fixed on El-MAVEN's end

### DIFF
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -266,7 +266,7 @@ void CSVReports::writeDataForPolly(const std::string& file, std::list<PeakGroup>
         for(auto grp: groups) {
             for(auto child: grp.children) {
 
-                int mlLabel =  (child.markedGoodByCloudModel) ? 1 : (child.markedBadByCloudModel) ? -1 : 0;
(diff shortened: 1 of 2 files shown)


src/core/libmaven/csvreports.cpp
```diff
@@ -266,7 +266,7 @@ void CSVReports::writeDataForPolly(const std::string& file, std::list<PeakGroup>
         for(auto grp: groups) {
             for(auto child: grp.children) {
 
-                int mlLabel =  (child.markedGoodByCloudModel) ? 1 : (child.markedBadByCloudModel) ? -1 : 0;
+                int mlLabel =  (child.markedGoodByCloudModel) ? 1 : (child.markedBadByCloudModel) ? 0 : -1;
```
saifulbkhan: Took me some time to understand this is a nested ternary check. Let's try to use less of these in the future. It looks cool in this case though (somewhat elegant).

Issue #924, opened 2018-11-27
Performance improvement for baseline correction


surbhipoddar:
**Describe the bug**
Currently, the whole EIC is pulled to detect the baseline. This lowers the performance of the algorithm.

**Expected behavior**
We can limit the number of relevant points around the peak we detect while baseline correction instead of detecting the whole EIC. This will help improve the performance of the algorithm.

saifulbkhan:
Since we have started using optimization flags on all three platforms, we should see a fair bit of performance increase. There can still be some improvements such as the one suggested above.

saifulbkhan:
An attempt has been made (available from release v0.8.0) to downsample a signal before calculating its baseline. This should increase browsing and peak detection performance while AsLS Smoothing is turned on.

shubhra-agrawal:
@chubukov FYI. Since the major critique for AsLSS was around its performance, we have optimized it in v0.8.0. Please let us know if you see an improvement for your datasets.

saifulbkhan:
Nobody seems to notice any significant slowdown while detecting peaks using AsLS as the baseline algo. We did benchmarks too, it is only a little slower than the default (thresholding as I call it) for smaller number of scans and even outperforms it when the scan density increases.

PR #1042, opened 2019-04-25
Improve AsLS baseline estimation runtime

saifulbkhan:
AsLS has cubic time complexity and there is not much we can do about it since the 3rd party solver is the bottleneck (cubic). But, we can reduce the length of the intensity vector because, we do not need to consider every point in the intensity signal. A down-sampling method is used to shorten the intensity vector before feeding it into the AsLS algorithm. The obtained baseline is then up-sampled back to the size of original intensity vector.

Issue: #924


src/core/libmaven/mzUtils.cpp
```diff
@@ -1050,4 +1050,123 @@ Series:  Prentice-Hall Series in Automatic Computation
...
+
+        // beta for Kaiser window, 5.0 seems to work for all EICs that we tested
+        double beta = 5.0;
+
+        return firDesignKaiser(numTaps, fc, beta, interpRate);
+    }
+
+    int approximateResamplingFactor(const size_t dataSize)
+    {
+        int resamplingFactor = static_cast<int>(dataSize / 100) - 1;
```
shubhra-agrawal: Considering the time range can differ a lot between experiment (20min-50min as per personal experience), just vector length of intensity might not give the best resampling factor.

A high res dataset run for 20min and a low res dataset run for 30 min might have the same vector length but if the same resampling factor is used for both, the resulting vector for the low res data might lose some features. 
I would suggest using the sample rt range as another argument for deciding the resampling factor. What do you think?

saifulbkhan: I gave this some thought. I admit that the function responsible for calculating the resampling factor is not very robust and uses a makeshift algorithm. I agree that the sampling rate is an important factor for a simple downsampling solution, our decimation method low-pass filters a signal (to obey the Nyquist criteria) which tries to prevent aliasing (that would otherwise occur had we not filtered). Therefore, the combined method (low-pass filtering + downsampling) should be enough to allow us to disregard the time factor. This does not, however, mean that its immune to information loss or aliasing. But, to the resampling function the intensity vector is just an array values, its span is irrelevant.

If you still feel that we should consider the RT range, then we can work out some formula (but I have a feeling that it too would be ad hoc).

shubhra-agrawal: Summarizing our discussion here: 
- Downsampling should not lose any information even for low-res data as you explained. It's safe to merge this in.
- Since high res data has higher data density, there is greater potential for downsampling without losing information. For this, we can use average RT difference between scans to determine the level of downsampling done for different datasets. We should test this once, but the current solution is good and can be merged.

PR #1047, opened 2019-04-29
Improvements to sample ordering

saifulbkhan:
This patch contains the following changes and patches to sample ordering UX:
  * Add hooks for updating some widgets (where samples are information is present) as soon as order of samples in project (samples) widget changes or when any of the widgets is deselected/selected.
  * Fix some crashes due to volatile PeakGroup reference being held by isotope plot and group alignment viz widget.
  * Enable drag-and-drop for project (samples) widget. This change, however, will remove the ability to sort samples according to any of the columns in samples widget. We can discuss whether we should keep this change.

### DIFF
--- a/src/gui/mzroll/point.cpp
+++ b/src/gui/mzroll/point.cpp
@@ -141,9 +141,8 @@ void EicPoint::mousePressEvent (QGraphicsSceneMouseEvent* event) {
 
     if (_group) Q_EMIT peakGroupSelected(_group);
     if (_peak)  Q_EMIT peakSelected(_peak);
-
-    _mw->groupRtWidget->plotGraph(_group);
-
+    if (_group)
+        _mw->groupRtWidget->plotGraph(_group);
     if ( _group && _group->isIsotope() == false ) {
         _mw->isotopeWidget->updateIsotopicBarplot(_group);
     }
--- a/src/gui/mzroll/projectdockwidget.cpp
+++ b/src/gui/mzroll/projectdockwidget.cpp
@@ -226,11 +226,11 @@ void ProjectDockWidget::changeSampleOrder() {
     }
 
     if (changed) {
-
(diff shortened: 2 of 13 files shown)


src/gui/mzroll/point.cpp
```diff
@@ -141,9 +141,8 @@ void EicPoint::mousePressEvent (QGraphicsSceneMouseEvent* event) {
 
     if (_group) Q_EMIT peakGroupSelected(_group);
     if (_peak)  Q_EMIT peakSelected(_peak);
-
-    _mw->groupRtWidget->plotGraph(_group);
-
+    if (_group)
```
shubhra-agrawal: Instead of doing a null check every time a method is called, it might be better to move this within the method. What do you think?

src/gui/mzroll/treedockwidget.cpp
```diff
@@ -285,6 +286,7 @@ void TreeDockWidget::setupScanListHeader()
 }
 
 void TreeDockWidget::addScanItem(Scan* scan) {
+        _scansList.append(scan);
```
shubhra-agrawal: This should happen after the null-check in the next line.

src/gui/mzroll/point.cpp
saifulbkhan: I think it should be in both the places. The method or the caller, neither should assume that the group is a valid one.

src/gui/mzroll/treedockwidget.cpp
saifulbkhan: Added.

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -1077,3 +1070,34 @@ void ProjectDockWidget::unloadSample(mzSample* sample) {
         }
     }
 }
+
```
shubhra-agrawal: :+1:

PR #1055, opened 2019-05-14
Fix epi

rish9511:
**Main highlights of this PR:**  

1. Adds a new class called `DownloadManager`. It's meant to be a generic class, i.e It can be used by anyone that requires something to be downloaded from internet. In future we might make use of it for the auto-update functionality

2. Use just one instance of `PollyIntegration`.


src/cli/peakdetector/peakdetector.pro
```diff
@@ -21,7 +22,8 @@ INCLUDEPATH +=  $$top_srcdir/src/core/libmaven     \
                 $$top_srcdir/3rdparty/libcdfread   \
                 $$top_srcdir/src/pollyCLI          \
                 $$top_srcdir/3rdparty/obiwarp      \
-                $$top_srcdir/3rdparty/Eigen
+                $$top_srcdir/3rdparty/Eigen        \
+                $$top_srcdir/src/gui
```
sahil21: @rish9511 Is it a good idea to give access to src/gui folder in cli?

src/gui/mzroll/downloadmanager.cpp
```diff
@@ -0,0 +1,89 @@
+#include "downloadmanager.h"
+#include <QNetworkAccessManager>
+#include <QNetworkRequest>
+#include  <QNetworkReply>
+#include <QDebug>
+#include "pollyintegration.h"
```
shubhra-agrawal: @rish9511 Is "pollyintegration.h" being used in downloadmanager? If not, let's remove this dependency

src/cli/peakdetector/peakdetector.pro
```diff
@@ -55,12 +57,14 @@ macx {
...
           $$top_srcdir/src/core/libmaven/classifier.cpp          \
           $$top_srcdir/src/core/libmaven/classifierNeuralNet.cpp \
+          $$top_srcdir/src/gui/mzroll/downloadmanager.cpp        \
           main.cpp                                               \
           parseoptions.cpp                                       \
           peakdetectorcli.cpp
 
 HEADERS += $$top_srcdir/src/core/libmaven/classifier.h          \
            $$top_srcdir/src/core/libmaven/classifierNeuralNet.h \
+           $$top_srcdir/src/gui/mzroll/downloadmanager.h        \
```
saifulbkhan: I would suggest that we put `downloadmanager` in PollyCLI module instead. It fits as a common set of utilities to be shared by other modules.

shubhra-agrawal: Since `downloadmanager` might be used for version upgrade etc as well, it should be a separate module instead of pollyCLI. If @rish9511 only plans to use this for polly, then let's move it to pollyCLI

saifulbkhan: Makes sense. It should still not be part of the GUI module. Maybe move it to libmaven? But then we will have to find not-Qt alternatives to QNetwork type of functionality.

src/gui/mzroll/downloadmanager.h
```diff
@@ -0,0 +1,44 @@
...
+{
+    Q_OBJECT
+public:
+    DownloadManager();
+    void setRequest(const QString& url, void* requester, bool async=true);
+    bool err;
+//    enum class RequestType {
+//        ASYNC,
+//        SYNC
+//    };
```
saifulbkhan: Please do not leave commented code around.

src/pollyCLI/pollyintegration.cpp
```diff
@@ -22,26 +29,92 @@ PollyIntegration::PollyIntegration(): nodePath(""), jsPath("")
       QString binDir = qApp->applicationDirPath() + QDir::separator() + ".." + QDir::separator() + ".." + QDir::separator() + ".." + QDir::separator();
       if(!QStandardPaths::findExecutable("node", QStringList() << binDir + "node_bin" + QDir::separator() ).isEmpty())
         nodePath = binDir + "node_bin" + QDir::separator() + "node";
+    #endif
+
+     _dlManager->setRequest("https://raw.githubusercontent.com/ElucidataInc/polly-cli/master/index.js", this);
```
sahil21: Change from 
`_dlManager->setRequest("https://raw.githubusercontent.com/ElucidataInc/polly-cli/master/index.js", this)`
to
`_dlManager->setRequest("https://raw.githubusercontent.com/ElucidataInc/polly-cli/master/prod/index.js", this)`

Reference: https://github.com/ElucidataInc/polly-cli/pull/4

src/gui/mzroll/downloadmanager.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+DownloadManager::DownloadManager():
+    _manager(new QNetworkAccessManager),
+    _reply(nullptr),
+    err(false)
+{
+}
+
+void DownloadManager::setRequest(const QString &url, void *requester, bool async)
```
saifulbkhan: Is the `requester` not being lost in this method? Maybe we can remove it if not needed.

src/cli/peakdetector/peakdetector.pro
shubhra-agrawal: It can be an independent module like pollyCLI and projectDB. Does not need to go into libmaven, necessarily.

src/gui/mzroll/downloadmanager.h
```diff
@@ -0,0 +1,44 @@
...
+
+
+private:
+    void download(bool async=true);
+
+public slots:
+    void dataAvailable();
+    void finished();
+    void error(QNetworkReply::NetworkError error);
```
saifulbkhan: I think the above three slots can be either private or protected. No need to expose these, since only the manager needs to call to them.

src/gui/mzroll/downloadmanager.cpp
```diff
@@ -0,0 +1,86 @@
...
+        emit downloaded();
+    }
+}
+
+
+void DownloadManager::error(QNetworkReply::NetworkError error)
+{
+    qDebug() << "error code :" << error;
+    qDebug() << "error string: "<< _reply->errorString();
+    err = true;
```
saifulbkhan: Don't you think storing the error string might be helpful, instead of just printing them? It could help us to inform the user on GUI or CLI about what the precise error was. The string can be empty if there was no error. We could add a checker (e.g. `errorOccurred`) to check whether the request terminated in error state, and we won't need the boolean `err` value.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -73,9 +74,9 @@ PollyElmavenInterfaceDialog::~PollyElmavenInterfaceDialog()
         delete (_loginform);
 }
 
-EPIWorkerThread::EPIWorkerThread()
+EPIWorkerThread::EPIWorkerThread(PollyIntegration* iPolly):
```
shubhra-agrawal: The only point of `EPIWorkerThread` was to have a new `pollyIntegration` object for every process. This was introduced to handle parallel processing in different apps which is not an option anymore since we now block the UI when one app is running.

So ideally, `EPIWorkerThread` class should be removed, even though it shouldn't lead to any practical issues in the functionality. Just want to make sure you have taken this into consideration

src/gui/mzroll/controller.h
```diff
@@ -15,7 +16,7 @@ class Controller:  public QObject
     Controller();
     ~Controller();
     MainWindow* getMainWindow() { return _mw; }
-
+    PollyIntegration* _iPolly;
```
saifulbkhan: Let us not have names of non-private members starting with an underscore.

src/gui/mzroll/pollyelmaveninterface.h
```diff
@@ -331,12 +331,14 @@ private Q_SLOTS:
...
 
+class PollyIntegration;
+
 class EPIWorkerThread : public QThread
 {
     Q_OBJECT
 
 public:
-    EPIWorkerThread();
+    EPIWorkerThread(PollyIntegration* iPolly);
```
saifulbkhan: Thank you for this grand unification!

src/cli/peakdetector/peakdetector.pro
saifulbkhan: In that case, this new module can merge with pollyCLI. Then the module will accommodate classes responsible for making requests over the network.

src/gui/mzroll/downloadmanager.cpp
rish9511: @shubhra-agrawal  
Yes, we are using it.

rish9511: Yes, it's being used

https://github.com/ElucidataInc/ElMaven/pull/1055/files#diff-de0fde6efde85e103aaf21acf14c9f3aR20

src/cli/peakdetector/peakdetector.pro
rish9511: Let's make a new folder called "common" ?
Even PollyCLI  can be part of it and can we save emdb  format from cli ?

src/pollyCLI/pollyintegration.cpp
rish9511: Okay, will do

src/gui/mzroll/pollyelmaveninterface.cpp
rish9511: I did test it out and everything seemed to work fine.
We should get rid of `EPIWorkerThread` if it's not being used

rish9511:
@sahil21 
Yes, index.js file should be removed as well. Will do that

src/pollyCLI/pollyintegration.cpp
rish9511: @sahil21 
The above url is nor working

src/cli/peakdetector/peakdetector.pro
rish9511: We should have a folder called "common" ideally.
We plan to change it

src/cli/peakdetector/peakdetector.pro
shubhra-agrawal: Sounds good! If it can be done by EoD, let's do it. If not, we can add it to the next release

src/gui/mzroll/downloadmanager.h
rish9511: Done Intentionally.
I want to have this enum. 
Can we leave it like this ?

src/gui/mzroll/pollyelmaveninterface.cpp
shubhra-agrawal: Yes. Can you add a TODO? We will take care of it in the next release.

src/gui/mzroll/downloadmanager.h
saifulbkhan: Oh, in that case let's add a `TODO` or something to notify readers of its future use.

src/gui/mzroll/downloadmanager.h
rish9511: They are connected to QNetworkReply's signals, hence `public`

saifulbkhan: I can only see these connections being made within this class itself. So making them private should not violate any rules, right?

src/gui/mzroll/downloadmanager.h
rish9511: Sure

src/gui/mzroll/downloadmanager.cpp
rish9511: `requester` can be anyone. For now it's just `PollyIntegration`, so it may seem like there's no point of `requester` but in future we might have more requesters. 
How will more `requesters` be handled that is still something that needs to be figured out.

src/gui/mzroll/downloadmanager.h
rish9511: They will be called by `QNetworkReply` and not by `DownloadManager`.
Let me know if you still have doubts regarding this

src/gui/mzroll/controller.h
rish9511: Ah, yes! 
Sorry about this

src/gui/mzroll/downloadmanager.cpp
rish9511: We should ideally

src/cli/peakdetector/peakdetector.pro
rish9511: Done

rish9511:
@saifulbkhan @shubhra-agrawal @sahil21 
Have resolved most of the comments
Please have a look again

src/gui/mzroll/downloadmanager.cpp
saifulbkhan: So multiple objects could request through a single DownloadManager? I was under the impression that one such object would be used for only one request. In case of multiple requesting clients, we would have to have multiple data values too, since those requests could be made at different points of time so the data returned each time (for the same URL) could be different.

For now, we can skip all these considerations, given our time constraint. Resolving this conversation.

src/gui/mzroll/downloadmanager.h
saifulbkhan: I am probably not understanding this well, but I made the slots private locally and it worked fine for me. That was all I was asking for.

src/gui/mzroll/downloadmanager.cpp
saifulbkhan: Yes. But since we are short on time, we can make this change when its actually needed. Resolving.

src/gui/mzroll/downloadmanager.h
rish9511: Thanks for checking, i just checked on my local as well and it worked. 
Wasn't expecting it to work though, I always thought they followed the same conventions as normal member functions but no, just read this on [Qt docs](https://doc.qt.io/qt-5/signalsandslots.html#slots):  "This means that a signal emitted from an instance of an arbitrary class can cause a private slot to be invoked in an instance of an unrelated class." 

Sorry my bad, will make this change

saifulbkhan:
@rish9511 Are the _index.js_xxxxxx files being deleted after every `PollyIntegration` object is destroyed? Because running the CLI n number of times leaves around n different index.js files in my bin folder.

rish9511:
@saifulbkhan 
> Are the _index.js_xxxxxx files being deleted after every `PollyIntegration` object is destroyed? Because running the CLI n number of times leaves around n different index.js files in my bin folder.

Not in CLI, thanks for pointing out. Fixing this

rish9511:
@saifulbkhan @shubhra-agrawal 
Made some fixes and changes wherever pointed out 
Have a look again  and let me know if anything's still left

PR #1056, opened 2019-05-14
Block unlicensed apps

shubhra-agrawal:
- Remove index.js file since it is pulled dynamically
- Fetch redirection URL templates dynamically
- Use app names instead of component names
- Add adverts for unlicensed apps in EPI dialog
- Change app name from Fluxomics to PollyPhi

### DIFF
--- a/src/cli/peakdetector/peakdetectorcli.cpp
+++ b/src/cli/peakdetector/peakdetectorcli.cpp
@@ -864,10 +864,15 @@ QString PeakDetectorCLI::_getRedirectionUrl(QString datetimestamp,
     switch (_currentPollyApp) {
     case PollyApp::PollyPhi: {
         QString workflowId = 
-            _pollyIntegration->obtainComponentId(PollyApp::PollyPhi);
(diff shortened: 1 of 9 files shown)


src/cli/peakdetector/peakdetectorcli.cpp
```diff
@@ -857,19 +857,18 @@ QString PeakDetectorCLI::_getRedirectionUrl(QString datetimestamp,
     QString redirectionUrl = "";
     switch (_currentPollyApp) {
     case PollyApp::PollyPhi: {
+        QString workflowId = 
+            _pollyIntegration->obtainComponentId(PollyApp::PollyPhi);
```
saifulbkhan: We have a different function for obtaining workflow Id. So, this should be

```C++
_pollyIntegration->obtainWorkflowId(PollyApp::PollyPhi);
```

I am experiencing a crash due to this when I run CLI. The failure to obtain `workflowId` should still not crash, and can be handled better.

PR #1057, opened 2019-05-15
Show EPI error when any API breaks

shubhra-agrawal:
- If API returns an error message, display it in the format:
error Message
Type: errorCode
Contact details for support

- If API does not return a proper error message that can be read by El-MAVEN, the complete standard error will be displayed
- If there is any other error in EPI that leads to a failure, a general message is displayed with contact details for support

### DIFF
--- a/src/pollyCLI/pollyintegration.cpp
+++ b/src/pollyCLI/pollyintegration.cpp
@@ -242,7 +242,7 @@ QString PollyIntegration::obtainComponentId(PollyApp app)
     for (auto elem : array) {
         auto jsonObject = elem.toObject();
         auto name = jsonObject.value("component").toString();
-        if (name == componentName)
+        if (name == _stringForApp(app))
             return QString::number(jsonObject.value("id").toInt());
     }
     return QString::number(-1);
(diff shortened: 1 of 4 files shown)


src/pollyCLI/pollyintegration.cpp
```diff
@@ -282,22 +295,75 @@ QStringList PollyIntegration::get_project_upload_url_commands(QString url_with_w
...
+        qDebug() << "**************";
+        qDebug() << errorString;
+        //errorString.replace("\n", "");
+        if (!errorString.isEmpty()) {
+            emit sendEPIError(errorString);
+            return true;
+        }
+    } else if (resultAndError.size() == 0) {
+        //no response or error
+        emit sendEPIError("Qt Process failed " + supportMessage);
```
saifulbkhan: This string would look better on a dialog with a full stop after "failed". "Qt Process failed. Contact..."

src/pollyCLI/pollyintegration.cpp
```diff
@@ -282,22 +295,75 @@ QStringList PollyIntegration::get_project_upload_url_commands(QString url_with_w
...
+                message = jsonObj[key].toString();
+            }
+            if (key == "type") {
+                type = jsonObj[key].toString();
+            }
+        }
+
+        if (!message.isEmpty() || !type.isEmpty()) {
+            QString errorMessage = message + "\n" +
+                                   "Type: " + type + "\n" +
```
saifulbkhan: Can we show "Type" only when the `type` string actually has some content? It appears as if we missed some information by showing empty _Type_ row.

Screenshot 2019-05-15 at 5 34 45 PM

PR #1061, opened 2019-05-16
Update README for Windows

rish9511:
QtNetwork depends on extra [OpenSSL dlls](https://indy.fulgan.com/SSL/openssl-1.0.2r-x64_86-win64.zip) to work on Windows.  

NOTE: We should not depend on [Stack Overflow](https://stackoverflow.com/a/34586308) for this and try to find a standard solution

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
 ### Windows
 
 - Download [MSYS2](http://www.msys2.org/) installer and follow the installation instructions provided on their website.
+- Download OpenSSL package using https://indy.fulgan.com/SSL/openssl-1.0.2r-x64_86-win64.zip
+- Extract the contents of OpenSSL package in `/c/msys64/mingw64/bin`
+- NOTE: To verify whether the above two steps have been executed correctly make sure you have libeay32.dll and ssleay32.dll inside `/c/msys64/mingw64/bin/`
 - Open MSYS2 and give the following commands to set up libraries and tool chains for El-MAVEN. Reopen MSYS2 when required:
 ##### **For 64 bit**:  _Following commands needs to be executed from msys2 shell_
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,9 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
 - `pacman --force -Syu`
 - `pacman --force -Su`
 - `pacman --force -Sy base-devel msys2-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-qt5 mingw64/mingw-w64-x86_64-hdf5 mingw64/mingw-w64-x86_64-netcdf mingw64/mingw-w64-x86_64-boost msys/git mingw-w64-x86_64-curl`
+-  `Download OpenSSL package using https://indy.fulgan.com/SSL/openssl-1.0.2r-x64_86-win64.zip`
```
saifulbkhan: Since this is not a command, can we write this as a regular instruction instead. Like:
> Download OpenSSL package using `https://indy.fulgan.com/SSL/openssl-1.0.2r-x64_86-win64.zip`

Alternatively, we could have a `curl` command in place for the equivalent effect instead of instructing them to download.

README.md
```diff
@@ -43,6 +43,9 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
...
+- `Extract the contents of OpenSSL package in /c/msys64/mingw64/bin`
```
saifulbkhan: Again this is an instruction not a shell command. So we can either have this:
> Extract the contents of OpenSSL package in `/c/msys64/mingw64/bin`

... or have an alternative unzip command such as:
`tar -xvf openssl-1.0.2r-x64_86-win64.zip /c/msys64/mingw64/bin`

README.md
```diff
@@ -43,6 +43,9 @@ Contributers can build El-MAVEN on Windows, Ubuntu or Mac systems by following t
...
+- `NOTE: To verify whether the above two steps have been executed correctly make sure you have  libeay32.dll and ssleay32.dll inside /c/msys64/mingw64/bin/`
```
saifulbkhan: Again, this is a suggestion not a command.

shubhra-agrawal:
@chubukov We have made some changes that might lead to a compilation error on your system. Please check out the updated Windows build instructions in the README before you build the latest develop. Apart from these instructions, all you need to do is place the Sentry key file in the root folder before compilation.

Let us know if we have missed anything.

PR #1064, opened 2019-05-21
Update widgets from EicPoint using static methods

saifulbkhan:
An EicPoint object drawn on an EIC is destroyed whenever that EIC
window scene is cleared. The EIC is cleared everytime it is
resized, so whenever a user clicks on an MS2 marker in the EIC
window, an EicPoint for that MS2 marker will trigger opening of
spectra widget, which in turn changes the size of EIC window
causing the EicPoint itself to be deleted before it could perform
other necessary widget updates. This was causing a crash because
the method in call was trying to use invalidated variables, past
their lifetime.

With this patch, all direct widget updates are now handled by new
static methods of the EicPoint class, that have stable copies of
required pointers.

### DIFF
--- a/src/gui/mzroll/point.cpp
+++ b/src/gui/mzroll/point.cpp
@@ -129,61 +129,57 @@ void EicPoint::mouseDoubleClickEvent(QGraphicsSceneMouseEvent*) {
     }
 
 }
-void EicPoint::mousePressEvent (QGraphicsSceneMouseEvent* event) {
-    //if (_group) _group->groupOveralMatrix();
 
+void EicPoint::mousePressEvent (QGraphicsSceneMouseEvent* event) {
     if (event->button() == Qt::RightButton)  {
         contextMenuEvent(event);
         return;
     }
 
     setZValue(10);
-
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/point.cpp
```diff
@@ -129,61 +129,61 @@ void EicPoint::mouseDoubleClickEvent(QGraphicsSceneMouseEvent*) {
...
+        emit peakSelected(_peak);
+
+    // make changes through static functions, since this object might be
+    // destroyed during the execution period of those functions.
+    auto mw = &(*_mw);
+    auto group = &(*_group);
+    auto peak = &(*_peak);
+    auto scan = &(*_scan);
+    _updateWidgetsForPeakGroup(_mw, _group, _peak);
+    _updateWidgetsForScan(_mw, _scan);
```
shubhra-agrawal: @saifulbkhan `mw`, `group`, `peak`, and `scan` are not being used, right?

saifulbkhan: Ah! I forgot to remove this from an earlier implementation. The pointers will be copied as arguments. Will clear all the unnecessary copies.

PR #1067, opened 2019-05-22
Fix untargeted EPI crash

saifulbkhan:
This PR contains the following fixes:
 -  Prevent crash while preparing untargeted groups for Polly.
 -  Prevent crash while preparing untargeted groups in CLI.
 -  Block untargeted groups from being uploaded to Polly.


shubhra-agrawal:
So the crash is resolved but untargeted table is still showing up in the table drop-down in FirstView.

PR #1083, opened 2019-05-31
Remove loess fit

saifulbkhan:
This set of patches not only removes the deprecated "Loess fit" alignment feature, but also fixes the issues in visual appearance of elements in Alignment dialog when viewed on low resolution screens.


src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -485,54 +414,6 @@ void BackgroundPeakUpdate::align() {
                         aligner.doAlignment(groups);
                         mainwindow->sampleRtWidget->setDegreeMap(aligner.sampleDegree);
                         mainwindow->sampleRtWidget->setCoefficientMap(aligner.sampleCoefficient);
-                } else if (alignAlgo == 2) {
-                        mainwindow->getAnalytics()->hitEvent("Alignment", "LoessFit");
-                        aligner.preProcessing(groups, mavenParameters->alignWrtExpectedRt);
```
shubhra-agrawal: `Aligner::preProcessing` is not being used anywhere. It was written to communicate data to the python executable from what I understand.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -485,54 +414,6 @@ void BackgroundPeakUpdate::align() {
...
-                                else{
-                                        errorMessage=errorMessage+"<br>"+"Incomplete data, re-run alignment";
-                                }
-                                qDebug()<<errorMessage;
-                                Q_EMIT alignmentError(errorMessage);
-                                return;
-                        }
-
-                        if(!parentObj.isEmpty()){
-                                aligner.updateRts(parentObj);
```
shubhra-agrawal: `Aligner::updateRts` is not being used anywhere and can be removed

src/gui/mzroll/background_peaks_update.cpp
saifulbkhan: Removed.

src/gui/mzroll/background_peaks_update.cpp
saifulbkhan: Deleted these as well.

Issue #859, opened 2018-09-05
Optimize Obi-warp


rish9511:
**Is your feature request related to a problem? Please describe.**
Obi-Warp takes a huge amount of time to perform Alignment, especially in the case of large sample files.
Many times it also leads to the UI getting freezed which ultimately forces the users to quit the application.
It will be of great help if the algorithm is optimized 
  

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
**Misc Todo**
1. Add all settings of alignment dialog in default_settings.xml
2. Give the option to restore to  default alignment settings.

shubhra-agrawal:
Obi-warp alignment is multiprocessed in v0.6.0 and above leading to ~3-4x increase in speed. Next step will be to resolve the known crash points and optimization for further speed improvement.

saifulbkhan:
More improvement has been made to improve run-time of OBI-Warp (sparse OBI-Warp followed by segmented interpolation) in release v0.8.0.

saifulbkhan:
OBI-Warp is fast enough now. Closing this.

PR #1085, opened 2019-05-31
Remove unnecessary vector copy

saifulbkhan:
Temporary matrices being used in `find_path` method are attempted
to be copied into ObiWarp's class matrices at the end. But these
matrices themselves are not being used anywhere else. The copy
operation was causing a crash due to attempts to free unallocated
memory. The copy operations have been removed to prevent this
crash.

Issue: #859


shubhra-agrawal:
Develop crashes 3 out of 10 times in obi-warp with 16 samples. This branch did not crash once in 10 times.
Merging this!

PR #1086, opened 2019-06-07
Fix filterline 

shubhra-agrawal:
- Remove 'scan' attribute from filterline
- Read `filter string` attribute as filterline from spectrum list, if available. Continue to use `runId`, otherwise


shubhra-agrawal:
@saifulbkhan the filterString contains some information about the settings used for a particular scan. Let's take the case of MRM data. 

For MRM data, the filterString contains the precursor Mz, product Mz and collision energy usually. All scans with this filterString are clubbed together under the same SRM ID so you can pull the EIC.  Sometimes there is additional information in the filterString like scan number, sample number, or time info which means the string matching fails and all these scans end up having distinct SRM IDs which is not desirable. The second commit is telling the method to remove the "scan" field from the filterString. 

Let me know if I'm not clear.

saifulbkhan:
Oh this is very clear. Thank you. I understand much better now.

PR #1087, opened 2019-06-10
Segmented alignment (read and write)

saifulbkhan:
This PR involves the following related changes:
1. Alignment data from `mzrollDB` will now be read in successfully, and restored using the segmented alignment technique used in MAVEN.
2. El-MAVEN will also use the same technique to save and restore alignment data, from now on, because it is potentially faster and shows minimal deviation from true alignment.

### DIFF
--- a/src/core/libmaven/mzAligner.cpp
+++ b/src/core/libmaven/mzAligner.cpp
@@ -581,15 +581,15 @@ bool Aligner::alignWithObiWarp(vector<mzSample*> samples,
 float AlignmentSegment::updateRt(float oldRt)
 {
     // fractional distance from start of a segement
-    if (oldRt >= seg_start and oldRt <= seg_end) {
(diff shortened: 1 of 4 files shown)


src/core/libmaven/mzAligner.cpp
```diff
@@ -577,3 +577,69 @@ bool Aligner::alignWithObiWarp(vector<mzSample*> samples,
...
+                    && scan->rt <= segment->seg_end) {
+                    seg = segment;
+                    break;
+                }
+            }
+
+            if (seg) {
+                double newRt = seg->updateRt(scan->rt);
+                scan->rt = newRt;
+                corcount++;
```
shubhra-agrawal: Not sure if this variable is being used somewhere?

saifulbkhan: Removed it.

PR #1091, opened 2019-06-10
Remove breakpoint in untargeted detection

shubhra-agrawal:
There was a breakpoint in untargeted peak detection.
If there were 1000 consecutive mass slices without any valid groups, the peak detection process was aborted.

Removing this breakpoint as it provides no obvious advantage.
At the same time, certain low-intensity datasets hit the breakpoint very frequently leading to fewer groups reported in untargeted detection


saifulbkhan (approved):
Makes sense. Approving.

PR #1093, opened 2019-06-10
Remove un-necessary function call

rish9511:
I have been profiling `libmaven` lately, mainly because I wanted to find out the reason as to why MavenTests take so much time to run in debug mode(CI needs to run tests in debug mode for test coverage info). The profiling information generated led me to one of the functions, `PeakDetector::addPeakGroup`, which was being called inside a loop and the result of which was not being used anywhere, I used GCC's warning to double check this. 
```
PeakDetector.cpp:407:32: warning: ignoring return value of ‘bool PeakDetector::addPeakGroup(PeakGroup&)’, declared with attribute warn_unused_result [-Wunused-result] addPeakGroup(group);
```

The function, `PeakDetector::addPeakGroup`, was being used to check the `rtOverlap` between the groups but was not being used in any decision making.  

Irrespective of whether the overall running time of MavenTests gets better or not, this function call should be removed. 

Please let me know if there's a reason why this function exists that I might have missed

**Update**
The second commit, 98247f71e449cd9db69b5fd95f1584dab88e7f28,  reduces the amount of time spent in `mzSample::caluclatemzRtRange`.   Reason being, Deque's [] operator is slow and it was being used at a lot of places that too inside a loop

### DIFF
--- a/src/core/libmaven/PeakDetector.cpp
+++ b/src/core/libmaven/PeakDetector.cpp
@@ -403,8 +403,7 @@ void PeakDetector::processSlices(vector<mzSlice *> &slices, string setName)
             if (j >= mavenParameters->eicMaxGroups)
                 break;
 
-            PeakGroup group = peakgroups[j];
-            addPeakGroup(group);
+            mavenParameters->allgroups.push_back(peakgroups[j]);
         }
 
         //cleanup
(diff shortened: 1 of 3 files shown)


rish9511:
@saifulbkhan 
Added another commit, please review again.

rish9511:
According to the profiling information there was a significant performance gain.  
Yes, Deque's [] does work in constant time but according to this [post](https://stackoverflow.com/a/18812132) it does a  lot of extra work making it slower compared to Vector's [] 

I also don't understand this completely but profiling information did help me verify the slowness

PR #1094, opened 2019-06-10
Allow filtering untargeted groups having no MS2 events

saifulbkhan:
This patch allows the user to filter untargeted groups based on whether it has any fragmentation events or not. This helps in narrowing down the number of groups they want to focus on in their DDA dataset.

### DIFF
--- a/src/gui/mzroll/forms/peakdetectiondialog.ui
+++ b/src/gui/mzroll/forms/peakdetectiondialog.ui
@@ -61,34 +61,40 @@
             </property>
            </widget>
           </item>
-          <item row="2" column="2">
-           <widget class="QLabel" name="label_19">
-            <property name="text">
-             <string>Limit Intensity Range</string>
+          <item row="0" column="3">
+           <widget class="QDoubleSpinBox" name="mzMin">
+            <property name="maximum">
+             <double>999999999.000000000000000</double>
             </property>
            </widget>
           </item>
-          <item row="1" column="4">
-           <widget class="QDoubleSpinBox" name="rtMax">
+          <item row="1" column="0">
+           <widget class="QLabel" name="label">
+            <property name="text">
+             <string>Time Domain Resolution</string>
+            </property>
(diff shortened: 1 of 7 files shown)


shubhra-agrawal:
@saifulbkhan Some conflicts here. Please rebase this and I'll test it tomorrow.

src/gui/mzroll/forms/peakdetectiondialog.ui
```diff
@@ -202,31 +216,17 @@
...
-           <widget class="QDoubleSpinBox" name="mzMin">
-            <property name="maximum">
-             <double>999999999.000000000000000</double>
-            </property>
-           </widget>
-          </item>
-          <item row="1" column="2">
-           <widget class="QLabel" name="label_18">
+          <item row="2" column="0" colspan="2">
+           <widget class="QCheckBox" name="mustHaveMs2">
```
shubhra-agrawal: @saifulbkhan If I check the box and reopen El-MAVEN, the box goes back to its unchecked state. Is that by design?

shubhra-agrawal: Okay, I see the advantage. If I load LCMS data in the next session, I would not want this box checked automatically. This is fine for now.

saifulbkhan: I do not recall making this decision explicitly. For it to save its checked state we might have to introduce a variable for this in the settings file as well. I can do that, if that is what we want.

shubhra-agrawal: @saifulbkhan Two ways to handle the situation I mentioned:
1. Disable/enable the checkbox based on LCMS/DDA data
2. Keep the behavior as it is but save the state in emDB

saifulbkhan: Got it. Will change it accordingly.

PR #1099, opened 2019-06-11
Handle EPI index.js fetch errors

saifulbkhan:
This set of commits includes the following changes:
1. Add better message for missing `index.js` file.
2. Fix login form button and status not being reset after EPI failure message box is accepted.
3. Fix EPI error message box spawning at random position on user's screen, when errors occur at login window.
4. Minor - change miscellaneous error labels to have a more appropriate prefix.
5. Add another check for `index.js` file before starting any process for communication through it.

### DIFF
--- a/src/pollyCLI/pollyintegration.cpp
+++ b/src/pollyCLI/pollyintegration.cpp
@@ -382,7 +382,9 @@ bool PollyIntegration::_hasError(QList<QByteArray> resultAndError)
         }
 
         // the most likely other reason could be…
-        emit receivedEPIError("Error: Unable to connect to the internet.\n\n"
+        emit receivedEPIError("There was an unexpected error. Please make sure "
+                              "you are connected to the internet and try again."
+                              "\n\n"
                               + supportMessage);
         return true;
     }
--- a/src/pollyCLI/pollyintegration.cpp
+++ b/src/pollyCLI/pollyintegration.cpp
@@ -358,7 +358,7 @@ bool PollyIntegration::_hasError(QList<QByteArray> resultAndError)
         QString errorString = QString::fromStdString(errorResponse.toStdString());
         errorString.replace("\n", "");
         if (!errorString.isEmpty()) {
-            QString errorMessage = "Unknown error: " + errorString + "\n\n" +
+            QString errorMessage = "Error: " + errorString + "\n\n" +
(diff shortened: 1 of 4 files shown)


src/pollyCLI/pollyintegration.cpp
```diff
@@ -357,14 +368,22 @@ bool PollyIntegration::_hasError(QList<QByteArray> resultAndError)
         QString errorString = QString::fromStdString(errorResponse.toStdString());
         errorString.replace("\n", "");
         if (!errorString.isEmpty()) {
-            QString errorMessage = "Unknown error: " + errorString + "\n" +
+            QString errorMessage = "Error: " + errorString + "\n\n" +
                                    supportMessage;
             emit receivedEPIError(errorMessage);
             return true;
         }
     } else if (resultAndError.size() == 0) {
```
shubhra-agrawal: @saifulbkhan In case of missing index.js file, the `resultAndError` list will be size 2 (Check `runQtProcess`:100). The list will only be empty in case of unknown errors originating from the index.js file

saifulbkhan: Oh! I just remembered why I added this final check. I think that, any kind of error during the execution of JS functions would at least have some kind of error message, and most of the functions (last time I checked) are printing them to the error stream, so our `resultAndError` should have that (and if not we should maybe ensure that) if the failure occurs during runtime. But if the file was downloaded, but can no longer be found (path changed, deleted, etc.), then the byte array will be empty because the execution call never takes place and consequently neither result nor error streams are written to. RIght? This was my motive. Let me know if you think I am wrong somewhere on that line.

shubhra-agrawal: got it. I forgot that the first `if` displays error messages without `type` and `message` keys as well, so the "index.js file could not be downloaded" will be taken care of in that block.

In that case, my only issue is with the message being sent for internet connectivity error. While internet connectivity could be the reason why EPI failed to perform a task, in my experience (previous EPI testing) that has not been the case. It might be better to say something along the lines of "There was an unexpected error. Please check your internet connection and try again." + supportMessage.

saifulbkhan: Okay. Making that change.

PR #1105, opened 2019-07-01
Fix include heirarchy

saifulbkhan:
A lot of header files in our source (and test) directory were including each other and a few of the more important headers were being included everywhere regardless of whether their declarations were needed in a source file or not. This not only, increases the number of translation units compiled after a change to a header file, but also increases the overhead of the build system by making it resolve entire classes, where a simple forward declaration would have been sufficient.

The goal of this set of patches is to fix as much of the above problem as possible. Unnecessary, inclusion of header files into other header files has been reduced to a great extent.

Note 1: It is also important that we be careful about not repeating this mistake in the future when making changes to our source. Authors and reviewers should both be cautious regarding this issue.

Note 2: I do not expect these commits to cause any unexpected regressions (in the form of crashes/loss of functionality), since, I believe, no logical change was made to any source file. But please compile and test out some basic functionalities of the El-MAVEN binaries. I have done the same on my end.


src/gui/mzroll/mzroll.pro
```diff
@@ -238,6 +238,7 @@ HEADERS +=  stable.h \
 SOURCES += mainwindow.cpp  \
 database.cpp \
  plotdock.cpp \
+ spectralhit.cpp \
```
shubhra-agrawal: how was this not present in the pro file already :o

saifulbkhan: Because I created it with that commit. The declaration and the definitions were both within `spectralhit.h` header file.

PR #1110, opened 2019-07-08
Enable optimizations for release builds

saifulbkhan:
Various flags that are required for the compiler to perform optimizations were being removed for all AppVeyor and Travis CI builds. Now they will only be disabled for PR builds, which still need the debug information (destroyed by optimizations) for proper test coverage calculations.


shubhra-agrawal:
Much faster on Windows!

PR #1116, opened 2019-07-17
Fix and optimize OBI-Warp alignment

saifulbkhan:
This PR primarily contains two changes that improve the overall user experience with OBI-Warp alignment using El-MAVEN:
 1. Make changes to the OBI-Warp library itself such that it uses safer (i.e., no manual allocation/deallocation or pointer based indexing) STL containers for its core vector (`VecF` `VecD` and `VecI`) and matrix (`matF`, `matD` and `matI`) data structures. These changes were made because, even after a [previous fix](https://github.com/ElucidataInc/ElMaven/pull/1085), we were observing certain crashes due to freeing of allocated memory associated with vectors and matrices at multiple different points in the library. While this whole change has had some effect on performance, the most often called methods have been optimized to keep the decline in performance to a minimal.
 2. Call OBI-Warp from `mzAligner` with only a sparse representation of the data set being aligned. This improves the runtime (since warping on less data points will obviously be faster) and we can interpolate the updated RT values for the remaining data points that were not sent to OBI-Warp for alignment. The interpolation logic is the same as the one that we use when restoring alignment values from emDB, using the segmented alignment technique (borrowed from MAVEN).

### DIFF
--- a/3rdparty/obiwarp/dynprog.cpp
+++ b/3rdparty/obiwarp/dynprog.cpp
@@ -165,18 +165,24 @@ void DynProg::bijective_anchors(VecI &mCoords, VecI &nCoords, VecF &scores, VecI
     //puts("equivs:"); mCoords.print(); nCoords.print(); scores.print();
     int lengthEquiv = mCoords.dim();
 
-    int *nC_arr = new int[lengthEquiv];  // Max length...
-    int *mC_arr = new int[lengthEquiv];  // Max length...
-    float *sC_arr = new float[lengthEquiv];  // Max length...
+    std::vector<int> nC_arr;
+    nC_arr.resize(lengthEquiv);  // Max length...
+    std::vector<int> mC_arr;
+    mC_arr.resize(lengthEquiv);  // Max length...
+    std::vector<float> sC_arr;
+    sC_arr.resize(lengthEquiv);  // Max length...
 
     // monoMaps do NOT include the first and last equivalents!
 
     // find the length of the equivalents for MONOTONIC pairs (gaps collapse)
 
     // Eliminate the equivalents that have the same m or n as the two corners
-    int *mCoordsS_arr = new int[lengthEquiv-2];
-    int *nCoordsS_arr = new int[lengthEquiv-2];
-    float *scoresS_arr = new float[lengthEquiv-2];
+    std::vector<int> mCoordsS_arr;
+    mCoordsS_arr.resize(lengthEquiv-2);
+    std::vector<int> nCoordsS_arr;
+    nCoordsS_arr.resize(lengthEquiv-2);
+    std::vector<float> scoresS_arr;
+    scoresS_arr.resize(lengthEquiv-2);
     int last_index = lengthEquiv - 1;
     int startm = mCoords[0];
     int startn = nCoords[0];
@@ -377,17 +383,20 @@ void _traceback(MatI &tb, MatF &smat, int m, int n, MatI &tbpath, VecI &mCoord, 
         cnt++;
     }
 
-    int *tmpEquiv_m = new int[cnt];
-    int *tmpEquiv_n = new int[cnt];
-    float *tmpScores = new float[cnt];
+    std::vector<int> tmpEquiv_m;
+    tmpEquiv_m.reserve(cnt);
+    std::vector<int> tmpEquiv_n;
+    tmpEquiv_n.reserve(cnt);
+    std::vector<float> tmpScores;
+    tmpScores.reserve(cnt);
 
     // Reverse the arrays
     int rev = cnt - 1;
     // This could be made a little faster...
     for (i = 0; i < cnt; ++i) {
-        tmpEquiv_m[i] = m_eqr[rev];
-        tmpEquiv_n[i] = n_eqr[rev];
-        tmpScores[i] = score_pathr[rev];
+        tmpEquiv_m.push_back(m_eqr[rev]);
+        tmpEquiv_n.push_back(n_eqr[rev]);
+        tmpScores.push_back(score_pathr[rev]);
         rev--;
     }
     delete[] n_eqr;
@@ -1053,8 +1062,8 @@ float entropy(MatF &mat, int rowNum, int numBins, float minVal, float scaleFacto
 
 //Subtract a value
 void _subtract(MatF &mat, int rowNum, float val, MatF &minused) {
-    float *matptr = mat.pointer(rowNum);
-    float *minusedptr = minused.pointer(rowNum);
+    std::vector<float> matptr = mat.pointer(rowNum);
+    std::vector<float> minusedptr = minused.pointer(rowNum);
     for (int i = 0; i < mat.cols(); ++i) {
         minusedptr[i] = matptr[i] - val;
     }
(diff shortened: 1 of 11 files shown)


3rdparty/obiwarp/mat.h
```diff
@@ -22,458 +22,396 @@ class MatD;
...
+    int _n;
+    VecF _dat;
+    // Constructors:
+    MatF();
+    MatF(int m, int n);
+    MatF(int m, int n, const float &val);
+
+
+    // (copied from vec.h)
+    // if (shallow == 1 (true)) then no memory is deleted upon destruction
```
shubhra-agrawal: Would be great if the method docs could be cleaned up. `shallow` is not an arg anymore

shubhra-agrawal: Consider it a suggestion not a requirement

saifulbkhan: Oh I had it as a task in my head. Maybe it got lost in all the excitement. Thanks for reminding. Will make sure the docs do not have any incorrect information.

3rdparty/obiwarp/dynprog.cpp
```diff
@@ -1062,20 +1062,20 @@ float entropy(MatF &mat, int rowNum, int numBins, float minVal, float scaleFacto
 
 //Subtract a value
 void _subtract(MatF &mat, int rowNum, float val, MatF &minused) {
-    std::vector<float> matptr = mat.pointer(rowNum);
-    std::vector<float> minusedptr = minused.pointer(rowNum);
+    float* matPtr = mat.rowData(rowNum);
```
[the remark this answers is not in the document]
shubhra-agrawal: We can talk in person if you think that would be easier

3rdparty/obiwarp/vec.cpp
```diff
@@ -359,95 +277,63 @@ VecF VecF::operator+(const VecF &A) {
...
 }
 
 double VecF::avg() const {
-    double total = 0;
-    for( int n = 0; n < _n; ++n) {
-        total += _dat[n];
-    }
-    return total/_n;
+    double total = std::accumulate(_dat.begin(), _dat.end(), 0.0);
+    return total / static_cast<double>(_n);
```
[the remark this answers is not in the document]
saifulbkhan: Because I believe that when doing real number arithmetic, its always a good practice to make sure all values involved are (or are type-casted to be) of the same type (`double` in this case).

3rdparty/obiwarp/vec.cpp
```diff
@@ -1436,95 +1178,63 @@ VecD VecD::operator+(const VecD &A) {
...
 */
 
 double VecD::sum() {
-    double *me = (double*)(*this);
-    double sum = 0;
-    for( int n = 0; n < _n; n++) {
-        sum += me[n];
-    }
-    return sum;
+    return std::accumulate(_dat.begin(), _dat.end(), 0.0f);
```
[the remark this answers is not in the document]
saifulbkhan: Yes! Absolutely correct. In fact, in case of `std::accumulate` the third argument very much decides the kind of operations performed in intermediate steps as well the resulting value. Passing `0.0f` here makes us lose precision on each step of accumulation. Nice eye! Will change this.

3rdparty/obiwarp/vec.cpp
```diff
@@ -1568,37 +1277,34 @@ double VecD::_zScore(double mean, double sigma, double x) {
...
     }
     out.take(newcnt, tmparr);
 }
 
 void VecD::hist(int num_bins, VecD &bins, VecI &freqs) {
-    int i;
-
     // Create the scaling factor
-    double _min; double _max;
+    double _min = 0.0f;
```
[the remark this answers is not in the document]
saifulbkhan: Yep. Should have been a part of my "replace all"s that I had to do when creating these template classes. Please let me know if there are more like this (you should pay attention in `VecD`, `VecI`, `MatD` and `MatI` classes). I will make any corrections I spot on my end as well.

3rdparty/obiwarp/vec.cpp
```diff
@@ -1978,10 +1635,10 @@ double VecD::covariance(VecD &x, VecD &y) {
 }
 
 double VecD::euclidean(VecD &x, VecD &y) {
-    VecF diff(x.size());
+    VecD diff(x.size());
```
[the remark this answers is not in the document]
saifulbkhan: Will remove.

3rdparty/obiwarp/dynprog.cpp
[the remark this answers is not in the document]
saifulbkhan: The STL `vector` container provides a lot of nice methods but is in fact simply an abstraction over a contiguous block of elements (read "array"). Its [`data`](https://en.cppreference.com/w/cpp/container/vector/data) method allows the programmer to access the underlying array by returning a pointer to the first element. This is what `rowData` method for the matrix classes returns (and that is why their return types are pointers of `float`, `double` or `int` depending on the type of matrix).

The `rowIters` method does not return a pointer but iterators to the first and last element of a row in a matrix. Since a matrix is internally just a flat vector structure (with rows arranged adjacent to each other), you can directly access an element by using pointers or iterators, which also give you the guarantee that if you move ahead by 1 you will reach the next element in the row, and therefore, enable looping over the entire row (just make sure you stop at the end of the row by either checking for the size traversed or by checking if your iterator points to the last element at each step).

Does that make sense? I might have complicated this a bit so let me know if you want to talk about this in person.

3rdparty/obiwarp/vec.cpp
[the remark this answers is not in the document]
shubhra-agrawal: I saw a few more. Let me find those again.

3rdparty/obiwarp/dynprog.cpp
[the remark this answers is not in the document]
shubhra-agrawal: Makes perfect sense. Thanks!

3rdparty/obiwarp/mat.cpp
```diff
@@ -53,11 +53,26 @@ MatF::MatF(const MatF &A) : _m(A._m), _n(A._n), _dat(A._dat) {
...
+{
+    return _dat.slice2((m * _n), (m * _n) + _n);
+}
+
+float* MatF::rowData(int m)
+{
+    return (_dat.data() + (m * _n));
+}
+
+std::vector<float>::iterator MatF::rowIter(int m)
```
shubhra-agrawal: As far as I can see this is not being used. You can get the same result by `rowIters.first` anyway.

saifulbkhan: Oh its not? I was using it at one point. I'll remove it now.

3rdparty/obiwarp/mat.cpp
```diff
@@ -1,3 +1,15 @@
+#include <utility>
```
shubhra-agrawal: @saifulbkhan Please clean this up?

3rdparty/obiwarp/mat.cpp
```diff
@@ -80,11 +80,6 @@ float* MatF::rowData(int m)
     return (_dat.data() + (m * _n));
 }
 
-std::vector<float>::iterator MatF::rowIter(int m)
```
shubhra-agrawal: Obsolete for `MatI` and `MatD` too.

3rdparty/obiwarp/mat.cpp
saifulbkhan: I am very confused as to what this even is and how it got here. Removing.

3rdparty/obiwarp/mat.cpp
saifulbkhan: Removed already, I think.

PR #1122, opened 2019-07-29
[Fix] Find compound by ID when loading mzroll

shubhra-agrawal:
The .mzroll files have a compound name, id and db name attached to every
group. In case the name is not available, El-MAVEN should search for
compounds using the compound ID and DB, instead of trying to access
group's compound name

### DIFF
--- a/src/gui/mzroll/mzfileio.cpp
+++ b/src/gui/mzroll/mzfileio.cpp
@@ -1175,9 +1175,18 @@ PeakGroup* mzFileIO::readGroupXML(QXmlStreamReader& xml, PeakGroup* parent)
         if (matches.size() > 0)
             group->compound = matches[0];
     } else if (!compoundId.empty()) {
-        Compound* c = DB.findSpeciesByIdAndName(compoundId,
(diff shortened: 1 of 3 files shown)


src/gui/mzroll/mzfileio.cpp
```diff
@@ -1175,9 +1175,18 @@ PeakGroup* mzFileIO::readGroupXML(QXmlStreamReader& xml, PeakGroup* parent)
         if (matches.size() > 0)
             group->compound = matches[0];
     } else if (!compoundId.empty()) {
-        Compound* c = DB.findSpeciesByIdAndName(compoundId,
-                                                group->compound->name,
-                                                DB.ANYDATABASE);
+        Compound* c;
```
saifulbkhan: It is a good practice to initialize variables (`nullptr` in this case) to a default value, since we cannot rely on the compiler to assign it our expected default.

shubhra-agrawal: Updated.

PR #1124, opened 2019-07-30
New Isotopes dialog

shubhra-agrawal:
The "Report Isotopes" button was available in the Peaks dialog and other settings for labeled data lay in the Options dialog. The UI was not intuitive when it came to label detection. Especially for manual peak picking.

All label detection settings have been moved to a new Isotope dialog in the main toolbar for easy access.

Issue: #185


src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -227,29 +208,25 @@ void PeakDetectionDialog::closeEvent(QCloseEvent* event)
...
         mainwindow->alignmentDialog->peakDetectionAlgo->setCurrentIndex(1);
         featureOptions->setChecked(true);
-        reportIsotopesOptions->setEnabled(false);
     }
     
     QString dbName = compoundDatabase->currentText();
     toggleFragmentation(dbName);
 }
 
+bool PeakDetectionDialog::getDbOptions()
```
saifulbkhan: You don't have to make this change but the name of this method can be a bit better (suggestion: `isDatabaseSearchSelected`).

saifulbkhan: or maybe `isDatabaseSearchEnabled`?

src/gui/mzroll/isotopedialog.cpp
```diff
@@ -76,3 +88,8 @@ void IsotopeDialog::onReset()
 {
     emit resetSettings(isotopeSettings->getSettings().keys());
 }
+
+bool IsotopeDialog::getReportIsotopes()
```
saifulbkhan: This can also have a better name, such as - `isIsotopeReportingEnabled`.

shubhra-agrawal:
@saifulbkhan Made the required changes. Please take a look.

> remaining 4 isotope flags change from C13Labeled_BPE, ... etc., which was probably named like that because it was meant for the bar-plot widget
The barplot flags still exist, as the checkboxes are part of the barplot widget. The isotopeWidget version has been removed and merged with `C13Labeled_BPE` and left as is to avoid confusion with `C13Labeled_Barplot`

saifulbkhan:
@shubhra-agrawal Thanks. Changes look good. Go ahead and merge if you feel all the necessary work is done.

PR #1125, opened 2019-07-31
Allow substituting untargeted features with known targets

saifulbkhan:
This set of commits contain the following overall changes:
1. Allow selection of a compound database, in automated feature detection process, which will be used to replace features having an m/z value similar to a compound with a targeted group containing that compound.
2. Assign the name of a compound database to a peak table (instead of "Peak Table X"), created from a peak detection step, which involved the use of that compound database.
3. Transfer the role played by "Exclude unfragmented groups" checkbox to "Match Fragmentation" options.

### DIFF
--- a/src/core/libmaven/PeakDetector.cpp
+++ b/src/core/libmaven/PeakDetector.cpp
@@ -437,7 +437,6 @@ void PeakDetector::identifyFeatures(const vector<Compound*>& identificationSet)
         return;
 
     vector<PeakGroup> toBeMerged;
-    GroupFiltering groupFiltering(mavenParameters);
     auto iter = mavenParameters->allgroups.begin();
     while(iter != mavenParameters->allgroups.end()) {
         auto& group = *iter;
@@ -453,17 +452,9 @@ void PeakDetector::identifyFeatures(const vector<Compound*>& identificationSet)
             if (mzUtils::withinXMassCutoff(group.meanMz,
                                            mz,
                                            mavenParameters->massCutoffMerge)) {
+                matchFound = true;
                 PeakGroup groupWithTarget(group);
                 groupWithTarget.compound = compound;
-
-                // since we are creating targeted groups, we should ensure they
-                // pass MS2 filtering criteria, if enabled
-                if (mavenParameters->matchFragmentationFlag
-                    && groupFiltering.filterByMS2(groupWithTarget)) {
-                    continue;
-                }
-
-                matchFound = true;
                 toBeMerged.push_back(groupWithTarget);
             }
         }
--- a/src/core/libmaven/PeakDetector.cpp
+++ b/src/core/libmaven/PeakDetector.cpp
@@ -437,6 +437,7 @@ void PeakDetector::annotateFeatures(const vector<Compound*>& annotationSet)
         return;
 
     vector<PeakGroup> toBeMerged;
+    GroupFiltering groupFiltering(mavenParameters);
     auto iter = mavenParameters->allgroups.begin();
     while(iter != mavenParameters->allgroups.end()) {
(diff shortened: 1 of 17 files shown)


src/core/libmaven/PeakDetector.cpp
```diff
@@ -427,3 +430,58 @@ void PeakDetector::processSlices(vector<mzSlice *> &slices, string setName)
...
+        bool matchFound = false;
+        for (auto compound : identificationSet) {
+            float mz = 0.0f;
+            if (compound->formula.length()) {
+                int charge = mavenParameters->getCharge(compound);
+                mz = compound->adjustedMass(charge);
+            } else {
+                mz = compound->mass;
+            }
+            if (mzUtils::withinXMassCutoff(group.meanMz,
```
shubhra-agrawal: Since targeted detection is done on the basis of compound->mz +/- cutoff, I would suggest sending `mz` as the first parameter and `group.meanMz` as the second parameter.

saifulbkhan: Oh I had thought that `withinXMassCutoff` was checking proximity of two values independent of the order of the arguments. But, now that I see it, the check is whether the second mass lies in a small region around the first (in both directions). So, order does matter and I agree that query range should be created from compound's m/z. Making this change.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1963,6 +1965,35 @@ int TableDockWidget::getLabeledGroupCount()
...
 }
 
+QString TableDockWidget::getTitleForId(int tableId)
+{
+  return _idTitleMap.value(tableId, QString(""));
+}
+
+bool TableDockWidget::setTitleForId(int tableId, const QString& tableTitle)
+{
+  if (_idTitleMap.contains(tableId))
```
shubhra-agrawal: This is not being utilized during creation of the table.

src/projectDB/projectversioning.cpp
```diff
@@ -18,8 +18,7 @@ namespace ProjectVersioning {
 map<Version, int> appDbVersionMap = {
     {Version("0.6.0"), 0},
     {Version("0.7.0"), 1},
-    {Version("0.8.0"), 2},
```
shubhra-agrawal: the project versioning changes don't make sense to me. Why have you removed things from here?

saifulbkhan: This is very strange. I do not remember making any of these changes. But I did resolve conflicts in this portion (using VS Code), today while rebasing on develop. Let me make see if I can undo this somehow.

src/gui/mzroll/tabledockwidget.cpp
saifulbkhan: Can you clarify a little here? I don't understand what is not being used while creation of tables. The map?

shubhra-agrawal: This checkpoint. If the tableId already exists and the function returns false, how are we using that information?

saifulbkhan: Right, we are not using it in anyway in the constructors, because its unlikely that a newly created ID in the table will have an entry in the table ID map. However, a user can later call this method to check if it has been registered with a title. Let me know if I should remove this as well, if you don't see it being used anytime.

shubhra-agrawal: I feel like if we are adding a method for error handling, we might as well actually handle that condition. However unlikely it might be. Not a major issue though. Take a call.

saifulbkhan:
@shubhra-agrawal I've made the changes. Please take a look and make sure I'm not missing something.

PR #1131, opened 2019-08-02
Library management widget

saifulbkhan:
A new widget named "Library Manager" has been added which allows the user to keep track of all databases loaded in the past. This widget is primarily meant for users who deal with many different compound databases in a single session, where locating and loading every library in every session becomes a cumbersome task. Moreover, this widget will also be useful by making more than just "the last loaded database" available to the user.

New databases can be imported, known databases can be loaded (or reloaded) and deleted from the widget interface.


src/gui/mzroll/ligandwidget.cpp
```diff
@@ -61,17 +62,17 @@ LigandWidget::LigandWidget(MainWindow* mw) {
...
+  libraryButton = new QToolButton(toolBar);
+  libraryButton->setIcon(QIcon(rsrcPath + "/librarymanager.png"));
+  libraryButton->setToolTip("Open Library Manager");
+  connect(libraryButton,
+          &QPushButton::clicked,
+          _mw->getLibraryManager(),
+          &LibraryManager::exec);
 
   toolBar->addWidget(databaseSelect);
   toolBar->addWidget(loadButton);
```
shubhra-agrawal: @saifulbkhan We can remove the load button now that the library widget exists. It's in the same position as the previous button so shouldn't be hard to find.

saifulbkhan:
> Ligandwidget.cpp: 470 `SaveCompoundList()` is not being used anywhere now that the Save option has been removed

I'll remove it, in that case.

src/gui/mzroll/ligandwidget.h
```diff
@@ -103,8 +102,7 @@ private Q_SLOTS:
 
     QTreeWidget *treeWidget;
     QComboBox *databaseSelect;
-    QToolButton *saveButton;
-    QToolButton *loadButton;
+    QToolButton *libraryButton;
     QLineEdit*  filterEditor;
     QPoint dragStartPosition;
     QHash<Compound *, QTreeWidgetItem *> CompoundsHash;
```
shubhra-agrawal: `QHash<QString,bool>alteredDatabases;` is not being used anymore

src/gui/mzroll/point.cpp
```diff
@@ -291,7 +291,6 @@ void EicPoint::linkCompound() {
 
             //upadte ligand widget
```
shubhra-agrawal: dbname is not being used either

src/gui/mzroll/librarymanager.cpp
```diff
@@ -0,0 +1,333 @@
...
+            existingRecord.numberOfCompounds = record.numberOfCompounds;
+            entries.replace(i, existingRecord.toJson());
+            recordAlreadyExists = true;
+        }
+    }
+
+    if (!recordAlreadyExists)
+        entries.append(record.toJson());
+
+    emit saveFinished(_saveEntries(entries));
```
shubhra-agrawal: I don't see a connected slot for this. Why is this signal being emitted?

src/gui/mzroll/librarymanager.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    if (recordsIndicesToBeDeleted.isEmpty()) {
+        emit deleteFinished(false);
+        return;
+    }
+
+    for (auto index : recordsIndicesToBeDeleted)
+        entries.removeAt(index);
+
+    emit deleteFinished(_saveEntries(entries));
```
shubhra-agrawal: Same question as above for `deleteFinished` signal

src/gui/mzroll/librarymanager.cpp
saifulbkhan: Oh this is just for the sake of API completion and uniformity. In my head, if the worker emits a `loadFinished` signal, then it should also emit similar sibling signals. Maybe we'll find a use for them in the future.

src/gui/mzroll/librarymanager.cpp
saifulbkhan: Same reason as the one for `saveFinished` given above. Let me know if you think they're not needed at all.

src/gui/mzroll/librarymanager.cpp
shubhra-agrawal: Got it. Makes sense to have this.

PR #1134, opened 2019-08-05
Remove child from the children vector on delete

saifulbkhan:
The class `PeakGroup` offers a method to delete a child but this method was only clearing the space-intensive data structures for this child and not removing it from the vector that stores
children of a peak group. This has been fixed. As a result, a deleted child will no longer appear as a child of the peak group in exported CSV (and possibly other places where it might have been causing inconsistencies).

Issue: #261

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1105,10 +1105,33 @@ void TableDockWidget::deleteGroups() {
             if (parentGroup->deleteChild(group)) {
                 QTreeWidgetItem *parentItem = item->parent();
                 if (parentItem) {
-                	parentItem->removeChild(item);
-              		delete (item);
-            	}
-          	}
+                    parentItem->removeChild(item);
+                    delete (item);
+
+                    // once a child is deleted, the pointers storing the
+                    // location of memory blocks of child `PeakGroup` objects,
+                    // may no longer be valid, therefore we update them.
+                    for (int i = 0; i < parentItem->childCount(); ++i) {
+                        QTreeWidgetItem* child = parentItem->child(i);
+                        if (!child)
+                            continue;
+
+                        auto name = child->text(1).toStdString();
+                        auto childGroupIter =
+                            find_if(begin(parentGroup->children),
+                                    end(parentGroup->children),
+                                    [&](PeakGroup& g) {
+                                        return g.getName() == name;
+                                    });
+                        if (childGroupIter != end(parentGroup->children)) {
+                            auto& childGroup = *childGroupIter;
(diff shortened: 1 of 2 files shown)


shubhra-agrawal:
I know not how or why because the code looks fine to me, but if I delete labels 1 and 2 for O-Acetyl-L-serine from the peak table and export it, I get this in the output file:

deleteChild
Label 2 is still here. Label 3 is missing and Label 4 has 0 values.

Let me know if you want me to debug this. I'm really curious.

saifulbkhan:
This is very strange indeed. But, I tried to do the same thing on my system and failed to reproduce this bug:

Screenshot 2019-09-02 at 5 53 39 PM

Did you run into this issue with some particular setup or settings? If you have the time, please, see if you can find the cause. Otherwise, I can get hold of a Linux system and try to reproduce this behavior.

shubhra-agrawal:
@saifulbkhan look at the nan and 0 values for label 3 in your image. The values are all basically null even though you deleted some other labels

saifulbkhan:
Whoops! You're right. I was only looking for missing/extra labels. Let me find out what's going on.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1103,10 +1103,33 @@ void TableDockWidget::deleteGroups() {
...
-              		delete (item);
-            	}
-          	}
+                    parentItem->removeChild(item);
+                    delete (item);
+
+                    // once a child is deleted, the pointers storing the
+                    // location of memory blocks of child `PeakGroup` objects,
+                    // may no longer be valid, therefore we update them.
+                    for (int i = 0; i < parentItem->childCount(); ++i) {
```
shubhra-agrawal: @saifulbkhan might be simpler to clear all child items and start afresh. What do you think?

saifulbkhan: I had thought of this, but wasn't sure how that would impact user experience when an isotopic child was deleted from a parent having many other children. Deleting and repopulating 5+ child nodes might lead to blocky behavior. I could still give it a try.

saifulbkhan: I just tried out what you suggested. For a group having 9 isotopic children, deleting one made the UI freeze for 2 seconds while clearing & re-entering was in progress. So, can we stick to the current logic?

shubhra-agrawal: Current logic is fine. Was just trying to simplify this re:maintenance.

I'm only concerned about the behavior if user clicks on any of the children while the data is being updated. Will try and test for it as much as possible.

PR #1135, opened 2019-08-09
[UI] Remove Save option from ligand widget

shubhra-agrawal:
There was a 'Save' button on the ligand widget toolbar to save altered databases to a file.
Since the ligand widget is not editable, the feature is redundant.


shubhra-agrawal:
Closing this since it has already been handled in #1131

PR #1145, opened 2019-09-06
[Ftr] Enable adduct detection

saifulbkhan:
This set of commits will allow automatic identification of commonly known adducts. In short, the following changes have been made:
1. Add a separate dialog with a list of common adducts. Each entry comes with a checkbox to mark adducts that the user wishes El-MAVEN to detect.
2. Addition of a "Detect adducts…" button that will allow enabling/disabling adduct detection. A limiter for adduct's RT deviation has also been added. Once peak detection is complete, adducts are added to the peak table as regular peak groups.
3. Names of adducts attached to each peak group is exportable in emDB and CSV formats. emDB and mzrollDB files containing groups with adduct names can also be used to load these groups back with their adduct information retained.
4. The settings used to detect them are saveable in emDB sessions as well as persisted between sessions, until changed.


shubhra-agrawal:
@saifulbkhan Sorting is alphabetical instead of alphanumeric. Even for charge column I'm getting -1 before -3

saifulbkhan:
@shubhra-agrawal I've made some of the changes you requested. I'll raise a new PR to fix the issue in peak table naming. And, I think we've already resolved our confusion around negative mass.

Please go over the changes once, and let me know if you run into any more issues.

shubhra-agrawal:
@saifulbkhan Rebase please.

saifulbkhan:
@shubhra-agrawal Rebased. There were many conflicts (more so after merging the slicing branch). I compiled and was able to successfully get adducts from one run. Tests also seem to be passing on my system. But please be extra careful while reviewing this branch, there might still be some non-obvious regressions, that I was not able to resolve.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -214,52 +216,174 @@ void PeakDetector::processMassSlices(const vector<Compound*>& identificationSet)
...
+                                               [](const Scan* scan,
+                                                   const double &val) {
+                                                   return scan->precursorMz < val;
+                                               });
+                        for (auto scanIter = low;
+                             scanIter != scanVector.end();
+                             scanIter++) {
+                            Scan* scan = *scanIter;
+                            if (scan->precursorMz >= slice->mzmin
+                                && scan->precursorMz <= slice->mzmax) {
```
shubhra-agrawal: Let's not worry about reducing the number of adduct slices at this stage. Adduct groups don't need to have fragmentation information though even if the `fragmentationMatchFlag` is on. As long as the parent ion passes our thresholds, the adduct groups can be created regardless of fragmentation information.

src/core/libmaven/datastructures/mzSlice.h
```diff
@@ -107,6 +111,8 @@ class mzSlice
             rtmax = b.rtmax;
             ionCount = b.ionCount;
             compound = b.compound;
+            compoundVector = b.compoundVector;
```
shubhra-agrawal: Why the addition of a compound vector for every slice? When would it have more than one compound assigned to it?

shubhra-agrawal: As in, where is it utilized. Even if I have x compounds per formula.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -415,6 +504,12 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
                                      mavenParameters->limitGroupCount));
         }
     }
+
+    // finally, we would like to remove adducts for which no parent ions were
+    // detected or if parent ion's RT is too different from adduct's.
+    sendBoostSignal("Filtering out false adducts…", 0, 0);
+    GroupFiltering groupFilter(mavenParameters);
+    groupFilter.filterAdducts(mavenParameters->allgroups);
```
shubhra-agrawal: As discussed offline, please change the adduct detection flow to mimic how isotopes are pulled in El-MAVEN. It will reduce the amount of time spent detecting and then deleting adduct groups where the parent is not found or the RT does not match.

src/core/libmaven/datastructures/mzSlice.h
saifulbkhan: It is used in `PeakDetector::processSlices` method. Peak-grouping is attempted for each compound in the vector for each slice. Again, as discussed, we may instead do adduct detection like we do it for isotopes.

src/core/libmaven/PeakDetector.cpp
saifulbkhan: Done. I've kept the `filterAdducts` method. I think its still required. But adduct slices are now only created from any detected peak-groups.

src/core/libmaven/PeakDetector.cpp
saifulbkhan: > Done. I've kept the `filterAdducts` method. I think its still required. But adduct slices are now only created from any detected peak-groups.

Please ignore the original message. It was meant as a reply to this comment.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -912,3 +927,20 @@ void PeakGroup::setSelectedSamples(vector<mzSample*> vsamples){
...
+        && _adduct->getName() != MassCalculator::MinusHAdduct->getName()
+        && _adduct->getName() != MassCalculator::PlusHAdduct->getName()) {
+        _type = GroupType::Adduct;
+    }
+}
+
+Adduct* PeakGroup::getAdduct() const
+{
+    if (isIsotope())
+        return parent->getAdduct();
```
shubhra-agrawal: Wondering if we should even calculate adduct isotopes. What's the formula you're using for adduct groups?

saifulbkhan: I don't think we intended to calculate adduct isotopes, explicitly. But to answer your question, the original compound's formula is used for enumerating labels and then the parent-ion's mass (which is itself adjusted with adduct mass) is added with respective isotope masses.

Issue #1143, opened 2019-09-04
Logging to a File


cpj2195:
It would be great if the Polly-cli logs can be dumped to a text file and anyone can read it from it. Since Readin from the output stream is not the right way to go.

saifulbkhan:
The changes requested have landed as part of release v0.9.0. Closing.

PR #1146, opened 2019-09-09
Enable logging for CLI

saifulbkhan:
Logs are a very useful way of recording and persisting the output from an ongoing process. It is also very easy to read them and extract meaningful information for developers who might be interested in debugging or querying the state of the process at any point of time. Since, CLI is now being used by mutliple parties for different kinds of integration, logging the output and response from Polly APIs can help them at runtime to diagnose whether the process completed successfully or whether it failed and its point of failure in such a case.

Issue: #1143

### DIFF
--- a/src/cli/peakdetector/peakdetectorcli.cpp
+++ b/src/cli/peakdetector/peakdetectorcli.cpp
@@ -7,18 +7,9 @@
 #include "mzUtils.h"
 #include "peakdetectorcli.h"
 
-PeakDetectorCLI::PeakDetectorCLI()
+PeakDetectorCLI::PeakDetectorCLI(Logger* log)
 {
-    QString parentFolder = "ElMaven";
-    QString logFile = "peakdetector_cli.log";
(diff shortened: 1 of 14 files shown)


src/cli/peakdetector/peakdetectorcli.cpp
```diff
@@ -1,13 +1,24 @@
...
+#include "common/logger.h"
 #include "mavenparameters.h"
 #include "masscutofftype.h"
 #include "mzUtils.h"
 #include "peakdetectorcli.h"
 
 PeakDetectorCLI::PeakDetectorCLI()
 {
+    QString parentFolder = "ElMaven";
+    QString logFile = "peakdetector_cli.log";
```
shubhra-agrawal: The same `logFile` name is defined in main as well. Use that so it does not have to be maintained in both the files.

saifulbkhan: Made the change.

PR #1151, opened 2019-09-26
Fix last database load

saifulbkhan:
The current behavior of always loading the last compound database loaded (other than default ones), has been changed to load only the selected database of the last El-MAVEN session.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -1897,16 +1897,11 @@ void MainWindow::_postCompoundsDBLoadActions(QString filename,
             analytics->hitEvent("PRM", "LoadedSpectralLibrary");
         }
 
-        // do not save NIST library files for automatic load when starting next
-        // session, unless they are less than 2Mb in size
-        QFileInfo fileInfo(filename);
+        setLastLoadedDatabase(filename);
+
         bool isMSPFile = filename.endsWith("msp", Qt::CaseInsensitive);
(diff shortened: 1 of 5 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2030,6 +2025,17 @@ void MainWindow::_warnIfNISTPolarityMismatch()
...
 }
 
+void MainWindow::setLastLoadedDatabase(QString filename)
+{
+    QFileInfo fileInfo(filename);
+    bool isMSPFile = filename.endsWith("msp", Qt::CaseInsensitive);
+    bool isSPTXTFile = filename.endsWith("sptxt", Qt::CaseInsensitive);
+    bool nistFile = isMSPFile || isSPTXTFile;
+    bool smallerThan2Mb = fileInfo.size() < 2000000;
+    if (!nistFile || smallerThan2Mb)
```
shubhra-agrawal: @saifulbkhan Why are we only saving the last set library if it's not a spectral library?

saifulbkhan: The check is meant to make sure that if the file _is_ a spectral library, it has to be less than 2 Mb. The assumption here is that non-spectral databases are usually within the 2 Mb limit. Let me know if you think that should be changed.

shubhra-agrawal: @saifulbkhan My bad. I thought it was an 'and' condition

PR #1152, opened 2019-09-26
Fix regex pattern for matching table names

saifulbkhan:
The pattern matching for querying whether tables of a certain name already exist or not, was incorrect and has been fixed.


shubhra-agrawal:
@saifulbkhan Issue in loading emDB with multiple tables

Screenshot 2019-09-27 at 5 18 29 PM

saifulbkhan:
@shubhra-agrawal Will fix.

PR #1155, opened 2019-10-01
Enable reading Mascot Generic Format files

saifulbkhan:
Mascot Generic Format (MGF) files are often used to store spectral information for annotated metabolites, several of which are available on the [GNPS](https://gnps.ucsd.edu/ProteoSAFe/libraries.jsp) public database.

The [libmgf](http://kirchnerlab.github.io/libmgf/)[1] library has been used with a lot of custom modifications to support the GNPS flavor of spectral databases. For now, only minimal information is read from these databases but the support can be extended in the future if there is some need.

[1]: J Proteome Res. 2010 May 7;9(5):2762-3. [doi: 10.1021/pr100118f](https://doi.org/10.1021/pr100118f).


shubhra-agrawal:
Seems to be working fine.

PR #1156, opened 2019-10-07
Fix emDB large dataset crash

saifulbkhan:
Autosave threads for emDB can take a long time to save session when there is a large number of samples and peak-groups in each table. If the connection is forcibly closed from another thread (for e.g. when the user explicitly commands for a save), then pending statements (that exist in the autosave thread) will be attempted to destruct using instructions from a thread to which they do not belong. This was leading to definite crashes in case of large datasets. This set of commits aim to fix this problem and make some quality of life changes to emDB session saving/loading.


src/projectDB/projectdatabase.cpp
```diff
@@ -1454,9 +1456,9 @@ void ProjectDatabase::deletePeakGroup(PeakGroup* group)
         return;
 
     vector<int> selectedGroups;
-    selectedGroups.push_back(group->groupId);
+    selectedGroups.push_back(group->savedGroupId);
```
shubhra-agrawal: `groupId` and `savedGroupId` are the same, no? why add two variables for the same.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -903,18 +905,18 @@ AutoSave::AutoSave(MainWindow* mw)
...
     _mainwindow->timestampFileExists = false;
-    saveTablesOnly = false;
+    singleGroupToBeSaved = nullptr;
 }
 
-void AutoSave::saveProjectWorker(bool tablesOnly)
+void AutoSave::saveProjectWorker(PeakGroup *singleGroup)
 {
-    saveTablesOnly = tablesOnly;
+    singleGroupToBeSaved = singleGroup;
```
shubhra-agrawal: @saifulbkhan Turns out I can batch select a few groups and mark them as good or bad. What would the behavior be in that case?

src/projectDB/projectdatabase.cpp
saifulbkhan: I added `savedGroupId` to store the unique database ID associated with a peak-group that has been saved in an emDB project. I did not want to reuse `groupId` because it is enumerated differently and because we had to get to a working build without disturbing whatever other existing ways `groupId` is currently being used in various parts of the codebase. I agree that, in a restored session both of these values are the same and having only one unique ID associated to the object seems like good software construction. I will make the necessary changes, now that we have some time to refactor. Thanks for pointing out.

src/gui/mzroll/mainwindow.cpp
saifulbkhan: We can? I did not know about that. What would probably happen is that only one the groups would actually be saved with the g/b labels. But I'll make sure this is handled correctly.

saifulbkhan:
@shubhra-agrawal I've made the changes requested. Please give it another shot.

src/core/libmaven/PeakGroup.h
```diff
@@ -305,6 +305,8 @@ class PeakGroup{
 
         inline void addChildIsoWidget(const PeakGroup& child) { childrenIsoWidget.push_back(child); childrenIsoWidget.back().parent = this;   }
 
+        inline void setGroupIdForChildren() { for (auto& child : children) child.groupId = groupId; }
```
shubhra-agrawal: groupId is supposed to be different for every group. Even children. MetagroupId is the id shared between the parent and children

saifulbkhan: It was not unique until now. See here: https://github.com/ElucidataInc/ElMaven/blob/develop/src/core/libmaven/isotopeDetection.cpp#L319

Only while exporting are they assigned unque IDs and `metaGroupId` is shared. Otherwise `metaGroupId` is used as counters for the order in which a group's children were created.

Let me know if you want me to change this to the behavior you describe.

shubhra-agrawal: I'd forgotten. This is fine.

Issue #861, opened 2018-09-06
C12 PARENT not being detected for certain compounds


saifulbkhan:
This is a rare but reproducible bug (with certain datasets) where some compounds do not have a C12 PARENT subgroup after successful isotope detection.

**To Reproduce**
1. Load samples (this was reproducible for Masha dataset, sample [neg C12] 1).
2. Do peak detection with default settings and isotope reporting enabled
3. The generated peak table will have some groups that have no isotope children, often at the lower end of the peak table.

**Expected behavior**
Every group in a labelled workflow should at least have a C12 PARENT isotope entry as its child group. The absence of a C12 PARENT would mean that the compound was not present in the sample at all and should not have been populated into the peak table in the first.

**Screenshots**
![image](https://user-images.githubusercontent.com/10397162/45142807-14cf1200-b1d7-11e8-93f7-c86897720cd7.png)

**System**
 - OS: MacOS
 - Version: 0.4.1

**Additional context**
We tried looking into this issue but haven't been able to get to the root of it yet. For now, we know that the peak that should be tagged as C12 PARENT and has the closest resemblance to parent group (ideally it should be the same as parent group) is being filtered out during isotope detection.

chubukov:
@saifulbkhan I know it's probably causing all sorts of issues now, but I think it would be great if the underlying solution was to allow metagroups that have no parent peak. (see tangentially related discussion in #336)

shubhra-agrawal:
@saifulbkhan @chubukov Tracked down the issue to isotope-parent correlation. The EIC pulled for the parent was based on the group->meanMz instead of the compoundMz and therefore the correlation was not accurate leading to the C12 parent being filtered out at times. Raising a PR for this.

saifulbkhan:
Fixed and merged. Available in v0.9.0.

PR #1160, opened 2019-10-15
[Fix] Use correct slice for isotope-parent correlation #861

shubhra-agrawal:
Isotopic groups are filtered out in case the EIC does not have a strong correlation with the parent EIC.
Since the parentMass used in the correlation calculation was the groupMz instead of compoundMz, the correlation was incorrect in some cases and resulted in filtration of C12 parent groups

### DIFF
--- a/src/core/libmaven/isotopeDetection.cpp
+++ b/src/core/libmaven/isotopeDetection.cpp
@@ -205,6 +205,11 @@ bool IsotopeDetection::filterIsotope(Isotope x, float isotopePeakIntensity, floa
     //here w should really be determined by the minRt and maxRt for the parent and child peaks
     if (parentGroup)
     {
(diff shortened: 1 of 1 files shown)


src/core/libmaven/isotopeDetection.cpp
```diff
@@ -205,6 +205,11 @@ bool IsotopeDetection::filterIsotope(Isotope x, float isotopePeakIntensity, floa
     //here w should really be determined by the minRt and maxRt for the parent and child peaks
     if (parentGroup)
     {
+        float isotopeMass = x.mass;
+        Compound* compound = parentGroup->getSlice().compound;
```
saifulbkhan: Suggestion: using `parentGroup->getCompound()` here would give you the same object, and seems more succinct.

PR #1165, opened 2019-10-21
[Ftr] Notify users about ability to load zlib compressed data

saifulbkhan:
This patch adds a message box that informs the user (only on the first run of an El-MAVEN version), that they can now load sample files containing zlib compressed data.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -183,14 +183,6 @@ using namespace mzUtils;
 	QString dataDir = ".";
 	unloadableFiles.reserve(50);
 
-    string logPath = QString(QStandardPaths::writableLocation(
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -1587,6 +1587,17 @@ void MainWindow::analyticsAverageSpectra(){
 
 void MainWindow::open()
 {
+    // TODO: temporarily added for informing user, remove after a few releases
+    if (!_versionRecordExists()) {
+        QMessageBox msgBox;
+        msgBox.setText("El-MAVEN is now capable of reading files containing "
```
shubhra-agrawal: @surbhipoddar Here's the message we're displaying on the File loading page. Let us know if this is fine or if any modification is required.

PR #1166, opened 2019-10-21
Fix cli error reporting

shubhra-agrawal:
Error messages have been added to CLI in the following cases:
- Missing NodeJS folder
- Invalid credential file for Polly

CLI will display the corresponding error and exit.

### DIFF
--- a/src/cli/peakdetector/peakdetectorcli.cpp
+++ b/src/cli/peakdetector/peakdetectorcli.cpp
@@ -1067,6 +1067,12 @@ QString PeakDetectorCLI::uploadToPolly(QString jsPath,
     // set jspath and nodepath for _pollyIntegration library .
     _pollyIntegration->jsPath = jsPath;
     _pollyIntegration->nodePath = nodePath;
+    if (!_pollyIntegration->checkNodeExecutable()) {
+        _log->error() << "NodeJS was not found on this system. "
+                      "Please install NodeJS and try again."
+                      << std::flush;
+        exit(1);
+    }
 
     // In case of CLI, we don't want persistent login as of now, so deleting
     // existing Token everytime and starting afresh. In future if persistent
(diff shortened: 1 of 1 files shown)


src/cli/peakdetector/peakdetectorcli.cpp
```diff
@@ -785,6 +785,11 @@ void PeakDetectorCLI::writeReport(string setName,
 
         // try uploading to Polly
         QMap<QString, QString> creds = _readCredentialsFromXml(pollyArgs);
+        if (creds.empty()) {
+            _log->error() << "Error in credentials file. Exiting.." << std::flush;
```
saifulbkhan: "Error in credentials file" seems like a more general error. If the file is empty, lets inform the user of that situation itself.

src/cli/peakdetector/peakdetectorcli.cpp
```diff
@@ -1067,6 +1067,12 @@ QString PeakDetectorCLI::uploadToPolly(QString jsPath,
     // set jspath and nodepath for _pollyIntegration library .
     _pollyIntegration->jsPath = jsPath;
     _pollyIntegration->nodePath = nodePath;
+    if (!_pollyIntegration->checkNodeExecutable()) {
+        _log->error() << "NodeJS was not found on this system. "
+                      "Please install NodeJS and try again."
+                      << std::flush;
```
saifulbkhan: If the user installs NodeJS on their system using some package manager, they very likely will still not be able to use the CLI. Although this does hint to what the actual problem is, so we can leave this as is.

src/cli/peakdetector/peakdetectorcli.cpp
shubhra-agrawal: Specific errors in credentials file are handled in `_readCredentialsFromXml` and logged in the cli log file. This is a message to the user before CLI exits.

src/cli/peakdetector/peakdetectorcli.cpp
shubhra-agrawal: Which was the case with me too. Although the node_bin thing is a special case for Mac and I could handle that here. Let me see.

PR #1173, opened 2019-11-20
Changes over v0.9.0-beta

saifulbkhan:
These set of commits contains the following changes:
1. FIx Polly redirection button being available when switching between accounts, in the same session.
2. Fix crash due to comparison of compounds (where one of them has empty fragment vectors and one has elements in fragment vectors).
3. Make parsing of MGF files more fuzzy and fail less on unexpected inputs (auto-generated files from this commit were responsible for the large diffs). Only the files `Scanner.l` and `Parser.ypp` from libmgf need to be reviewed. 
4. Unload a compound database from compound widget as well as global compound storage when the DB entry is deleted from the Library Manager.
5. Move to Qt 5.12.5 on AppVeyor (latest LTS release) and Qt 5.12.3 on Travis (last LTS release available from homebrew).
6. Fix Polly application selection not updating (or partially updating) styled-selection within the navigation-list.


3rdparty/libmgf/mgf/MgfSpectrum.h
```diff
@@ -305,12 +305,12 @@ class MGFP_EXPORT MgfSpectrum : public Collection<MassAbundancePair>
     /** Get identifier to a PubMed study associated with this data.
      *  @return A string storing the PubMed identifier.
```
shubhra-agrawal: Please update the function doc as well

3rdparty/libmgf/src/Parser.cpp
```diff
@@ -828,222 +828,288 @@ namespace mgf {
   case 61:
 /* Line 664 of lalr1.cc  */
 #line 222 "Parser.ypp"
-    { driver.context.currentSpectrum.setIT_MODS(*(yysemantic_stack_[(4) - (3)].stringVal)); delete (yysemantic_stack_[(4) - (3)].stringVal); }
+    { driver.context.currentSpectrum.setINSTRUMENT(*(yysemantic_stack_[(3) - (2)].stringVal)); delete (yysemantic_stack_[(3) - (2)].stringVal); }
```
shubhra-agrawal: @saifulbkhan Could you explain what these numbers (3), (2) represent?

3rdparty/libmgf/mgf/MgfSpectrum.h
saifulbkhan: Oh missed this. Will do.

3rdparty/libmgf/src/Parser.cpp
```diff
@@ -828,222 +828,288 @@ namespace mgf {
...
 #line 238 "Parser.ypp"
-    { driver.context.currentSpectrum.setMSLEVEL((yysemantic_stack_[(4) - (3)].integerVal)); }
+    { driver.context.currentSpectrum.setCHARGE(*((yysemantic_stack_[(4) - (3)].integerlistnode)->integers_)); delete (yysemantic_stack_[(4) - (3)].integerlistnode); }
     break;
 
   case 78:
 /* Line 664 of lalr1.cc  */
 #line 239 "Parser.ypp"
-    { driver.context.currentSpectrum.setFILENAME(*(yysemantic_stack_[(4) - (3)].stringVal)); delete (yysemantic_stack_[(4) - (3)].stringVal); }
+    { driver.context.currentSpectrum.setMSLEVEL((yysemantic_stack_[(3) - (2)].integerVal)); }
```
shubhra-agrawal: This was the original statement. Is this an intentional change? I can't tell since I don't understand this part of the code.

`#line 238 "Parser.ypp"	#line 238 "Parser.ypp"
    { driver.context.currentSpectrum.setMSLEVEL((yysemantic_stack_[(4) - (3)].integerVal)); }`

3rdparty/libmgf/src/Parser.cpp
saifulbkhan: These `.cpp` files are auto-generated. You can instead look at this [highlighted diff](https://github.com/ElucidataInc/ElMaven/pull/1173/files/c1e1030b6825d2a372fb1061f36463e47dc87ca8#diff-64411394478601012964536bd244bd06R221-R222) and see that the numbers represent the tokens captured in order (kind of like regex groups).

3rdparty/libmgf/src/Parser.cpp
saifulbkhan: I can see its causing some confusion, but this whole file `Parser.cpp` is auto-generated. I should have mentioned that the only files you should check out are `Parser.ypp` and `Scanner.l`. The latter tokenizes our file stream and the former parses those tokens into meaningful objects.

3rdparty/libmgf/src/Parser.cpp
shubhra-agrawal: Got it.

PR #1174, opened 2019-11-27
Fix crash reports

saifulbkhan:
Crash reporting has been broken since El-MAVEN v0.8.0 was tagged. The main issue on Windows happened to be an incorrect format of endpoint file being deployed (UTF-16 when ASCII plain-text was expected).

It is very likely that a similar issue was responsible for reporting being broken on MacOS as well. But from the next release onward, we will attempt to outsource most of crash handling to a promising library - [sentry-native](https://github.com/getsentry/sentry-native) - which despite being in beta, handles a larger number of exceptions compared to our current Breakpad setup, and is more simpler to set-up (not to mention direct integration with Sentry). We are using the Crashpad back-end of sentry-native.

However, it might take a lot more more work to get this library integrated in our Windows builds, since sentry-native only supports MSVC builds on Windows, for now. To use it for El-MAVEN, would mean either compiling and making the binaries and libraries available for MinGW platform (which Crashpad does not support either), or, switching El-MAVEN itself to the MSVC ecosystem, which is an equally, if not more, time-consuming task. Therefore, we are sticking to our current system of crash-reporting. More efforts will be made in regards in later releases.

Some additional points worth mentioning:
1. There is no GUI prompt right now on MacOS asking the user's permission to report a crash. Dump-files are silently sent to Sentry. This is worth discussing about.
2. Release versions of crash-reports cannot be sent to Sentry through the MacOS reporting system due to a [bug](https://github.com/getsentry/sentry-native/issues/84). They seem to be working on it, and hopefully will be fixed before our next release comes in. But since this is going to be our only release with missing release information, we will know that its this one. :)
3. There is an extra commit that adds items to the GUI application's "Help" menu, making El-MAVEN crash due to various different exceptions. This might help when verifying the basic reporting mechanism. However, we will not get very detailed report from these intentional crash points. 
4. Most of these mock triggers (from point 3) do not even bring up the crash reporter on Windows, which probably means our handler is missing these signals.
5. The mock triggers have only been added for debugging and review process and _must_ not reach `develop`.

Even with these several limitations, I would suggest that we get this branch merged (after a proper review, of course), so that we have some basic crash-reporting which can be improved gradually. Much better than our users reporting crashes that never reach us.

Comments and suggestions are welcome.


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2650,6 +2652,88 @@ void MainWindow::createMenus() {
...
+    });
+
+    QAction* sigfpeAction = helpMenu->addAction("Raise SIGFPE");
+    connect(sigfpeAction, &QAction::triggered, []() {
+        ::raise(SIGFPE);
+    });
+
+    QAction* sigillAction = helpMenu->addAction("Raise SIGILL");
+    connect(sigillAction, &QAction::triggered, []() {
+        ::raise(SIGILL);
```
rish9511: If this is just for testing purposes then should we not write tests for this instead ?
Feels a bit odd to have it  in the main UI

rish9511:
Looks fine. Its more clean and easy to implement
Good stuff

> However, it might take a lot more more work to get this library integrated in our Windows builds, since sentry-native only supports MSVC builds on Windows

Sentry also did not support MinGW builds, we had to convert the symbols(PDB) to something that was compatible with Sentry(Breakpad symbols). I think it should be the same with `sentry-native`

src/gui/mzroll/mainwindow.cpp
saifulbkhan: Yes this is just temporary. I wanted to trigger exceptions and see how many of them were being captured on each platform. I left them here for the reviewer's test purposes as well. But this commit will be removed from the branch before it gets merged to develop.

saifulbkhan:
@rish9511 Right, Sentry itself isn't able to read PDBs simply extracted from MinGW executables (instead Breakpad's `dump_syms` gives us a .sym file which we are currently using).

But the senty-native problem is not that we cannot get symbols out. It is that we cannot link their `dll` with El-MAVEN, as it is. When compiled on Windows, their HTTP transport API looks for `winsock` headers (the native and only socket library available on Windows) which binds it strictly to the MSVC environment. To be able to compile sentry-native such that its linkable with El-MAVEN, we will need a Unix transport layer, which for now only Cygwin provides. From what I read, compiling things on Cygwin and then using it with MinGW builds gives you suboptimal results, even if you get it to work. Right now, that sounds like too much effort for something uncertain.

Maybe our solution lies in modifying the sentry-native library to use a [platform-independent networking library](https://www.boost.org/doc/libs/1_71_0/doc/html/boost_asio/overview/networking.html). Hopefully, I will get to do some more experiments in the upcoming release.

rish9511:
@saifulbkhan 
Alright, thanks for the  explanation
Good to go then 👍

src/gui/mzroll/main.cpp
```diff
@@ -27,11 +28,21 @@
 #include "controller.h"
 #include "elmavenlogger.h"
 
+#ifdef __OSX_AVAILABLE
```
shubhra-agrawal: What's the difference between the `__OSX_AVAILABLE` and `Q_OS_MAC` macro?

saifulbkhan: It's just that `__OSX_AVAILABLE` is independent of the Qt framework. This one is provided by the OSX native C++ headers themselves.

PR #1175, opened 2019-11-27
[Fix] Cohort file deleted before validation

saifulbkhan:
An EPI file-uploader thread handles upload as well as removal of temporary files created for pushing to Polly. This thread deletes files as soon as they are uploaded, which leads to a case where the redirection URL always being generated for GSheet, even if a vaid sample-cohort file had existed and was uploaded.

### DIFF
--- a/src/gui/mzroll/pollyelmaveninterface.cpp
+++ b/src/gui/mzroll/pollyelmaveninterface.cpp
@@ -821,10 +821,7 @@ QString PollyElmavenInterfaceDialog::_getRedirectionUrl(QString datetimestamp,
             return redirectionUrl;
 
         // send to google sheets if sample cohort file is not valid
-        QString CohortFileName = _writeableTempDir + QDir::separator()
-                                 + datetimestamp
-                                 + "_Cohort_Mapping_Elmaven.csv";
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -927,6 +924,7 @@ QStringList PollyElmavenInterfaceDialog::_prepareFilesToUpload(QDir qdir,
         QString sampleCohortFileName = _writeableTempDir + QDir::separator() + datetimestamp +
                                         "_Cohort_Mapping_Elmaven.csv";
         _mainwindow->projectDockWidget->prepareSampleCohortFile(sampleCohortFileName);
+        _lastCohortFileWasValid = _pollyIntegration->validSampleCohort(CohortFileName);
```
shubhra-agrawal: Could you reset this variable to `false` after every logout or every time an upload process is complete?

I can't think of any point where this might fail, since this is only used in PollyPhi, but suggesting this as a fail-safe.

saifulbkhan: Right, that makes sense. No flags from a previous session should be carried over to the next one.

PR #1177, opened 2019-12-04
Ftr mixpanel analytics

saifulbkhan:
GA works very well for tracking anonymous user behaviour but forbids any email-based usage tracking. Mixpanel is a similar service that allows storing additional user information which can help us those specific users better. That said, individual user identification will be entirely optional and up to the user to decide. If the user chooses not to identify themselves, a randomly generated UUID will be used, very similar to our current GA setup.

For now, we are only exploring Mixpanel as a service and a very small number of triggers have been added to a new "Support" dialog. If the results are good, we will completely move from GA to Mixpanel over the course of next few releases.


src/gui/mzroll/infodialog.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+        auto name = ui->nameEntry->text();
+        auto email = ui->emailEntry->text();
+        QRegularExpression emailPattern("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~"
+                                        "-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]"
+                                        "{0,61}[a-zA-Z0-9])?(?:\\.[a-zA"
+                                        "-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-"
+                                        "zA-Z0-9])?)*$");
+        auto match = emailPattern.match(email);
+        if (match.hasMatch()) {
```
shubhra-agrawal: If the pattern does not match, show a message saying "Add a valid email ID" or something.

src/common/mixpanel.cpp
```diff
@@ -8,19 +8,29 @@
...
+                                        + QDir::separator()
+                                        + "mixpanel.ini");
     _settings = new QSettings(settingsPath, QSettings::IniFormat);
     if (!_settings->contains("mixpanel-uuid")) {
-        _clientId = QUuid::createUuid().toString();
+        // use an existing Google Analytics ID if available
+        if (gaSettings->contains("analytics-cid")) {
+            _clientId = gaSettings->value("analytics-cid").toString();
+        } else {
+            _clientId = QUuid::createUuid().toString();
```
shubhra-agrawal: For the first ever user session, is the GA uuid created and written first or Mixpanel?

saifulbkhan: The `Analytics` object is created before `Mixpanel` object in `MainWindow` constructor. In fact `Mixpanel` and `InfoDialog` are the last children to be initialized.

src/gui/mzroll/infodialog.cpp
saifulbkhan: Added a label for informing user in three different scenarios (sign-in, sign-out and invalid email).

PR #1178, opened 2019-12-06
Beta fixes (v0.9.1-beta.1)

saifulbkhan:
The following issues have been fixed:
* Crash when bookmarking a group repeatedly.
* "Load" and "Delete" buttons being enabled by default, leading to a crash when library is empty.
* Crash when toggling mass tolerance switch with a group created by hovering over peaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,36 +148,55 @@ _To make sure the environment has ben setup correctly make sure the correct vers
 
 
 #### <a name="build"> </a> Build
-- `cd ElMaven`
-- For release builds : `qmake CONFIG+=release build.pro`
-- For debug builds : `qmake CONFIG+=debug build.pro`
-- `make -j4`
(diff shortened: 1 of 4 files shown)


shubhra-agrawal:
@saifulbkhan I have a minor UX issue with the Library Manager widget that could be resolved in this branch.

When I import a database, I see a very short-lived loader on the bottom right which usually does not come into my notice. There is no other indication that the DB has been loaded. I have to check the list to confirm if it's loaded. If the entry could be highlighted in the list, it might be easier? Or close the manager once the DB is loaded?

saifulbkhan:
@shubhra-agrawal Both import and load options will now close the `LibraryManager` after the load operation begins (so users will even be able to see their focused progress bar, just as it used to be without the manager).

Issue #1168, opened 2019-10-24
Filter formula string before upload


saifulbkhan:
El-MAVEN does not parse and filter the formula string for illegal characters (non-alphanumeric) before assigning it to a compound. The mass calculation method is also not good enough to handle certain cases (for e.g., alphabet followed by a special character followed by a numeric character). The CSV database file also generated for these compounds (with malformed formula) therefore also contains any special characters which can lead to downstream issues (most notably breaking Polly applications expecting El-MAVEN to send valid input for these cases).

**Expected behavior**
Before exporting data or passing it to third-party applications, El-MAVEN should do its job of ensuring that all files being exported obey certain rules of what is valid for each class of data.

**System (please complete the following information):**
 - OS: any
 - Version: v0.8.0

saifulbkhan:
Special characters are now (in v0.10.0) filtered out from formula strings during loading phase.

PR #1191, opened 2019-12-20
[Fix] Filter formula string for special characters #1168

saifulbkhan:
Only a subset of alpha-numeric characters are read from formula strings from any compound database. But internally the formula is still saved as the same string that was provided with the database. This, although it does not affect El-MAVEN much, can be problematic for downstream analysis where valid chemical formula (as in KEGG DB notation) may be expected. Therefore, with this commit, all formula strings will be stripped as soon as they are assigned to a compound, and exported to different formats/destinations as such.

### DIFF
--- a/src/core/libmaven/Compound.h
+++ b/src/core/libmaven/Compound.h
@@ -1,6 +1,7 @@
 #ifndef COMPOUND_H
 #define COMPOUND_H
 
+#include "constants.h"
 #include "standardincludes.h"
 #include "PeakGroup.h"
 
@@ -28,6 +29,11 @@ class Compound{
         *@param - _groupUnlinked will check  wether Compound and PeakGroup are linked or not
         */
         bool      _groupUnlinked;
+
+        /**
+         * Formula of the compound.
+         */
+        string _formula;
 
     public:
         enum class Type {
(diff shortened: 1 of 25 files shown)


src/core/libmaven/Compound.h
```diff
@@ -65,12 +71,52 @@ class Compound{
...
+         * @details It should be noted that this is only a filter and not a
+         * parser of any sort. This function can be used to strip a possible
+         * formula string to a form which is equivalent to how El-MAVEN
+         * currently understands a checmical formula. The parsing scheme in
+         * `MassCalculator::getComposition` can be used as a guide.
+         * @param formulaString Any arbitrary string.
+         * @return A string stripped of all non-chemical characters.
+         */
+        static string filterFormula(string formulaString)
+        {
```
shubhra-agrawal: @saifulbkhan How about stripping extra spaces within the same function? since that leads to mass calculation errors within El-MAVEN as well. Or do you want to take that up separately?

saifulbkhan: @shubhra-agrawal This function will remove any character from the `formulaString` that is not part of the combined string `CHE_FORMULA_ALPHA_UPP + CHE_FORMULA_ALPHA_LOW + CHE_FORMULA_COFF`. That includes spaces too, so, I am guessing we do not need to filter for them separately.

shubhra-agrawal: You're right. Resolving this.

shubhra-agrawal:
@saifulbkhan Rebase the branch and I'll approve this.

saifulbkhan:
@shubhra-agrawal Rebased on current develop. Please approve.

src/core/libmaven/jsonReports.cpp
```diff
@@ -67,8 +67,7 @@ void JSONReports::_writeCompoundLink(PeakGroup& grp, ofstream& filename)
     filename << "\"compoundId\": "<< _sanitizeJSONstring(compoundID) ;
     string compoundName = grp.getCompound()->name;
     filename << ",\n" << "\"compoundName\": "<< _sanitizeJSONstring(compoundName);
-    string formula = grp.getCompound()->formula;
-    filename << ",\n" << "\"formula\": "<< _sanitizeJSONstring(formula);
+    string formula = grp.getCompound()->formula();    filename << ",\n" << "\"formula\": "<< _sanitizeJSONstring(formula);
```
shubhra-agrawal: Please check

saifulbkhan: This is a harmless (as far as compilation is concerned) alternative to `\n`, and comes from Windows systems, they use carriage-return character for newlines. I will still get rid of it though. Thanks for pointing out.

shubhra-agrawal:
@saifulbkhan just saw Appveyor build is failing. Segmentation fault in MavenTests coming from the ref_mzUtils branch. Is someone working on this?

saifulbkhan:
This branch segfaulted itself. Twice. Can't be a coincidence. I will spin up a Windows instance and take a look.

shubhra-agrawal:
Checked AppVeyor builds. ftr_tests was also failing with the same error before getting merged.

saifulbkhan:
I ran MavenTests many times and it crashes occasionally at [`mzAligner.cpp:512`](https://github.com/saifulbkhan/ElMaven/blob/fix_filter_formula/src/core/libmaven/mzAligner.cpp#L512) - that too only on Windows. Since this crash is also reproducible on `develop`, I am going to assume this PR had nothing to do with it and merge it once CI passes this time. The fix for that crash can be made separately.

shubhra-agrawal:
Yes, this is not related to the PR. You can merge it.

PR #1194, opened 2019-12-24
[Fix] Allow ignoring close event

saifulbkhan:
When project saving feature was added, a prompt would appear asking the user to save their session before the exit. Once the "close" button of the main application was clicked, there was no way of cancelling the exit event (with or without saving). With this patch, the user will have the option to ignore the exit event by either selecting "Cancel" button or pressing "Esc" when then save prompt appears.


saifulbkhan:
> If I say 'yes' to saving the project and then click cancel instead of giving a filename, the app closes.

Fair point. Since, "Yes" is the default button and it might accidentally be pressed, the user should be able to cancel and return to their session. I'll make the change.

PR #1196, opened 2019-12-24
[Fix] Export labelled peaks in peak-detailed format

saifulbkhan:
Peaks of isotopic groups were not being exported in our peak-detailed format. Instead only the parent group's peaks were being written (and without any isotope label information). This has been fixed.

### DIFF
--- a/src/core/libmaven/csvreports.cpp
+++ b/src/core/libmaven/csvreports.cpp
@@ -205,8 +205,7 @@ void CSVReports::insertPeakInformationIntoCSVFile(PeakGroup* group)
     if (group->childCount() == 0) {
         writePeakInfo(group);
     } else {
-        for (auto& child : group->children)
-        writePeakInfo(&child);
+        insertIsotopes(group, true);
     }
 }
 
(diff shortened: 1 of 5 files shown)


saifulbkhan:
I will forward your request. I am hoping they will understand what the new Shiny app means from the context.

src/core/libmaven/csvreports.cpp
```diff
@@ -199,10 +200,14 @@ void CSVReports::addGroup (PeakGroup* group) {
...
-
-      writePeakInfo(group);
-
+void CSVReports::insertPeakInformationIntoCSVFile(PeakGroup* group)
+{
+    if (group->childCount() == 0) {
+        writePeakInfo(group);
+    } else {
+        for (auto& child : group->children)
+        writePeakInfo(&child);
```
shubhra-agrawal: @saifulbkhan When a group CSV is created, we check which tracers have been selected during the time of export and only those labels are added to the CSV. Please make the behavior consistent between group and peak CSVs.

saifulbkhan: I have made the change. Also the behavior you describe was disabled (`userSelectedIsotopesOnly` defaulted to false) always. I have enabled it. Let me know if this is not something that needs to happen right now.

shubhra-agrawal: @saifulbkhan I'm not sure if only user selected isotopes should be exported or all detected labels should go in the CSV. Since tracer selection usually happens before peak detection, I can't picture why the user would go and deselect one tracer before exporting. Maybe you should consult the product team before making this call?

tests/MavenTests/testCSVReports.cpp
```diff
@@ -396,57 +399,59 @@ void TestCSVReports::verifyUntargetedPeakReport(vector<mzSample*>& samplesToLoad
...
     QVERIFY(peakValues2[2] == "210.150269@16.714417");
     QVERIFY(peakValues2[3] == "");
-    // QVERIFY(peakValues2[4] == "testsample_3");
-    // QVERIFY(peakValues2[5] == "210.150055");
-    // QVERIFY(peakValues2[6] == "210.150101");
+    QVERIFY(peakValues2[4] == "");
+    // QVERIFY(peakValues2[5] == "testsample_3");
+    // QVERIFY(peakValues2[6] == "210.150055");
     // QVERIFY(peakValues2[7] == "210.150101");
-    // QVERIFY(peakValues2[8] == "16.719");
```
shubhra-agrawal: @saifulbkhan Now that we have the method for approx equality, maybe it's time to uncomment these test cases?

shubhra-agrawal: Doesn't have to be in this PR though.

saifulbkhan: We are going to scrap these tests in favor of new ones, hopefully very soon. Or at least, they'll be modified to a large extent. But until then, I will un-comment them. It is relevant to this PR, so I am doing it here itself.

shubhra-agrawal: @saifulbkhan I see this test block has been removed. Were these redundant?

saifulbkhan: Yes, that block was testing untargeted peak-groups that were isotope-labelled. This does not make sense and test cases were the same as the above unlabelled block.

PR #1198, opened 2020-01-10
[Ftr] Read scaling metadata

saifulbkhan:
A metadata (CSV) file is used in El-MAVEN to allow setting various parameters for sample list to be set. This CSV file can now also be used to supply the normalization (scaling) factor.

### DIFF
--- a/src/projectDB/projectdatabase.cpp
+++ b/src/projectDB/projectdatabase.cpp
@@ -113,7 +113,7 @@ void ProjectDatabase::saveSamples(const vector<mzSample*>& samples)
         samplesQuery->bind(":color_blue", s->color[2]);
         samplesQuery->bind(":color_alpha", s->color[3]);
 
-        samplesQuery->bind(":normal_const", s->getNormalizationConstant());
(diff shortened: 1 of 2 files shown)


src/projectDB/projectdatabase.cpp
```diff
@@ -113,7 +113,7 @@ void ProjectDatabase::saveSamples(const vector<mzSample*>& samples)
         samplesQuery->bind(":color_blue", s->color[2]);
         samplesQuery->bind(":color_alpha", s->color[3]);
 
-        samplesQuery->bind(":normal_const", s->getNormalizationConstant());
```
shubhra-agrawal: @saifulbkhan Is this intentional?

saifulbkhan: Yes. Unfortunately, our schema was initialized with `norml_const` as the column for storing the scaling factor, but the save method was using `normal_const` (which should have been the correct one). Making this change in the schema sounds like too much work for a spelling mistake, but is possible.

Issue #1171, opened 2019-11-06
Allow exporting metadata file template


shubhra-agrawal:
**Is your feature request related to a problem? Please describe.**
Manually entering cohort information in El-MAVEN is difficult, especially if I have 20+ samples. Creating a metadata file from scratch is also a lot of manual work since the sample filenames need to be entered in Excel.

**Describe the solution you'd like**
Once the user has imported their samples, allow exporting a metadatafile template with the correct file structure and all sample names filled in. Once the user has entered cohort, scaling or injection order information using Excel, El-MAVEN should read this file and assign the correct information to samples. 

Maybe there can be a sync button?? with a list of all exported templates in this session so the user can switch between different cohort definitions.

saifulbkhan:
Available with v0.10.0.

PR #1199, opened 2020-01-10
[Ftr] Export sample metadata file template #1171

saifulbkhan:
Editing metadata of samples within El-MAVEN, when the sample list is large quickly becomes tedious. For this reason we have metadata import facility. But generation of this metadata (CSV) file separately, is itself tedious for large number of samples. With this change users will be able to export a metadata template for their current set of loaded samples, which they can easily edit in a spreadsheet software and then use the import facility to get the desired effect.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -2234,10 +2234,20 @@ void MainWindow::loadCompoundsFile()
 
 // open function for set csv
 void MainWindow::loadMetaInformation() {
-	QStringList filelist =
-			QFileDialog::getOpenFileNames(this, "Select Set Information File To Load",
-					".",
-					"All Known Formats(*.csv *.tab *.tab.txt);;Tab Delimited(*.tab);;Tab Delimited Text(*.tab.txt);;CSV File(*.csv)");
+    auto loadedSamples = getSamples();
+    if (loadedSamples.empty())
+        return;
+    auto lastSample = loadedSamples.back();
+    auto sampleDir = QFileInfo(lastSample->fileName.c_str()).dir();
+
+    QStringList filelist =
+        QFileDialog::getOpenFileNames(this,
+                                      "Select Set Information File To Load",
+                                      sampleDir.path(),
+                                      "All Known Formats(*.csv *.tab *.tab.txt);;"
+                                      "Tab Delimited(*.tab);;"
+                                      "Tab Delimited Text(*.tab.txt);;"
+                                      "CSV File(*.csv)");
 
     if ( filelist.size() == 0 || filelist[0].isEmpty() ) return;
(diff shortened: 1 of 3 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2215,7 +2225,7 @@ int MainWindow::loadMetaCsvFile(string filename){
     int lineCount=0;
     map<string, int>header;
     vector<string> headers;
-    static const string allHeadersarr[] = {"sample", "set", "injection order"};
+    static const string allHeadersarr[] = {"sample", "set", "cohort," "scaling", "injection order"};
```
shubhra-agrawal: Typo. Comma is inside the quotes.

saifulbkhan: Corrected (locally). Thanks for pointing out.

shubhra-agrawal:
Tried exporting the template and using that as the metadata file. Since the injection orders are 0 for every sample in the template, El-MAVEN gives an error message on loading the metadata file for corrupted injection order values for all samples.

Tried giving a unique injection order to every sample and the error disappeared.


Would also suggest adding an information dialog after the template has been exported, informing user that the file is at x location and can be edited and imported back into El-MAVEN.

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -152,9 +181,15 @@ void ProjectDockWidget::prepareSampleCohortFile(QString sampleCohortFileName) {
     vector<mzSample*> loadedSamples = _mainwindow->getSamples();
 
 	QTextStream out(&file);
-	out << "Sample" << ",Cohort" << "\n";
+    out << "Sample" << ","
```
shubhra-agrawal: @saifulbkhan Please ensure this branch is tested with PollyPhi before merging

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -186,10 +194,14 @@ void ProjectDockWidget::prepareSampleCohortFile(QString sampleCohortFileName) {
         << "Scaling" << ","
         << "Injection Order" << "\n";
 	for (const auto& sample : loadedSamples) {
+        QString injectionOrder = "";
+        if (sample->getInjectionOrder() > 0)
+            QString::number(sample->getInjectionOrder());
```
shubhra-agrawal: the variable `injectionOrder` is not getting updated

saifulbkhan:
@shubhra-agrawal Resolved conflicts with develop and made the changes that you had suggested.

shubhra-agrawal:
@saifulbkhan great! Let me know once Phi is tested.

saifulbkhan:
@shubhra-agrawal I should have mentioned, the cohort files being generated are working with PollyPhi (tested using demo dataset using KNOWNS).

PR #1201, opened 2020-01-10
Improve mass slicing

saifulbkhan:
Several small improvements, fixes and some merge steps have been added to the active mass slicing method used when performing untargeted detection. More information can be found in the commit log.

To-do before merge:
- [x] Discard code that prints slice information to a file.
- [x] Implement RT matching as described in  #1203 and #1204.
- [x] Fix unselected samples being sliced and detected for.


.travis.yml
```diff
@@ -98,13 +98,16 @@ script:
       gcc -v;
 
       if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-        export QTEST_FUNCTION_TIMEOUT=1200000
```
shubhra-agrawal: @saifulbkhan there's a 30 min lag in test because of mass slicing changes?

run.sh
```diff
@@ -24,13 +24,3 @@ if [[ "$OSTYPE" == "darwin"* && flag -eq 100 ]]; then
 		cp $SENTRY_MACOSX_BIN/crashpad_handler bin/El-MAVEN.app/Contents/MacOS/;
 	fi
 fi
-
-if [ -f tests/MavenTests/test.xml ]; then
```
shubhra-agrawal: @saifulbkhan Why has this been removed?

shubhra-agrawal: oh there's a separate command for running tests. Nevermind.

src/core/libmaven/Compound.cpp
```diff
@@ -76,7 +77,12 @@ float Compound::adjustedMass(int charge) {
     *@return    -    total mass by formula minus loss of electrons' mass 
     *@see  -  double MassCalculator::computeMass(string formula, int charge) in mzMassCalculator.cpp
     */
-    return MassCalculator::computeMass(formula,charge);
+     if (!formula.empty()) {
+        return MassCalculator::computeMass(formula,charge);
+     } else if (neutralMass != 0.0f) {
+         return MassCalculator::adjustMass(neutralMass, charge);
+     }
+     return mass;
```
shubhra-agrawal: This has not been initialized, no?

src/core/libmaven/PeakDetector.cpp
```diff
@@ -304,77 +304,78 @@ void PeakDetector::alignSamples(const int& method) {
...
             mavenParameters->clsf->scoreEICs(eics);
+        auto highestIntensity = 0.0f;
+        auto mzAtHighestIntensity = 0.0f;
+        auto rtAtHighestIntensity = 0.0f;
+
+        for (auto eic: eics) {
+            for (size_t i = 0; i < eic->size(); ++i) {
+                auto intensityAtIdx = eic->intensity[i];
+                if (intensityAtIdx > highestIntensity) {
+                    highestIntensity = intensityAtIdx;
```
shubhra-agrawal: @saifulbkhan are these values being used somewhere? can't find it.

.travis.yml
saifulbkhan: Our new slicing method is _that_ much slower.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -463,6 +463,14 @@ void PeakDetector::identifyFeatures(const vector<Compound*>& identificationSet)
                     continue;
                 }
 
+                // we should filter the annotated group based on its RT, if the
```
shubhra-agrawal: `filterByMS2` is computationally heavier than the RT check. Maybe switch them around? just a suggestion.

src/core/libmaven/PeakDetector.cpp
saifulbkhan: Sorry, these were left behind. Was using them to export slices as CSV. Will get rid of it.

src/core/libmaven/PeakDetector.cpp
saifulbkhan: Good suggestion. Changed locally.

src/gui/mzroll/database.cpp
```diff
@@ -850,12 +853,20 @@ int Database::loadCompoundCSVFile(string filename){
...
             compound->expectedRt = rt;
 
-            if (mz == 0) mz = MassCalculator::computeMass(formula,charge);
+            if (mass != 0.0f)
+                compound->neutralMass = mass;
+
+            if (mz == 0 && !formula.empty()) {
+                mz = MassCalculator::computeMass(formula,charge);
+            } else if (mass != 0.0f) {
+                mz = MassCalculator::adjustMass(mass, 1);
```
shubhra-agrawal: Why not send `charge` instead of 1 here?

.travis.yml
shubhra-agrawal: As discussed offline, let's restrict the RT range for slicing to reduce the time or use other sample files.

src/gui/mzroll/database.cpp
saifulbkhan: Will change.

src/projectDB/projectdatabase.cpp
```diff
@@ -199,7 +199,7 @@ int ProjectDatabase::saveGroupAndPeaks(PeakGroup* group,
     groupsQuery->bind(":meta_group_id", group->metaGroupId);
     groupsQuery->bind(":tag_string", group->tagString);
     groupsQuery->bind(":expected_mz", group->expectedMz);
-    groupsQuery->bind(":expected_rt_diff", group->expectedRtDiff);
+    groupsQuery->bind(":expected_rt_diff", group->expectedRtDiff()); // do we need this anymore?
```
shubhra-agrawal: It's best to leave it in the emDB so users can query this directly.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -109,99 +115,184 @@ void MassSlices::algorithmB( MassCutoff *massCutoff,int rtStep ) {
...
-        else if(i==2) num = "rd";
-        else num = "th ";
+        if (i==0) {
+            num = "st";
+        } else if (i==1) {
+            num = "nd";
+        } else if (i==2) {
+            num = "rd";
+        } else {
+            num = "th";
```
shubhra-agrawal: i = 20 will be 21th :P

Sounds like just writing "slice number i+1" will be easier. No need to correct it, just nitpicking here.

saifulbkhan: Ya that sounds better. I thought about doing some modulus math to get the right suffix but then 11 to 19 would need special treatment. Instead I will change the message to just display numbers.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -109,99 +115,184 @@ void MassSlices::algorithmB( MassCutoff *massCutoff,int rtStep ) {
...
+         end(slices),
+         [](const mzSlice* slice, const mzSlice* compSlice) {
+             if (slice->mz == compSlice->mz) {
+                 return slice->rt < compSlice->rt;
+             }
+             return slice->mz < compSlice->mz;
+         });
+
+    // reduction: merge slices sharing significant ROI
+    for (auto first = begin(slices); first != end(slices); ++first) {
```
shubhra-agrawal: Would suggest creating a separate method for getting the vector of all slices. Just for cleanliness' sake.

saifulbkhan: Alright. This one function has gotten too large anyway.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -237,27 +257,346 @@ void MassSlices::algorithmC(float ppm, float minIntensity, float rtWindow) {
     cerr << "#algorithmC" << slices.size() << endl;
 }
 
-//Function to check if slice is already present in cache
-mzSlice*  MassSlices::sliceExists(float mz, float rt) {
-    pair< multimap<int, mzSlice*>::iterator,  multimap<int, mzSlice*>::iterator > ppp;
-    // putting all mz slices in cache in a particular range in ppp
-    ppp = cache.equal_range( (int) (mz* 10) );
-    multimap<int, mzSlice*>::iterator it2 = ppp.first;
+mzSlice* MassSlices::sliceExists(float mzMinBound,
```
shubhra-agrawal: So this method is not being used in algorithmB, right?

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -237,27 +257,346 @@ void MassSlices::algorithmC(float ppm, float minIntensity, float rtWindow) {
...
+            }
+        }
+        sendSignal("Reducing slices…", first - begin(slices), slices.size());
+    }
+}
+
+void MassSlices::mergeNeighbouringSlices(MassCutoff* massCutoff,
+                                         float rtTolerance)
+{
+    auto slicesInProximity = [&](vector<mzSample*>& samples,
```
shubhra-agrawal: The whole method is in a lambda function. From what I understand, it's supposed to be used for small blocks of code. It's hurting readability here.

src/core/libmaven/mzMassSlicer.cpp
saifulbkhan: Not anymore, no.

src/core/libmaven/mzMassSlicer.cpp
saifulbkhan: I was not aiming for prettiness. While prototyping, we had multiple merge steps, and all of them had a common pattern of iteration and conditionally merging. The iteration part was common and therefore I wrote `_mergeSlices` method for that part. The merging logic was different for each step and that is why the different lambdas were sent to `_mergeSlices` to achieve different kinds of merges. But you are right. Now that we have only one merge step, I should get rid of all this quirky code.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -366,120 +366,14 @@ void MassSlices::_reduceSlices()
                 secondSlice->ionCount = -1.0f;
             }
         }
-        sendSignal("Reducing slices…", first - begin(slices), slices.size());
+        sendSignal("Reducing redundant slices…",
```
shubhra-agrawal: I would suggest moving the code for deleting slices here if possible. Right now if I go through algorithmB, it looks like there is one call for reducing slices, then another loop for removing slices, then a call for removing duplicate slices and so on. Maybe we can also merge `reduceSlices` and `removeDuplicateSlices` while we're at it? I've not seen the code for both methods in detail so I'll let you take that call.

shubhra-agrawal: or just call `removeDuplicateSlices` from within `reduceSlices`.

saifulbkhan: Right. The deletion part definitely belongs here.

But just like you, I am not familiar with the `removeDuplicateSlices` method, and I suspect its largely useless. We had agreed to get rid of it some time ago. I will do that now.

shubhra-agrawal: I remember we had a conversation about the utility of `removeDuplicateSlices` while testing, but not whether it would make a difference in the output. I too suspect it shouldn't affect the output but worth a sanity check, no? Do we have saved output from a previous version of this branch?

saifulbkhan: Not at the moment. But we did check while deciding. This method would, at best, lead to reduction of 1 or 2 slices among several thousands.

shubhra-agrawal: Cool, then. Resolving.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -237,89 +245,355 @@ void MassSlices::algorithmC(float ppm, float minIntensity, float rtWindow) {
...
+        auto comparisonEic = sample->getEIC(comparisonMzMin,
+                                            comparisonMzMax,
+                                            comparisonRtMin,
+                                            comparisonRtMax,
+                                            1,
+                                            1,
+                                            "");
+
+        // obtain the highest intensity's mz and rt in the EICs
+        highestIntensity = eic->maxIntensity;
```
shubhra-agrawal: This will always store the attributes for the last sample, right?

saifulbkhan: Right. This is a bug.

src/core/libmaven/mzMassSlicer.cpp
```diff
@@ -516,32 +516,55 @@ pair<bool, bool> MassSlices::_compareSlices(vector<mzSample*>& samples,
...
-        // obtain the highest intensity's mz and rt in the EICs
-        highestIntensity = eic->maxIntensity;
-        rtAtHighestIntensity = eic->rtAtMaxIntensity;
-        mzAtHighestIntensity = eic->mzAtMaxIntensity;
-        highestCompIntensity = comparisonEic->maxIntensity;
-        rtAtHighestCompIntensity = comparisonEic->rtAtMaxIntensity;
-        mzAtHighestCompIntensity = comparisonEic->mzAtMaxIntensity;
-        delete eic;
-        delete comparisonEic;
+#pragma omp parallel
```
shubhra-agrawal: Good thinking with multiprocessing this block! Should help with the run time now.

PR #1202, opened 2020-01-15
[Ref] Refactor jsonReports.cpp and tests done using doctest

sakshikukreja14:
"jsonReports.cpp" has been refactored along with the refactoring of "jsonReports.h". The name of some functions earlier being used have been changed to make it simpler to use. Secondly, the testing of the same file has been done using doctest libarary. This library can be used to test libmaven code. The directory "3rdparty/doctest" has been added that contains the needed header files. Another directory "tests/doctest" has been added that passes the json files to which the cpp file has been tested.


saifulbkhan:
@sakshikukreja14 Let's move `3rdParty/doctest/json.hpp` to `3rdParty/json/json.hpp`.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
-string JSONReports::sanitizeJSONstring(string s) {
-    boost::replace_all(s, "\"","\\\"");
-    //replace(s.begin(),s.end(),"\"","\\\""); //escape quotes
-    s="\""+s+"\""; //quote the whole string
-    return s;
-}
+#include "doctest.h"
+#include "json.hpp"
+
+using json = nlohmann::json;
```
saifulbkhan: Let us cluster includes before any namespace declarations. And in general, follow the order:
1. Defines (#define) go first, then includes (#include) and then other preprocessor macros.
2. Namespace, constants, forward declarations, etc. will follow preprocessor directives.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+#include "mzSample.h"
+#include "mzUtils.h"
+#include "mavenparameters.h"
+#include "datastructures/mzSlice.h"
+#include "databases.h"
+#include "classifierNeuralNet.h"
+#include "PeakDetector.h"
+
+
+JSONReports::JSONReports(MavenParameters* _mp, bool pollyUpload):
```
saifulbkhan: Nitpick: rename `_mp` to `mp` (or better yet `parameters`). We reserve the underscore prefix for private attributes and methods.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+    _uploadToPolly(pollyUpload)
+{
+    _mavenParameters=_mp;
```
saifulbkhan: Let us assign `_mavenParameters` in the constructor's initializer list itself.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+}
+
+
+void JSONReports::_writeGroup(PeakGroup& grp, ofstream& myfile) {
+
```
saifulbkhan: [Style guide violation](https://github.com/ElucidataInc/ElMaven/blob/develop/STYLE_GUIDE.md#braces).

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+    _uploadToPolly(pollyUpload)
+{
+    _mavenParameters=_mp;
+}
+
+
+void JSONReports::_writeGroup(PeakGroup& grp, ofstream& myfile) {
+    
+    //add labels to json file
+    char lab = grp.label;
```
saifulbkhan: Please name variables to be more expressive. `lab` can be confusing, when `label` is just two more characters.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+    myfile << ",\n" << "\"rtmin\": " << grp.minRt ;
+    myfile << ",\n" << "\"rtmax\": " << grp.maxRt ;
+    myfile << ",\n" << "\"maxQuality\": " << grp.maxQuality ;
+}
+
+
+void JSONReports::_writeCompoundLink(PeakGroup& grp,ofstream& myfile){
+    
+    myfile << setprecision(10);
+    double mz;
```
saifulbkhan: It is a good practice to _always_ initialize any declared variables. That would make this:
```C++
double mz = 0.0f;
```

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+}
+
+
+void JSONReports::_writePeak(PeakGroup& grp, ofstream& myfile, vector<mzSample*> vsamples){
+
+    myfile << setprecision(10);
+    
+    myfile << ",\n"<< "\"peaks\": [ " ;
+
+    for(std::vector<mzSample*>::iterator it = vsamples.begin(); it != vsamples.end(); ++it) {
```
saifulbkhan: Just a suggestion, when dealing with complicated types (especially when there is templates involved), you can use the `auto` keyword instead to save yourself the energy of figuring out what type the rvalue would be.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+        if (it!=vsamples.begin()) {
```
saifulbkhan: Please add, spaces between operators and comparators. You seem to ignore spaces a bit too much.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+    
+    int charge = _mavenParameters->getCharge(grp.getCompound());
+    slice->mz = grp.getExpectedMz(charge);
+
+    if (slice->mz == -1) {
+        slice->mz = grp.meanMz;
+    }
+    
+    
+    EIC* eic=NULL;
```
saifulbkhan: Avoid using `NULL` for initializing pointers. Use `nullptr` instead. It was introduced especially for this use case. `NULL` is also a C construct. Read a short note [here](http://www.stroustrup.com/bs_faq2.html#null).

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+        _writeEIC(grp,myfile,*it);
+    }
+    
+    myfile << "\n]" ; //peaks
+    
+    myfile << "}" ; //group
+}
+
+
+void JSONReports::_writeEIC(PeakGroup& grp, ofstream& myfile, mzSample* it){
```
saifulbkhan: Name pointer to an `mzSample` as `sample` instead of `it`, which makes it look like an iterator.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+void JSONReports::save(string filename,vector<PeakGroup> allgroups,vector<mzSample*> samples) {
+    ofstream myfile(filename.c_str());
+    myfile << setprecision(10);
+
+    myfile << "{\"groups\": [" <<endl;
+
+    int groupId=0;
+    int metaGroupId=0;
+   
+    for(int i=0; i < static_cast<int>(allgroups.size()) ; i++ ) {
```
saifulbkhan: Instead of casting vector size to be an integer, you can init your index (`i`), as a size type (`size_t`).

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+        PeakGroup& grp = allgroups[i];
+
+        //if compound is unknown, output only the unlabeled form information
+        if( grp.getCompound() == NULL || grp.childCount() == 0 ) {
+            grp.groupId= ++groupId;
+            grp.metaGroupId= ++metaGroupId;
+            if(groupId>1) myfile << "\n,";
+
+            _writeGroup(grp,myfile);  
+            if( grp.hasCompoundLink() )
```
saifulbkhan: Add spaces after `if` and not inside the brackets.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+                _writeCompoundLink(grp,myfile);
+            _writePeak(grp,myfile,samples);
+            //writeEIC(grp,myfile,samples);
+        }
+        else { //output all relevant isotope info otherwise
+            //does this work? is children[0] always the same as grp (parent)?
```
saifulbkhan: Try to contain comments on a separate lines of their own.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,678 @@
...
+                if(groupId>1) myfile << "\n,";
+            
+                _writeGroup(grp.children[k],myfile);
+                if( grp.children[k].hasCompoundLink() )
+                    _writeCompoundLink(grp,myfile);
+                _writePeak(grp.children[k],myfile,samples);
+                //writeEIC(grp.children[k],myfile,samples);
+            }
+        }
+        //Q_EMIT(updateProgressBar("Writing to json file. Please wait...", i, allgroups.size() - 1));
```
saifulbkhan: If you see commented code that is no longer of any obvious use, just remove it.

src/core/libmaven/jsonReports.h
```diff
@@ -1,32 +1,37 @@
...
+    void _writeGroup(PeakGroup& grp,ofstream& myfile);
+    void _writePeak(PeakGroup& grp,ofstream& myfile, vector<mzSample*> vsampleNames);
+    void _writeEIC(PeakGroup& grp,ofstream& myfile, mzSample*);
+    void _writeCompoundLink(PeakGroup& grp,ofstream& myfile);
+    string _sanitizeJSONstring(string s);
+    
+    float _outputRtWindow = 2.0;
+    bool _uploadToPolly;
+    MavenParameters* _mavenParameters;
+};
```
saifulbkhan: Where is the documentation? One of the items on your checklist while refactoring should be add appropriate documentation for the parts you have changed.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,674 @@
...
+                if( grp.children[k].hasCompoundLink() )
+                    _writeCompoundLink(grp, myfile);
+                _writePeak(grp.children[k], myfile, samples);
+            }
+        }
+    }
+    myfile << "]}"; //groups
+    myfile.close();
+}
+
```
saifulbkhan: Can we put a separator here to signify that the following code is part of the tests? My suggestion:

```C++
////////////////////////////////////////////////////////////////////////////////
```

And let's do this in all the files we are writing tests for.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,674 @@
...
+    return s;
+}
+
+
+class JsonReportsFixture{
+
+    protected:
+            vector<mzSample*> samples;
+            vector <PeakGroup> allgroups;
+            MavenParameters* mavenparameters;
```
saifulbkhan: I am curious why you made these variables protected and not private.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,674 @@
...
+    private:
+            Databases database;
```
saifulbkhan: Private members start with an underscore. So `_database`.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,674 @@
...
+            for(size_t l = 0 ; l < rootInput["groups"][input]["peak"][i]["eic"]["rt"].size() ; l++)
+                REQUIRE(rootInput["groups"][input]["peak"][i]["eic"]["rt"][l] ==
+                                                                rootSaved["groups"][saved]["peak"][i]["eic"]["rt"][l]);
+                                                                                              
+            for(size_t l = 0; i < rootInput["groups"][input]["peak"][i]["eic"]["intensity"].size() ; l++)
+                REQUIRE(rootInput["groups"][input]["peak"][i]["eic"]["intensity"][l] == 
+                                                                rootSaved["groups"][saved]["peak"][i]["eic"]["intensity"][l]);
+                                
+        }
+    }
```
saifulbkhan: I think, at this point, you should delete the "test.json" file that was created in the beginning of this function.

src/core/libmaven/jsonReports.cpp
```diff
@@ -1,286 +1,674 @@
...
+        for(size_t i = 0 ; i < rootInput["groups"][input]["peak"].size(); i++ ){
+                                    
+            string SNameInput = "";
+            string SNameSaved = "";
+
+            REQUIRE( !rootInput["groups"][input]["peak"][i]["sampleName"].is_null() );
+            SNameInput = rootInput["groups"][input]["peak"][i]["sampleName"].get<string>();
+                                    
+            REQUIRE( !rootInput["groups"][input]["peak"][i]["sampleName"].is_null() );
+            SNameSaved = rootSaved["groups"][saved]["peak"][i]["sampleName"].get<string>();
```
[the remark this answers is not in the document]
sakshikukreja14: There were some cases where sample names were empty values and the "REQUIRE" statement in this case was giving a segment fault. Thus, they have been checked for null values.

src/core/libmaven/jsonReports.cpp
sakshikukreja14: Making them private in fixture would limit their access in the test_case. Because we needed maven parameters and other variables to make json object and  call the corresponding function. It was a necessity to make them protected and not private.

src/core/libmaven/libmaven.pro
```diff
@@ -100,7 +107,8 @@ SOURCES = 	base64.cpp \
                 groupClassifier.cpp \
                 groupFeatures.cpp \
                 svmPredictor.cpp \
-    zlib.cpp
+                zlib.cpp \
```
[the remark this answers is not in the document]
sakshikukreja14: I faced this same error in doctest.pro. It was giving a segment fault for this and not running the test case. I have removed now.

src/core/libmaven/jsonReports.cpp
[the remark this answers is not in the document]
saifulbkhan: Hmm. Empty sample names. That's strange. But okay.

src/core/libmaven/jsonReports.cpp
saifulbkhan: Since the test cases only need to access these structures, they can simply be functions:
```C++
vector<mzSample*> samples();
vector <PeakGroup> allgroups();
MavenParameters* mavenparameters();
```

tests/doctest/doctest.pro
```diff
@@ -0,0 +1,52 @@
+include($$mac_compiler)
+DESTDIR = $$top_srcdir/bin/
+
+
+OBJECTS_DIR=$$top_builddir/tmp/doctest/
+include($$mzroll_pri)
+TEMPLATE = app
+TARGET = doctest
+
+
```
saifulbkhan: Can you also add:
```
CONFIG+=cmdline
```
here? This will make our `doctest` binary a proper CLI executable.

src/core/libmaven/jsonReports.h
```diff
@@ -1,32 +1,65 @@
...
+     * @param myfile
+     */
+    void _writeEIC(PeakGroup& grp, ofstream& myfile, mzSample*);
+
+    /**
+     * @brief _writeCompoundLink writes Compound Link of group
+     * @param grp
+     * @param myfile
+     */
+    void _writeCompoundLink(PeakGroup& grp, ofstream& myfile);
```
saifulbkhan: Please don't name variables as `myfile`. It seems too toy-ish for professional code.

PR #1205, opened 2020-01-17
Ref csv reports

sakshikukreja14:


src/core/libmaven/csvreports.h
```diff
@@ -1,161 +1,209 @@
...
+     */
+    SEP = ",";
+    }
+    
+    /**
+     *@param -  incremental group numbering. 
+     *Increment by 1 when a group is added 
+     *for csv report  
+     */
+    int groupId;
```
saifulbkhan: Is this being used? If so I do not thing that it should be publicly accessible.

src/core/libmaven/csvreports.cpp
```diff
@@ -1,517 +1,1087 @@
...
+            }
+        }
+        inputPeakFile.close();
+        savedPeakFile.close();
+        remove("polly.csv");
+    }
+}
+
+
+TEST_CASE_FIXTURE(CSVReportFixture, "Testing untargettedGroups")
```
saifulbkhan: Do we need a separate fixture for untargeted groups? The functions for obtaining them are different anyway. Let's move the sub-cases in the same fixture.

src/core/libmaven/csvreports.cpp
```diff
@@ -1,517 +1,1087 @@
...
+///////////////////////////Test Cases//////////////////////////////
+class CSVReportFixture
+{
+    private:
+            vector<mzSample*> _samples;
+            vector <PeakGroup> _allgroups;
+            MavenParameters* _mavenparameters;
+            std::list<PeakGroup> _isotopeGroups;
+
+    private:
```
saifulbkhan: Remove this extra private modifier.

PR #1208, opened 2020-01-24
[Ftr] Auto-update release builds

saifulbkhan:
Once merged, this will notify and allow users using release builds of El-MAVEN, to update their current installation to the latest available version of the application.

---

Although the basic update functionality has been added, some work still remains:
- [x] Revert to using master packaging branch during deployment once the [changes required in elmaven-packaging repo](https://github.com/ElucidataInc/elmaven-packaging/pull/4) are merged.
- [x] Discuss with product team, how we can use Mixpanel login and the new support dialog for the update functionality. Maybe replace the terrible "Getting Started" dialog.


src/common/downloadmanager.cpp
```diff
@@ -11,6 +11,10 @@ DownloadManager::DownloadManager():
     _manager(new QNetworkAccessManager),
     _reply(nullptr),
     err(false)
+{
```
shubhra-agrawal: linting.

src/gui/mzroll/autoupdater.cpp
```diff
@@ -0,0 +1,124 @@
...
+             << maintenanceToolPath
+             << "for maintenancetool executable.";
+    if (!QFile::exists(maintenanceToolPath)) {
+        qWarning() << "Maintenance tool was not found at its expected path. "
+                      "Cannot proceed with update process.";
+        return;
+    }
+
+    QProcess* maintenanceToolProcess = new QProcess;
+    connect(maintenanceToolProcess,
```
shubhra-agrawal: Who is watching these two signals? `maintenanceToolStarted` and `maintenanceToolClosed`?

saifulbkhan: Right now, nothing. But I felt like there should be some hooks (for future uses) that lets the application know that the tool was started and then closed.

src/gui/mzroll/autoupdater.cpp
```diff
@@ -0,0 +1,124 @@
...
+            &QProcess::started,
+            this,
+            &AutoUpdater::maintenanceToolStarted);
+    connect(maintenanceToolProcess,
+            static_cast<void(QProcess::*)(int, QProcess::ExitStatus)>(&QProcess::finished),
+            this,
+            &AutoUpdater::maintenanceToolClosed);
+
+    maintenanceToolProcess->setProgram(maintenanceToolPath);
+    maintenanceToolProcess->setArguments({"--updater"});
```
rish9511: Why not the `silentUpdate` argument ?

rish9511:
@saifulbkhan @shubhra-agrawal 

> The maintenance tool allows the user to select between "Remove or add components", "Update components" and "Remove all components". The first two seem to be working the same way for now but I suspect it might confuse some users. Is it possible to hide the other options?

Imo the `silentUpdate` argument would be better for this use case. Displaying another UI for asking users to install the update might add some little amount of confusion.  The users are more used to the convention of how updates happen on mobile devices. 
Before making the update we definitely should prompt a UI that briefly informs the user about the new version and ask them whether they want to update or not

src/gui/mzroll/autoupdater.cpp
saifulbkhan: Because with `silentUpdate`, the updater during its init stage tries to (not sure why) run the control script (from packaging repo), which includes some custom GUI logic (for our privacy policy page). And this crashes the non-GUI tool before being able to even check for updates.

There must be a solution for this, which I am open to, but we are a bit tight on time here. Future releases can have behind the scenes auto-updates.

rish9511: @saifulbkhan 
Sounds familiar. And IIRC I did find a fix for it
I think the the logic inside the control script should be made to run only if the `MaintenanceTool` is running in the **Installer** mode(this is the online/offline installer that we prepare using our packaging script) and not in the **Updater** mode(this is when we run the `Maintenancetool` and choose the **Update mode** or when it's run using the **silentUpdate** argument)

saifulbkhan:
> I'm getting the update prompt behind the splash screen/getting started dialog. Is it possible to add some lag time to the prompt? Even 30 seconds should work. (A toast instead of a messagebox might be less intrusive but I'm guessing that's a product call)

I will add a 30 second delay to the prompt.

> I tried installing the update from v0.9.2. The installation of v0.9.5 happened successfully but I opened the app and the title shows 0.9.3 instead. Not sure what happened here.

Yes I think that was the change I had made between v0.9.2 and v0.9.5. Showing v0.9.3 in the window title, regardless of the actual tag. Still, I will check to make sure nothing funny is going on.

> The maintenance tool allows the user to select between "Remove or add components", "Update components" and "Remove all components". The first two seem to be working the same way for now but I suspect it might confuse some users. Is it possible to hide the other options?

> Once the update is installed, the maintenance tool gives me an option to "Restart" which sounds like it will restart El-MAVEN with the new version but it just restarts the maintenance tool. I see that restarting El-MAVEN might not be within the scope of this PR but is it possible to hide that option? I would suggest opening the folder or just displaying the system path for the new binary if possible. Not a requirement though.

I agree with your reasoning here. But unfortunately, I was not able find any documentation (or even threads) around customizing the _maintenance tool_ itself.

src/gui/mzroll/autoupdater.cpp
rish9511:
```
function Component()
{
    ...

    installer.setDefaultPageVisible(QInstaller.StartMenuSelection, false)
    
    if(installer.isInstaller()) {
        installer.addWizardPage(component, "PolicyPage", QInstaller.PerformInstallation)
        gui.pageByObjectName("DynamicPolicyPage").entered.connect(enteredPolicyPage)
    }

   ...
}
```
I remember adding the `if` condition in `installerscript.qs` itself, so that it does not add the **Policy page**  in the first place. 
Let me know if this works

saifulbkhan: @rish9511 Got it. Will take a look at this and explore silent updates in the next release.

.travis.yml
```diff
@@ -133,14 +136,16 @@ after_success:
     fi
 
     export PACKAGING_REPO=elmaven-packaging
-    git clone https://github.com/ElucidataInc/elmaven-packaging.git
+    # TODO: reset this to master branch after merge
```
shubhra-agrawal: Has the packaging PR been merged? please update

saifulbkhan: Yes it has. I will remove the comment.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2725,28 +2740,13 @@ void MainWindow::createMenus() {
...
 	connect(faq, SIGNAL(triggered()), signalMapper, SLOT(map()));
 
-	QAction* start = helpMenu->addAction("Getting Started");
-	connect(start,SIGNAL(triggered(bool)), gettingstarted, SLOT(show()));
-
 	signalMapper->setMapping(doc, 1);
 	signalMapper->setMapping(tutorial, 2);
 	signalMapper->setMapping(faq, 3);
 
-	connect(signalMapper, SIGNAL(mapped(int)), this, SLOT(openURL(int)));
```
shubhra-agrawal: The Help menu URLs are dependent on this. Has this been moved somewhere else?

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2725,28 +2740,13 @@ void MainWindow::createMenus() {
...
-
 	menuBar()->show();
 }
 
-void MainWindow::openURL(int choice)
```
shubhra-agrawal: this too

src/gui/mzroll/mainwindow.cpp
saifulbkhan: I just searched to make sure that this is not being used anywhere. Maybe you are confusing with `openUrl` which is a Qt utility for opening URLs. This removed one seems redundant.

shubhra-agrawal: The `signalMapper` maps which link is opened on clicking which option in the Help menu. The `openURL` method uses the Qt method `openUrl` internally to open such links but I just checked and that method has been removed in this PR for some reason.

Just compiled it to confirm and the clicks aren't working.

saifulbkhan: This should not have been removed for any reason. I will replace the functionality.

shubhra-agrawal:
@saifulbkhan I've approved it. Would prefer it if the product team tests this before merging it

saifulbkhan:
I have approval from the product team. Merging the branch.

Issue #171, opened 2017-03-03
marking in peaks table: enable "unmark" and also "export good+unmarked"


chubukov:
Some additional features that might be nice:

- be able to remove a good/bad mark
- be able to export good+unmarked 

As suggested by Swetabh, the second one could be part of a massive overhaul where you can filter the table however you like before export, but I think this simple fix would be useful.  In a sense, setting the peak picking parameters is how you pre-filter the table by mass, intensity, etc.

Suggested by @myasteadman

lparsons:
This would be very useful for us as well. It's quite annoying when one accidentally marks a peak as good/bad but in reality I haven't made a decision (it's an undecided or unclear peak). Having peaks unmarked is the only way to keep track of that situation at the moment.

saifulbkhan:
@chubukov Both your requests - unmarking a marked peak and exporting all peaks excluding the bad ones - are now available in v0.10.0.

PR #1211, opened 2020-01-24
[Ftr] Allow unmarking peak-groups #171

saifulbkhan:
Users will now be able to unmark any previously marked (good or bad) peak-groups. Additionally, they will also be able to export non-bad peak groups to CSV.

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -2164,14 +2164,14 @@ QWidget *TableToolBarWidgetAction::createWidget(QWidget *parent) {
 
     QToolButton *btnGood = new QToolButton(parent);
     btnGood->setIcon(QIcon(rsrcPath + "/markgood.png"));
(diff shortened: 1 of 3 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1011,20 +1017,22 @@ void TableDockWidget::setGroupLabel(char label) {
       QVariant v = item->data(0, Qt::UserRole);
       PeakGroup *group = v.value<PeakGroup *>();
       if (group != NULL) {
-        if (!(group->label=='g'||group->label=='b')){
+        group->setLabel(label);
+        if (group->label=='g' || group->label=='b') {
```
shubhra-agrawal: good/bad label was being checked before setting the label to ensure that changing the label for the same group (from 'g' to 'b' or vice versa) is not counted as a new group being marked.

saifulbkhan: Corrected.

PR #1214, opened 2020-01-27
Ref compound

sakshikukreja14:


src/core/libmaven/csvreports.cpp
```diff
@@ -60,6 +60,10 @@ CSVReports::CSVReports(string filename,
         _insertGroupReportColumnNamesintoCSVFile(
             filename, _prmReport, _includeSetNamesLine);
     }
+
+    else{
+        _reportStream.open(filename.c_str(), ios::out);
+    }
```
saifulbkhan: So, what's this? I don't understand what we are doing here.

sakshikukreja14: This for opening the file when it is for function "write for polly". As we need no headings like for peak file and group file. Therefore, It has been given in else part. 
I will add this as comment in the code

saifulbkhan: Can you remind me, why we are having to change things in the CSVReports class in this PR? We probably discussed it, but its been a long time since and I cannot recall what it was. :|

sakshikukreja14: Yeaa, because after we refactored Compound class, there was problem of isotopes while writing to csvReports class. 
For example, 'glucose-6-phosphate' was being changed to 'fructose-6-phosphate'.

saifulbkhan:
@sakshikukreja14 In your second commit of this PR (commit ID: [`e5a811d`](https://github.com/ElucidataInc/ElMaven/pull/1214/commits/e5a811d6022ea07ff6c1f82e630266cfcbfcfe5a)), you mention some change that would allow the JsonReports tests to check against groups regardless of their order. But I do not see any relevant code change there.

saifulbkhan:
Similarly, commit [ae54164](https://github.com/ElucidataInc/ElMaven/pull/1214/commits/ae5416480021d68ae5c9a85385354eb6a092f1bf), is also empty - even though there is a description for a change.

So, I just checked your [`ref_csvReports`](https://github.com/sakshikukreja14/ElMaven/commits/ref_csvReports) branch (that was already merged in v0.10 cylce), and your first three commits in this PR are from that branch. Maybe this was the result of an incorrect rebasing from your original base (which I suspect was `ref_csvReports`) to the current `develop`.

Let me know when you fix this, and I will continue reviewing the actual content of this PR.

src/core/libmaven/Compound.h
```diff
@@ -46,147 +94,157 @@ class Compound{
...
-         */
-        string db;
+        *@param  -   Reaction represents simple compound reaction
+        *Check Reaction class in mzSample.h
+        */
+
+        vector<Reaction*>reactions;
+
+        // TODO: from MAVEN (upstream), find out what this is
+        string adductString;
```
saifulbkhan: Pretty sure we do not need this anymore, since `Adduct`s are separately embedded to a `PeakGroup` object, to create a adductized compound form.

src/core/libmaven/Compound.h
```diff
@@ -46,147 +94,157 @@ class Compound{
...
         /**
-         * @brief Name of database this compound belongs to for example KEGG,
-         * ECOCYC, etc.
-         */
-        string db;
+        *@param  -   Reaction represents simple compound reaction
+        *Check Reaction class in mzSample.h
+        */
+
+        vector<Reaction*>reactions;
```
saifulbkhan: Is `Compound::reactions` used anyhow?

src/core/libmaven/Compound.h
```diff
@@ -46,147 +94,157 @@ class Compound{
...
+        float neutralMass();
 
         /**
-         * @brief Obtain the formula set for this compound.
-         * @return A string representing a valid chemical formula.
+         * @brief Categories of this compound. For e.g., amino acids, nucleic
+         * acids, peptide, etc.
          */
-        string formula() const { return _formula; }
+        vector<string> category;
```
saifulbkhan: Let's make this (and the any other attributes below) private, too?

src/core/libmaven/Compound.cpp
```diff
@@ -72,17 +79,35 @@ bool Compound::operator == (const Compound& rhs) const
...
+    _formula = rhs._formula;
+    _srmId = rhs._srmId;
+    _expectedRt = rhs._expectedRt;
+    _charge = rhs._charge;
+    _mass = rhs._mass;
+    _precursorMz = rhs._precursorMz;
+    _productMz = rhs._productMz;
+    _collisionEnergy = rhs._collisionEnergy;
+    category = rhs.category;
+}
```
saifulbkhan: These are not the only attributes that should be copied to give you an identical `Compound` object. Take a look at all the properties being compared in the `==` operator above. All of those should be here, at least.

src/core/libmaven/Compound.h
saifulbkhan: And add tests for them too?

src/core/libmaven/Compound.h
```diff
@@ -12,29 +12,77 @@ class Fragment;
...
+
+        /**
+         *@param  -   number of charge of compound (electron loss or gain)
+        */
+        int _charge;
+
+        /**
+         *@param   -  mass of this compund
+        */
+        float _mass;
```
saifulbkhan: In almost all cases, this member stores the compound's m/z (mass divided by charge) and not the mass. Only when the compound is initialized and charge is not available, is this set to the neutral mass. So, I suggest we rename it to `_mz` and change its corresponding setter and getter to `setMz()` and `mz()`, respectively.

src/core/libmaven/databases.cpp
```diff
@@ -213,8 +213,19 @@ bool Databases::addCompound(Compound* c) {
         for(unsigned int i = 0; i < compoundsDB.size(); i++) {
             Compound* currentCompound = compoundsDB[i];
             if ( currentCompound->db() == c->db() && currentCompound->id() == c->id()) {
-                currentCompound = c;
```
saifulbkhan: Why was the assignment not sufficient? The fix makes the same changes, but assigns individual attributes instead. The original one-liner looked more elegant. Maybe it is because the assignment does not take care of all the members of a compound? I mentioned this in another comment earlier.

tests/test-libmaven/utilities.h
```diff
@@ -0,0 +1,171 @@
+#ifndef DTUTILS_H
+#define DTUTILS_H
```
saifulbkhan: "utilities.h" is a very general name. Ideally we'd reserve this for some project-wide utilities. Let's call it `testutils.h`? or anything that tells the user that file brings in test fixtures. Also change the include guards to according to the file name.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -350,9 +344,7 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
             break;
 
         mzSlice* slice = slices[s];
-        Compound* compound = slice->compound;
-        if (compound != nullptr && compound->hasGroup())
-            compound->unlinkGroup();
```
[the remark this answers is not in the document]
sakshikukreja14: Because you said now we don't link compounds with groups and you told me to remove these functions and corresponding code where those functions were being used.

src/gui/mzroll/database.cpp
```diff
@@ -874,23 +874,15 @@ int Database::loadCompoundCSVFile(string filename)
...
-            compound->expectedRt = rt;
-
-            if (mass != 0.0f)
-                compound->neutralMass = mass;
-
-            if (mz == 0 && !formula.empty()) {
-                mz = MassCalculator::computeMass(formula,charge);
-            } else if (mass != 0.0f) {
-                mz = MassCalculator::adjustMass(mass, charge);
-            }
```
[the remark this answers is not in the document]
sakshikukreja14: No, this was not intentional. May be because of the initial rebasing, these checks had been removed and same was with neutral mass. I had added yesterday where ever i got them. But still which were left I have added now after your review

src/core/libmaven/Compound.h
```diff
@@ -12,29 +12,77 @@ class Fragment;
...
+         * @brief Name of database this compound belongs to for example KEGG,
+         * ECOCYC, etc.
+         */
+        string _db;
+
+        float _neutralMass;
+
     public:
         enum class Type {
             MS1,
```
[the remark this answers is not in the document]
sakshikukreja14: yea sure i will make that change

src/core/libmaven/Compound.h
```diff
@@ -46,147 +94,157 @@ class Compound{
...
+         * @brief Vector of intensities of fragments generated from this
+         * compund.
+         */
+        vector<float>fragmentIntensities;
+
+        /**
+         * @brief Collection of indices of fragment values mapping to its
+         * ionisation type information.
          */
-        static string filterFormula(string formulaString)
```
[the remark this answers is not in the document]
sakshikukreja14: This function was not being used anywhere in the codebase of ElMaven. So, I removed it

src/core/libmaven/Compound.h
```diff
@@ -46,147 +94,157 @@ class Compound{
...
 
-        /** QQQ mapping */
-        string method_id;   /**@param  -  TODO*/
-        float precursorMz;	/**@param  -  QQQ parent ion  mz value   */
-        float productMz;    /**@param  -  QQQ child ion   mz value */
-        float collisionEnergy; /**@param  -   QQQ collision energy of this compound   */
         float logP; // TODO: find out what value this is.
 
         // TODO: from MAVEN (upstream), find out what this is
         bool virtualFragmentation;
```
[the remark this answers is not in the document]
sakshikukreja14: Yes, both of these are being in used files related to databases.

src/core/libmaven/Compound.h
sakshikukreja14: yes, It is just being set in database file while reading from database and nowhere else is used. So, we can remove it.

src/core/libmaven/Compound.h
sakshikukreja14: umm.. it's being set from database but not as such used for computations.

tests/test-libmaven/utilities.h
sakshikukreja14: As testutils was similar to the file in maven tests. Therefore, I have named it as testLibmavenUtils

src/core/libmaven/databases.cpp
sakshikukreja14: umm.. Actually, I tried doing this. Assigning all the values but the thing is  currentCompound is getting its value from compoundDB[i] and only specific attributes from 'c' are being set. Thus, using "currentCompound = c" is not giving the desired results.

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
saifulbkhan: Yes. Sorry, I only saw this later and forgot to delete this comment.

src/core/libmaven/Compound.h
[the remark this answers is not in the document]
saifulbkhan: It was being used in the original `Compound::setFormula` method. Can you make restore this method and make that change?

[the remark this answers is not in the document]
sakshikukreja14: oh, yea sure

src/core/libmaven/Compound.h
saifulbkhan: Let's remove anything regarding reactions for now.

sakshikukreja14: Okkay sure

src/core/libmaven/databases.cpp
saifulbkhan: What are the desired results here? Because if we are assigning attributes of a compound to another one, then there should be a reason to do this only partially.

src/core/libmaven/Compound.cpp
saifulbkhan: This one still needs attention.

tests/test-libmaven/utilities.h
saifulbkhan: um ... testutils.h from MavenTests and testutils.h from test-libmaven will almost never be included in the same namespace. So its okay to name them the same. `testLibmavenUtils` looks ugly :P

src/core/libmaven/databases.cpp
sakshikukreja14: Due to this assignment, original problem of isomers shows back that 'glucose-6-phosphate' doesn't come in the result set and instead 'fructose-6-phosphate' is shown

src/core/libmaven/Compound.cpp
sakshikukreja14: I made the change, assigned all attributes. But I will still have a look

tests/test-libmaven/utilities.h
sakshikukreja14: Okkay, I will make the change

src/core/libmaven/Compound.h
```diff
@@ -11,85 +11,186 @@ class Fragment;
...
+
+        string id();
+
+        void setName(string name);
+
+        string name();
+
+        void setFormula(string formula) {
+            _formula = filterFormula(formula);
+        }
```
saifulbkhan: Put definition in `compound.cpp`.

sakshikukreja14: hanji.. just a sec

Issue #782, opened 2018-07-11
Option to visualize area being integrated using shift-drag


sunil20dhakad:
Just a suggestion if we can have 'Area' as a default setting instead of 'AreaTop' because most of   the people are using 'Area' for their calculation.
 
@shubhra-agrawal @shefalilathwal @chubukov FYI

chubukov:
@sunil20dhakad what's the basis for the claim that most people are using Area? Actual surveys or just a feeling?

You can do whatever you like, but I would caution against doing this. I think one reason that Maven has given pretty robust results without a particularly detailed peak model is precisely because of the use of AreaTop. If you use Area, you have to be a little more careful in making sure peak bounds and baseline are properly defined. Additionally, it would be good to have a good visualization of the area calculation, which isn't really there.

lparsons:
We use AreaTop as well, and I don't think changing the default is warranted. However, it would make sense to preserve a users setting for this between sessions.

shefalilathwal:
@lparsons @chubukov I think it's the bias from the fact that most of the people we work with use Area. It's fair to say that we don't know what most people in the field are using. I am curious to hear under what conditions you would use Area/AreaTop/Peak Height. To me, area seems to be the most logical measure, especially when the peak shapes are not gaussian and you want to calculate %labels or try to be quantitative.  Also, agreed about visualization of area calculation - we need a better way to see what's being integrated.

lparsons:
Visualization of what is being integrated should be a high priority, IMHO. Also see #476.

raaisakuk:
I have been using Area (unknowingly) because when I do further processing using Polly, it chooses 'Area' from the Peak Detailed format as far as I know. @Raghavdata can confirm
Also, atleast when we use El maven around the lab, we always use shift+drag with carefully selecting the baseline and I was assuming until now that area is being calculated for that region.

raaisakuk:
@shubhra-agrawal I wanted to ask, which one of these columns should be used for further analysis like fractional enrichment calculation or quantification of peaks?
peakArea | peakAreaTop | peakAreaCorrected | peakAreaTopCorrected (I get these in peak detailed format output)
While analysing a data recently, I felt that peakAreaTop reflected my peaks better than peakArea

shubhra-agrawal:
Transparency around the area being integrated seems important, then.

[...]

Issue #476, opened 2017-11-28
Better control of peak integration


lparsons:
It seems difficult to understand (at least visually) how the peak intensity is being calculated. It is often necessary to set some background threshold as well as the width of a peak. It's not clear from the interface what those values are set to (perhaps use shading to make this clear?). It is also not easy/possible to change these values either manually or through the setting of configuration parameters. (Suggested by Juan Carlos Garcia Canaveras in the Rabinowitz Lab)

chubukov:
@lparsons I just wrote this in an email to @sahil21 but maybe I should just comment here

Couple of different things going on here. A) I suspect nobody who's not a classic Maven user knows what AreaTop means (as far as I know Maven is the only piece of software that uses this term). (It's a very slightly smoothed height: the average of the highest intensity scan and the two scans around it). So maybe there’s confusion about how the default intensity is calculated. B) I do think some people expect to be able to integrate peaks the way you would do in a more dedicated software that takes on one sample at a time (click on left edge, click on right edge, click on baseline, etc). I think the shift-drag functionality probably should be documented and shown as some sort of tip (not sure people would guess this on their own). (PS I noticed that shift-drag is inconsistent about calculating Area #528).

In general, I guess it would be nice to be able to isolate a particular peak, and do some much more detailed/careful peak integration (tangent/shoulder baselines, specify bounds, etc, etc). That could turn Maven into a more full-fledged choromatographic analysis program. It also seems like a lot of effort and it's not really the way that we use Maven, so I for one haven't been pushing for it. Of course would be very happy to contribute ideas/test/etc if we decided to take this on.

raaisakuk:
@sahil21 I agree with the visual part completely. In the current plot it is really hard to understand what exactly is being integrated. I think some of my feature requests are around this. Now that we are planning to use El MAVEN across the lab, this will become all the more crucial.

saifulbkhan:
This use-case should now be satisfied using the peak-editor widget introduced in v0.11.0. Closing this.

PR #1218, opened 2020-01-30
[Ftr] Visualize integrated area #782

saifulbkhan:
Being able to (at least) visualize the area being integrated, either manually or automatically, has been missing from the software. This PR attempts to take a shot at solving this problem.

Pressing _Shift_ key will enter "area integration mode" (the same will happen using integration button on EIC toolbar), in which chromatogram is faded and upon mouse dragging becomes prominent highlighting the area that will be considered for integration of the peaks. Maybe the visuals can be improved, but it seems good enough to me for now.

Some other notes:
1. Releasing _Shift_ before mouse button exits integration mode without registering any selected peak area. This is intentional.
2. The area for all peak-groups in peak tables will be visible in the EIC in a similar to how it looks while being manually integrated.
3. Unfortunately, editing the bounds of an existing peak-group is still not possible (#476), yet. This involves not only adding some interactivity to the EIC widget, but also inspecting all the assumptions that have been made until now about the immutability of peak-group bounds, once created. That seemed a bit out-of-scope for this PR.

### DIFF
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -81,6 +81,7 @@ EicWidget::EicWidget(QWidget *p) {
     _mouseEndPos = _mouseStartPos = QPointF(0,0); //TODO: Sahil, added while merging eicwidget
     _ignoreTolerance = false;
     _ignoreMouseReleaseEvent = false;
+    _selectionLine = nullptr;
 
 	connect(scene(), SIGNAL(selectionChanged()), SLOT(selectionChangedAction()));
     connect(this, &EicWidget::eicUpdated, this, &EicWidget::setGalleryToEics);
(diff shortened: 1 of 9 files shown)


saifulbkhan:
@shubhra-agrawal As discussed offline, I have reinstalled the red selection line, which @surbhipoddar suggested to keep persistent for groups in peak tables (it will serve to display RT bounds during and after integration). We decided to keep the area shading as it is. Also, the bug you mentioned (peaks showing as if integrated after zooming in) is also fixed now.

src/gui/mzroll/eicwidget.cpp
```diff
@@ -918,7 +1074,12 @@ void EicWidget::unSetPeakTableGroup(PeakGroup* group)
     }
 }
 
-void EicWidget::replot(PeakGroup* group) {
+void EicWidget::replot(PeakGroup* group)
+{
+    if (_areaIntegration) {
```
shubhra-agrawal: Why is areaIntegration relevant in the replot method?

Issue #1215, opened 2020-01-27
unable to start concurrent sessions on version 0.9 and above


chubukov:
Starting another instance when one is already running leads to the following error. This happens to me on both 0.9 and 0.9.1 that I installed using the packaged installer, but not with 0.8 which I compiled myself.

This is on windows server 2016. 

Seems like a permissions issue with the log file.

![image](https://user-images.githubusercontent.com/13080968/73194734-a6a42b80-40fa-11ea-978e-f44e3d5601ad.png)

saifulbkhan:
@chubukov Sorry about this. We are aware and have identified the cause of this issue. It will be fixed this with v0.10.0 (expected mid February).

saifulbkhan:
Should no longer be an issue in v0.10.0. Sorry about this mishap.

PR #1219, opened 2020-01-31
[Fix] Re-enable running concurrent sessions #1215

saifulbkhan:
The log files, that were introduced for some utilities in the last release, were putting a restriction of being editable by only one process (most prominently on Windows), which lead to a second parallel session of El-MAVEN to throw a write exception as soon as it started. To prevent this event, log files will now write to a separate timestamped log, that share the same prefix for files from a single session but are different for files of a different El-MAVEN/peakdetector session.

Note: this fix also relies on the fact that, as soon as an executable begins they create at least one log that is created using the static timestamp - which is true for both El-MAVEN (which creates a DownloadManager log) and peakdetector (which creates a log for itself) - while starting up.

### DIFF
--- a/src/cli/peakdetector/main.cpp
+++ b/src/cli/peakdetector/main.cpp
@@ -33,7 +33,8 @@ int main(int argc, char *argv[]) {
 	// Polly CLI part over..
 
     QString parentFolder = "ElMaven";
-    QString logFile = "peakdetector_cli.log";
+    QString logFile = QString::fromStdString(Logger::constant_time()
+                                             + "_peakdetector_cli.log");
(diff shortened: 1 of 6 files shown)


shubhra-agrawal:
@saifulbkhan Will the log files be deleted at any time? Also, what's the approx size of each log file?

saifulbkhan:
Logs will not be deleted automatically. I don't know the average size, but the size of the largest one on my system right now is 72 KB.

shubhra-agrawal:
Got it. I have a personal distaste for leaving around indefinite number of log files so it'll be good if we can delete month old logs? 

Approving this PR anyway.

Issue #571, opened 2018-01-29
Isotope widget and peak hover intensity always displays "height" regardless of peak quantitation type setting


lparsons:
The bookmark table displays "intensity" in the specified *Peak Quntitation Type* (sic) (e.g. AreaTop,  Height, etc.). However, the Isotope Widget and Peak hover always displays the intensity quantified using the "Height" method.
![image](https://user-images.githubusercontent.com/645128/35535437-923a4b60-0511-11e8-8a19-d81a0d9640ad.png)
![image](https://user-images.githubusercontent.com/645128/35535427-8a42e930-0511-11e8-9f2f-2adc1918d706.png)

saifulbkhan:
@lparsons The isotope widget will now show intensity using the globally selected quantitation type, starting v0.10.0.

PR #1220, opened 2020-01-31
[Fix] Update isotope and peak intensity with quantitation #571

saifulbkhan:
This PR contains fixes for the inconsistencies pointed out in #571. Depending on the quantitation type set by the user:
* update the abundance values shown in isotope widget,
* update the quantity shown in the tool-tip that pops up when hovering on a peak.

### DIFF
--- a/src/gui/mzroll/isotopeswidget.cpp
+++ b/src/gui/mzroll/isotopeswidget.cpp
@@ -323,12 +323,16 @@ void IsotopeWidget::populateByParentGroup(vector<Isotope> masslist, double paren
 		if (!peak)
 			continue;
 
+        vector<mzSample*> samples = {_selectedSample};
+        auto quantity = child.getOrderedIntensityVector(samples,
+                                                        _mw->getUserQuantType()).at(0);
+
 		mzLink link;
 		link.mz1 = parentMass;
 		link.mz2 = child.expectedMz;
 		link.note = isotopeName;
-		link.value1 = child.expectedAbundance;
-		link.value2 = peak->peakIntensity;
+        link.value1 = child.expectedAbundance;
+        link.value2 = quantity;
 		isotopeParameters->links.push_back(link);
 	}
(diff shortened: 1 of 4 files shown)


src/gui/mzroll/isotopeswidget.cpp
```diff
@@ -323,12 +323,16 @@ void IsotopeWidget::populateByParentGroup(vector<Isotope> masslist, double paren
...
+                                                        _mw->getUserQuantType()).at(0);
+
 		mzLink link;
 		link.mz1 = parentMass;
 		link.mz2 = child.expectedMz;
 		link.note = isotopeName;
-		link.value1 = child.expectedAbundance;
-		link.value2 = peak->peakIntensity;
+        link.value1 = child.expectedAbundance;
+        link.value2 = quantity;
```
shubhra-agrawal: `quantity` is a vector but`link.value2` expects a float. Has this been tested?

saifulbkhan: `quantity` is the 0th value in from the intensity vector. Yes this has been tested.

shubhra-agrawal: Oops. Didn't notice the `.at(0)`, sorry.

But I don't quite understand. The value2 needs to be the intensity for the selected peak. Why take the lowest intensity.

saifulbkhan: Sending only one sample (see line 326), which is the peak's sample. So we are getting the correct quantity.

shubhra-agrawal: got it. Forgot about that. Guess I shouldn't be reviewing PRs in batches :P

shubhra-agrawal:
Rebase please.

saifulbkhan:
I will rebase and merge, soon. Working on something else ATM.

PR #1222, opened 2020-02-05
[Ftr] Push directly to Polly project

saifulbkhan:
Users may want to use services on Polly, that they have licenses for, but is not part of the three existing integrations (FirstView, PollyPhi and QuantFit). Currently, they will have to export their files and upload to Polly as two separate steps. Instead we are adding a "Send to Project" feature, where they can simply upload their data to Polly without having a workflow associated to it.

- [x] To-do: Revert to using the master "index.js", once [this PR](https://github.com/ElucidataInc/polly-cli/pull/14) gets merged.

### DIFF
--- a/src/gui/mzroll/pollyelmaveninterface.cpp
+++ b/src/gui/mzroll/pollyelmaveninterface.cpp
@@ -56,9 +56,13 @@ PollyElmavenInterfaceDialog::PollyElmavenInterfaceDialog(MainWindow* mw)
     gotoPollyButtonAlt->setVisible(false);
 
     groupSetCombo->addItem("All Groups");
+    groupSetComboAlt->addItem("All Groups");
     groupSetCombo->addItem("Only Good Groups");
+    groupSetComboAlt->addItem("Only Good Groups");
     groupSetCombo->addItem("Exclude Bad Groups");
+    groupSetComboAlt->addItem("Exclude Bad Groups");
     groupSetCombo->setCurrentIndex(0);
+    groupSetComboAlt->setCurrentIndex(0);
 
     _selectedMode = SendMode::PollyApp;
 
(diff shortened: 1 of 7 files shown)


src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -68,13 +72,61 @@ PollyElmavenInterfaceDialog::PollyElmavenInterfaceDialog(MainWindow* mw)
...
+            &PollyElmavenInterfaceDialog::_enableNewProjectUi);
+    connect(newProjectRadio,
+            &QRadioButton::toggled,
+            newProjectRadioAlt,
+            &QRadioButton::setChecked);
+    connect(newProjectRadioAlt,
+            &QRadioButton::toggled,
+            newProjectRadio,
+            &QRadioButton::setChecked);
+    connect(newProjectEntry,
```
shubhra-agrawal: This is great!

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -966,6 +1117,107 @@ QStringList PollyElmavenInterfaceDialog::_prepareFilesToUpload(QDir qdir,
...
+
+    // write a JSON and CSV file for each peak table
+    QList<QPointer<TableDockWidget>> peakTableList =
+        _mainwindow->getPeakTableList();
+    peakTableList.append(_mainwindow->getBookmarkedPeaks());
+    for (TableDockWidget* peakTable : peakTableList) {
+        if (peakTable->getGroups().isEmpty())
+            continue;
+
+        QString tableName = TableDockWidget::getTitleForId(peakTable->tableId);
```
shubhra-agrawal: The table name could have spaces, might be a good idea to strip it before using it as the filename.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -966,6 +1117,107 @@ QStringList PollyElmavenInterfaceDialog::_prepareFilesToUpload(QDir qdir,
...
+                               + tableName
+                               + ".json";
+        peakTable->exportJsonToPolly(_writeableTempDir, jsonFilename, false);
+        filenames.append(jsonFilename);
+
+        QCoreApplication::processEvents();
+
+        QString csvFilename = datetimestamp + "_" + tableName + ".csv";
+        peakTable->treeWidget->selectAll();
+        peakTable->prepareDataForPolly(_writeableTempDir,
```
shubhra-agrawal: Group Summary is helpful for LCMS and DDA data but Polly requires Peak detailed format for MRM datasets. Maybe send both formats?

saifulbkhan: I will just send both, in case the other is also needed for something specific.

src/gui/mzroll/pollyelmaveninterface.cpp
saifulbkhan: The next line (1134) does exactly this.

shubhra-agrawal: Sorry for that. Distracted by other stuff.

shubhra-agrawal:
@saifulbkhan Looks fine. Just waiting for the Peak detailed CSV and @surbhipoddar's comment.

surbhipoddar:
@shubhra-agrawal @saifulbkhan : It makes sense to the peak table that is being sent to the projects. We should only send the JSON corresponding to the selected peak table. 

However, we do not need the good/bad peaks to be sent. I think this feature will change with ML integration.

shubhra-agrawal:
@surbhipoddar The ML integration is a proprietary feature that won't be available for other users. The rest of the users will continue using the current system for peak curation and will need the option to choose to send their curated data to Polly.

saifulbkhan:
@surbhipoddar @shubhra-agrawal I guess there isn't any harm in providing the good/bad selection set, so I've went ahead and added both the dropdowns (peak-table and curation set) for selective upload.

PR #1223, opened 2020-02-05
[ref] Refactored "mzUtils" file and tested using doctest

sakshikukreja14:
mzUtils contained some functions which were not being used. These functions have been removed from the file. Also, the functions which did not follow the latest coding style of c++ have been refactored.
Secondly, the tests have been applied to functions in the mzUtils using doctest.

### DIFF
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -116,20 +116,20 @@ string mzSample::getFileName(const string& filename)
     return ("");
 }
 
-void mzSample::loadAnySample(const char* filename)
+void mzSample::loadAnySample(string filename)
 {
-    if (mystrcasestr(filename, "mzCSV") != NULL) {
-        parseMzCSV(filename);
-    } else if (mystrcasestr(filename, "mzdata") != NULL) {
-        parseMzData(filename);
-    } else if (mystrcasestr(filename, "mzxml") != NULL) {
-        parseMzXML(filename);
-    } else if (mystrcasestr(filename, "mzml") != NULL) {
-        parseMzML(filename);
-    } else if (mystrcasestr(filename, "cdf") != NULL) {
-        parseCDF(filename, 1);
+    if (compareStr(filename, "mzCSV") != " ") {
+        parseMzCSV(filename.c_str());
+    } else if (compareStr(filename, "mzdata") != " " ) {
+        parseMzData(filename.c_str());
+    } else if (compareStr(filename, "mzxml") != " ") {
+        parseMzXML(filename.c_str());
+    } else if (compareStr(filename, "mzml") != " ") {
+        parseMzML(filename.c_str());
+    } else if (compareStr(filename, "cdf") != " ") {
+        parseCDF(filename.c_str(), 1);
     } else {
-        parseMzData(filename);
+        parseMzData(filename.c_str());
     }
 }
 
@@ -155,7 +155,7 @@ void mzSample::checkSampleBlank(const char* filename)
     }
 }
 
-void mzSample::loadSample(const char* filename)
+void mzSample::loadSample(string filename)
 {
     // Loading and Decoding the file
     // catch any error while parsing
@@ -174,10 +174,10 @@ void mzSample::loadSample(const char* filename)
     calculateMzRtRange();
 
     // Setting Sample name
-    sampleNaming(filename);
+    sampleNaming(filename.c_str());
 
     // Checking if a sample is blank or not
-    checkSampleBlank(filename);
+    checkSampleBlank(filename.c_str());
 }
 
 void mzSample::parseMzCSV(const char* filename)
@@ -208,7 +208,7 @@ void mzSample::parseMzCSV(const char* filename)
     while (getline(myfile, line)) {
         lineNum++;
         vector<string> fields;
-        mzUtils::split(line, ',', fields);
+        mzUtils::split(line, ",", fields);
         if (fields.size() >= 5 && lineNum > 1) {
             ss.clear();
 
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -118,15 +118,15 @@ string mzSample::getFileName(const string& filename)
 
 void mzSample::loadAnySample(string filename)
 {
-    if (contains(filename, "mzCSV") == true) {
+    if (contains(filename, "mzCSV")) {
         parseMzCSV(filename.c_str());
-    } else if (contains(filename, "mzdata") == true ) {
+    } else if (contains(filename, "mzdata")) {
         parseMzData(filename.c_str());
-    } else if (contains(filename, "mzxml") == true) {
+    } else if (contains(filename, "mzxml")) {
         parseMzXML(filename.c_str());
-    } else if (contains(filename, "mzml") == true) {
+    } else if (contains(filename, "mzml")) {
         parseMzML(filename.c_str());
-    } else if (contains(filename, "cdf") == true) {
+    } else if (contains(filename, "cdf")) {
         parseCDF(filename.c_str(), 1);
     } else {
         parseMzData(filename.c_str());
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -118,15 +118,15 @@ string mzSample::getFileName(const string& filename)
 
 void mzSample::loadAnySample(string filename)
 {
-    if (compareStr(filename, "mzCSV") != " ") {
+    if (contains(filename, "mzCSV") == true) {
         parseMzCSV(filename.c_str());
-    } else if (compareStr(filename, "mzdata") != " " ) {
+    } else if (contains(filename, "mzdata") == true ) {
         parseMzData(filename.c_str());
-    } else if (compareStr(filename, "mzxml") != " ") {
+    } else if (contains(filename, "mzxml") == true) {
         parseMzXML(filename.c_str());
-    } else if (compareStr(filename, "mzml") != " ") {
+    } else if (contains(filename, "mzml") == true) {
         parseMzML(filename.c_str());
-    } else if (compareStr(filename, "cdf") != " ") {
+    } else if (contains(filename, "cdf") == true) {
         parseCDF(filename.c_str(), 1);
     } else {
         parseMzData(filename.c_str());
@@ -208,7 +208,7 @@ void mzSample::parseMzCSV(const char* filename)
     while (getline(myfile, line)) {
         lineNum++;
         vector<string> fields;
-        mzUtils::split(line, ",", fields);
+        fields = mzUtils::split(line, ",");
         if (fields.size() >= 5 && lineNum > 1) {
             ss.clear();
 
--- a/src/core/libmaven/mzUtils.cpp
+++ b/src/core/libmaven/mzUtils.cpp
@@ -448,7 +448,7 @@ namespace mzUtils {
 
         //find best fit
         if (ycoord.size()<3){
-            pair<float, float> res = make_pair(0.0, 0.0);
+            pair<float, float> res = make_pair(FLT_MAX, FLT_MAX);
             return res;
         }
         vector<float>yobs = ycoord;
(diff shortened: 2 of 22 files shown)


src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,210 @@
...
-
-        row = CsvParser_getRow(csvparser);
-        const char **rowFields = CsvParser_getFields(row);
-        for (int i = 0 ; i < CsvParser_getNumFields(row) ; i++) {
-            v.push_back(rowFields[i]);
-        }
+    void split(const string& str, const string& sep, vector<string>& result)
+    {
+        boost::split(result, str, boost::is_any_of(sep));
     }
```
saifulbkhan: Why can we not return the `result` instead of passing it as a reference?

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,210 @@
...
-            v.push_back(s.substr(i, j-i));
-            i = ++j;
-            j = s.find(c, j);
-
-            if (j == string::npos)
-                v.push_back(s.substr(i, s.length( )));
+    std::string makeLowerCase(string &str)
+    {
+        for (size_t i = 0; i != str.length(); ++i) {
+            str[i] = std::tolower(str[i]);
```
saifulbkhan: Use `transform(str.begin(), str.end(), str.begin(), ::tolower);`. More idiomatic.

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,210 @@
...
+    string compareStr(const string str1,const string str2)
+    {
+        string s = str1;
+        string p = str2;
+        transform(s.begin(), s.end(), s.begin(), ::tolower);
+        transform(p.begin(), p.end(), p.begin(), ::tolower);
+        if(boost::algorithm::contains(s, p))
+            return str1;
+        else
+            return " ";
```
saifulbkhan: A space is not a good fallback return value. What if my `str1` is a space only? You will not be able to tell whether the check was successful or not.

The ideal name for this method would be `contains`. There should be an option to ignore case (which is off by default). And the return type would be a boolean. So here's my prototype:
```C++
bool contains(const string& first, const string& second, const bool ignoreCase = false);
```

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,210 @@
...
-    string substituteInQuotedString(const string& s, const string& chars, const string& substitutions ) {
-        string result;
-        for (string::size_type pos = 0; pos < s.size(); ++pos) {
-            char c = s[pos];
-            string::size_type subst_pos = chars.find_first_of(c);
-            if (subst_pos != string::npos) c = substitutions[subst_pos];
-            result += c;
-        }
-        return result;
+    string compareStr(const string str1,const string str2)
```
saifulbkhan: This is a personal preference but if possible avoid using `1`, `2` or other numeric suffixes to distinguish parameters. Give them better names. Especially in this case, when one is supposed to be checked as the substring of the other.

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,210 @@
...
+                jhigh = ilx;
+
+            for (j = jlow, sum = 0.0; j <= jhigh; ++j)
+                sum += x[j] * input[i-j];
+            result[i] = sum;
         }
     }
 
-    void gaussian1d_smoothing (int ns, int nsr, float *data)
+    void gaussian1d_smoothing (int numSample, int smoothWindowLen, float *data)
```
saifulbkhan: Again, if `data` is the value of interest, then it should be returned and not passed as a pointer reference.

src/core/libmaven/mzUtils.cpp
```diff
@@ -308,131 +247,56 @@ data		1-D array[ns] of smoothed data
...
-Job      :   find the kth smallest element in the array
-Notice   :   use the median() macro defined below to get the median.
-Reference:
-Author:  Wirth, Niklaus
-Title:  Algorithms + data structures = programs
-Publisher:  Englewood Cliffs: Prentice-Hall, 1976
-Physical description: 366 p.
-Series:  Prentice-Hall Series in Automatic Computation
----------------------------------------------------------------------------*/
     float kth_smallest(float a[], int n, int k)
```
saifulbkhan: Is this method even used?

src/core/libmaven/mzUtils.cpp
```diff
@@ -291,13 +227,16 @@ data		1-D array[ns] of smoothed data
     }
 
     float median(float* y, int n) {
```
saifulbkhan: Let's get rid of this other `median` method. Wherever it is being used, send a vector instead.

src/core/libmaven/mzUtils.cpp
```diff
@@ -502,59 +378,52 @@ Series:  Prentice-Hall Series in Automatic Computation
...
+        float t_test = (meanA - meanB) / sqrt(((stdA * stdA) / n1) +
+                                              ((stdB * stdB) / n2));
         return t_test;
     }
 
-
-    int countBelow(vector<float>& y, float ymax) {
-        vector<float>::iterator itr = lower_bound(y.begin(), y.end(), ymax);
-        int lb = itr-y.begin();
+   int countBelow(vector<float>& y, float ymax)
```
saifulbkhan: This method is being used in only one place in the whole codebase. Can we move it to `CompareSamplesLogic` space itslef?

src/core/libmaven/mzUtils.cpp
```diff
@@ -1011,14 +839,15 @@ Series:  Prentice-Hall Series in Automatic Computation
...
+        for ( size_t i = 0; i<strm.total_out; ++i ) {
             uncompressedBytes += uncomp[ i ];
         }
         free( uncomp );
 #endif
         return true ;
     }
 
+
     bool strcasecmp_withNumbers(const std::string& a, const std::string& b ) {
```
saifulbkhan: Change name to `compareStringsWithNumbers`.

saifulbkhan:
@sakshikukreja14 Also, test-libmaven executable is segfaulting on Windows.

src/core/libmaven/mzUtils.cpp
sakshikukreja14: nope.. not used. I will remove it

src/core/libmaven/mzUtils.cpp
sakshikukreja14: This wasn't done because data array contains the values which are used and being edited within the function.

src/core/libmaven/mzSample.cpp
```diff
@@ -116,20 +116,20 @@ string mzSample::getFileName(const string& filename)
...
+    if (contains(filename, "mzCSV") == true) {
+        parseMzCSV(filename.c_str());
+    } else if (contains(filename, "mzdata") == true ) {
+        parseMzData(filename.c_str());
+    } else if (contains(filename, "mzxml") == true) {
+        parseMzXML(filename.c_str());
+    } else if (contains(filename, "mzml") == true) {
+        parseMzML(filename.c_str());
+    } else if (contains(filename, "cdf") == true) {
+        parseCDF(filename.c_str(), 1);
```
saifulbkhan, line 130: The expression `contains(string, subString)` itself evaluates to a boolean primitive and would satisfy the conditional. You need not check its result against `true` - that is redundant logic. In fact it is more readable, without being checked against true.

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,215 @@
...
 #include <boost/iostreams/filtering_streambuf.hpp>
+#include <boost/algorithm/string.hpp>
 
 /**
  * random collection of useful functions 
  */
 namespace mzUtils {
 
-    int randInt(int low, int high){
+    const long double   Pi = 3.1415926535897932384626433832795028841968;
```
saifulbkhan, line 17: Do we not get a `Pi` constant from the standard math library? We get `M_PI` from the cmath library, that's for sure.

src/core/libmaven/mzUtils.cpp
```diff
@@ -7,274 +7,215 @@
...
+    vector<string> split(const string& str, const string& sep)
+    {
+        vector<string> result;
+        boost::split(result, str, boost::is_any_of(sep));
+        return result;
     }
 
-    void removeSpecialcharFromStartEnd(vector<string>& fields) {
-        for(unsigned int i=0; i < fields.size(); i++ ) {
+    void removeSpecialcharFromStartEnd(vector<string>& fields)
```
saifulbkhan, line 64: Capitalize "c" for `Char` in the function name.

src/core/libmaven/mzUtils.cpp
```diff
@@ -593,93 +426,90 @@ Series:  Prentice-Hall Series in Automatic Computation
...
         float min_s = 0;
         float minR = 1e99;
 
         //find best fit
-        if (ycoord.size()<3) return;
-        vector<float>yobs=ycoord;
-        int ysize=yobs.size();
-        int midpoint  = int(ysize/2);
+        if (ycoord.size()<3){
+            pair<float, float> res = make_pair(FLT_MAX, FLT_MAX);
```
saifulbkhan, line 451: Can we use `numeric_limits<float>::max()` for the maximum allowed value for floating point values. `FLT_MAX` is a C thing. Also use `numeric_limits` whenever you need to obtain the min/max/lowest representable values for any numeric type.

src/core/libmaven/mzUtils.cpp
```diff
@@ -593,93 +426,90 @@ Series:  Prentice-Hall Series in Automatic Computation
...
-           cerr << "fitting yobs:" << endl; for(int i=0; i < ysize; i++ )
-           cerr << setprecision(2) << yobs[i] << ", "; cerr << endl;
-           */
-
         bool converged = false;
         int ittr = 0;
 
-        if (greaterZeroCount <= 3 ) return;
+        if (greaterZeroCount <= 3 ){
+            pair<float, float> res = make_pair(FLT_MAX, FLT_MAX);
```
saifulbkhan, line 479: Same thing here regarding`FLT_MAX` again.

src/core/libmaven/mzUtils.cpp
```diff
@@ -593,93 +426,90 @@ Series:  Prentice-Hall Series in Automatic Computation
...
+            return res;
+        }
         while (!converged ) {
             if ( ittr++ > 20 ) break;
-            float Rsqr=0;
-            for(int i=0; i < ysize; i++ )  { Rsqr += POW2(exp(-0.5*POW2(x[i]/s)) - yobs[i]); }
-            //       cerr << "\t\ts=" << s << " Rsqr=" << Rsqr << endl;
+            float Rsqr = 0;
+            for(int i = 0; i < ysize; i++ ){
+                Rsqr += POW2(exp(-0.5*POW2(x[i]/s)) - yobs[i]);
```
saifulbkhan, line 486: Use C++ methods `pow` and `powf`for calculating exponents.

src/core/libmaven/mzUtils.cpp
saifulbkhan: No problem.

src/core/libmaven/mzUtils.cpp
sakshikukreja14: Saiful, POW2 is a macro defined in mzUtils that does the square of a number.

src/core/libmaven/mzUtils.cpp
```diff
@@ -593,93 +424,92 @@ Series:  Prentice-Hall Series in Automatic Computation
...
-           cerr << setprecision(2) << yobs[i] << ", "; cerr << endl;
-           */
-
         bool converged = false;
         int ittr = 0;
 
-        if (greaterZeroCount <= 3 ) return;
+        if (greaterZeroCount <= 3 ){
+            pair<float, float> res = make_pair(FLT_MAX,
+                                               FLT_MAX);
```
saifulbkhan, line 479: I see you have changed the above lines to use `numeric_limits` but not here. In general, most suggestions from reviewers should apply to the whole change-set, unless it is something specific. So, what I usually do when I get a review suggestion, is that I would look for other instances (that were not addressed by the reviewer) where that suggestion can be applied and fix those too.

We can let this one be as it is, but you can make this a practice in future PRs.

PR #1226, opened 2020-02-10
[fix] Fixed crashing test case in "mzAligner"

sakshikukreja14:


src/core/libmaven/mzAligner.h
```diff
@@ -63,7 +64,8 @@ class Aligner {
      * @param sampleName Name of the sample associated with this segment.
      * @param seg Pointer to the AlignmentSegment to be added.
      */
-    void addSegment(string sampleName, AlignmentSegment* seg);
+    void addSegment(string sampleName, AlignmentSegment seg,
+                    map<string,vector<AlignmentSegment>>& alignmentSegment_private);
```
saifulbkhan: Two things here:
1. Update the documentation for this method.
2. Please learn to align method's parameters like:
  ```C++
  void addSegment(string sampleName,
                  AlignmentSegment seg,
                  map<string, vector<AlignmentSegment>>& alignmentSegment_private);
  ```

PR #1235, opened 2020-02-17
[UI] Allow filtering unfragmented groups using independent filter

saifulbkhan:
This PR is for the following two changes:
1. Move non-MS2 exclusion check to fragmentation options: this will allow us to perform peak-detection with fragmentation matching that only works as a whether-to-score-or-not flag. If elimination of unfragmented groups is desired, the "Exclude unfragmented groups" option will be available under "Match Fragmentation" options, independent of the mode of detection.
2. Prohibit closing dialog when detection in progress: this will make sure that no-interaction is possible while peak-detection is ongoing, which otherwise leads to confusion.
3. Fix - fragmentation events disappearing from EIC widget when tolerance switch is toggled on.

### DIFF
--- a/src/core/libmaven/groupFiltering.cpp
+++ b/src/core/libmaven/groupFiltering.cpp
@@ -39,8 +39,13 @@ void GroupFiltering::filter(vector<PeakGroup> &peakgroups)
             continue;
         }
 
+        if (_mavenParameters->mustHaveFragmentation
+            && peakgroups[i].ms2EventCount == 0) {
+            peakgroups.erase(peakgroups.begin() + i);
+            continue;
+        }
+
         i++;
-        
     }
 
 }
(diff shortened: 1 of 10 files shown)


src/core/libmaven/groupFiltering.cpp
```diff
@@ -98,12 +103,14 @@ bool GroupFiltering::filterByMS1(PeakGroup &peakgroup)
 
 bool GroupFiltering::filterByMS2(PeakGroup& peakgroup)
 {
+    if (peakgroup.ms2EventCount == 0)
```
shubhra-agrawal, line 106: I would expect `filterByMS2` to return `true` if it was not successful in matching the pattern, regardless of the presence/absence of fragmentation data. thoughts?

saifulbkhan: As a standalone function in itself, yes. I agree. I will change the logic and the calls to the function accordingly.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -502,6 +511,26 @@ void PeakDetectionDialog::_setAdductWindowState()
...
+        featureOptions->setDisabled(true);
+        dbSearch->setDisabled(true);
+        matchFragmentationOptions->setDisabled(true);
+        peakScoringOptions->setDisabled(true);
+        computeButton->setDisabled(true);
+        resetButton->setDisabled(true);
+    } else {
+        featureOptions->setEnabled(true);
+        dbSearch->setEnabled(true);
+        matchFragmentationOptions->setEnabled(true);
```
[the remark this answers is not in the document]
saifulbkhan: This method disables/enables almost every UI element, depending on whether peak detection is currently in progress or not. If enabled, subsequent calls (detection mode, which DB selected, etc.), will then decide the state for dependent elements. I have compiled and rechecked as well.

PR #1241, opened 2020-02-26
Alpha fixes

saifulbkhan:
Several fixes for v0.10.0-alpha. Each fix is an individual commit.

### DIFF
--- a/src/gui/mzroll/mzfileio.cpp
+++ b/src/gui/mzroll/mzfileio.cpp
@@ -1057,15 +1057,15 @@ void mzFileIO::_readPeakTablesFromSQLiteProject(const vector<mzSample*> newSampl
             group->setAdduct(DB.findAdductByName(group->getAdduct()->getName()));
 
         // assign group to bookmark table if none exists
-        if (group->searchTableName.empty())
-            group->searchTableName = "Bookmark Table";
+        if (group->tableName().empty())
+            group->setTableName("Bookmark Table");
 
         // find appropriate tables and populate them
         auto allTablesList = _mainwindow->getPeakTableList();
         allTablesList.push_back(_mainwindow->bookmarkedPeaks);
         TableDockWidget* table = nullptr;
         for (auto t : allTablesList)
-            if (t->windowTitle().toStdString() == group->searchTableName)
+            if (t->windowTitle().toStdString() == group->tableName())
                 table = t;
 
         if (table)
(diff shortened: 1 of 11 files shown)


src/gui/mzroll/mzfileio.cpp
```diff
@@ -1057,15 +1057,15 @@ void mzFileIO::_readPeakTablesFromSQLiteProject(const vector<mzSample*> newSampl
             group->setAdduct(DB.findAdductByName(group->getAdduct()->getName()));
 
         // assign group to bookmark table if none exists
-        if (group->searchTableName.empty())
-            group->searchTableName = "Bookmark Table";
+        if (group->tableName().empty())
+            group->setTableName("Bookmark Table");
```
shubhra-agrawal: Make sure that the table name is updated when the bookmarked group is merged to any peak table.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -660,7 +660,7 @@ void TableDockWidget::exportGroupsToSpreadsheet() {
                               return group.getCompound()->type() == Compound::Type::PRM;
                             });
   bool prmGroupExists = prmGroupAt != end(allgroups);
-  bool includeSetNamesLines = true;
+  bool includeSetNamesLines = false; // when should this be true?
```
shubhra-agrawal: This should be true when the user selects the "Group Summary Matrix Format with Set Names" option.

shubhra-agrawal: Let me know if that option is working.

shubhra-agrawal:
I'm not quite sure about the bug or fix in the persistent message box. The rest looks fine; I've added comments where I want you to double-check something.

src/gui/mzroll/tabledockwidget.cpp
saifulbkhan: I see. Thanks for the info. I'll make sure it works.

src/gui/mzroll/mzfileio.cpp
saifulbkhan: Will fix.

saifulbkhan:
@shubhra-agrawal The message box that is supposed to inform the user about ongoing save process, before they exit El-MAVEN, was not staying for the entire duration of the save process. This was because it was stack allocated in a function (`MainWindow::saveProject`) and was destroyed as soon as the function returned. This made El-MAVEN window interact-able, during final save which is something that we do not want. I only made the message box heap allocated so that it lives until the application destroys itself (and therefore cleaning any associated allocations, including this message box).

PR #1245, opened 2020-03-02
[fix] Fixed the issue of no default isotope option in "Isotope Dialog"

sakshikukreja14:
In isotope dialog, there was no default isotope checked. This could be an issue while providing default parameters as no isotope would be detected.
This has been fixed ans isotope "C13" has been made default option.

### DIFF
--- a/src/gui/mzroll/isotopedialog.cpp
+++ b/src/gui/mzroll/isotopedialog.cpp
@@ -68,8 +68,12 @@ IsotopeDialog::IsotopeDialog(MainWindow* parent) : QDialog(parent) {
                         && !N15Labeled_BPE->isChecked()
                         && !S34Labeled_BPE->isChecked())
                         C13Labeled_BPE->setChecked(true);
-                } else
+                } else {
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/isotopedialog.cpp
```diff
@@ -62,6 +62,14 @@ IsotopeDialog::IsotopeDialog(MainWindow* parent) : QDialog(parent) {
...
                                "Isotope Detection Switched",
                                state);
+                if (checked) {
+                    if (!D2Labeled_BPE->isChecked()
+                        && !C13Labeled_BPE->isChecked()
+                        && !N15Labeled_BPE->isChecked()
+                        && !S34Labeled_BPE->isChecked())
+                        C13Labeled_BPE->setChecked(true);
+                } else
+                    C13Labeled_BPE->setChecked(false);
```
saifulbkhan: Maybe we do not need this `else` part at all. Or if we are unchecking C13 checkbox in the `else` condition, then we should uncheck them all. Can you make the change?

PR #1246, opened 2020-03-05
Beta fixes

sakshikukreja14:


src/core/libmaven/adductdetection.cpp
```diff
@@ -42,9 +42,10 @@ AdductDetection::findAdducts(const vector<PeakGroup>& parentIons,
 
         auto compound = parentGroup.getCompound();
         for (const auto adduct : adductsList) {
-            if (*adduct == *(parentGroup.getAdduct()))
-                continue;
-
+            if(parentGroup.getAdduct()){
```
saifulbkhan: Please make a habit of putting spaces before and after braces (unless its braces enclosing a function's parameters).

sakshikukreja14: Okay, I will take care of this

src/gui/mzroll/gallerywidget.cpp
```diff
@@ -329,56 +329,60 @@ void GalleryWidget::wheelEvent(QWheelEvent *event) {
 		}
```
saifulbkhan: Did we make any programmatic change in this file at all? If not please try to not put code formatting changes in the same commit as those that are supposed to serve as fixes or enhancements (such as this one).

saifulbkhan: Oh and the reason to not do this is, that it makes it difficult for me to review these commits. It gives me the impression that some logic was changed and I would go over it without realizing that its just a stylistic change.

sakshikukreja14: Okkay, I may have added it because of some printing statements i used to debug. But surely, will take care in future

PR #1248, opened 2020-03-16
Beta fixes

sakshikukreja14:


src/gui/mzroll/adductwidget.cpp
```diff
@@ -80,8 +80,6 @@ void AdductWidget::selectAdductsForCurrentPolarity()
             // to be unselected
             (*it)->setCheckState(0, Qt::Checked);
             (*it)->setDisabled(true);
-        } else {
-            (*it)->setDisabled(false);
```
saifulbkhan: This check was supposed to re-enable the checkbox for parent adduct rows ([M+H]+ and [M-H]-), if the current polarity was switched to the other type. So, if negative polarity samples are loaded, we want the [M-H]- adduct to be selected by default and remain un-checkable by disabling it. But if these negative samples are unloaded (for e.g., by deleting all samples from sample widget), and then positive samples are loaded again in that same session, we want [M-H]- to get re-enabled and now make [M+H]+ to be selected by default and un-checkable.

Therefore, there should be an `else if` like:
```C++
else if (SIGN(adduct->getCharge()) != SIGN(_mw->getIonizationMode())
         && adduct->isParent()
         && !_mw->getVisibleSamples().empty()) {
    // re-enable and uncheck the item (*it)
}
```

Issue #1249, opened 2020-03-23
Sample order for OBI-Warp dropdown should be sorted.


sakshikukreja14:
Currently, in 'Align' dialog, the OBI-Warp reference samples appear to be in random fashion. 

Expected Outcome-
The list of reference samples must now appear in lexicographical order.

PR #1250, opened 2020-03-23
[fix] Fixed the order of reference samples in OBI-Warp in "Align" dialog #1249

sakshikukreja14:
The list reference samples in 'Align' dialog for OBI-Warp appeared in random fashion. It has now been modified to appear in lexicographical order.

### DIFF
--- a/src/gui/mzroll/alignmentdialog.cpp
+++ b/src/gui/mzroll/alignmentdialog.cpp
@@ -143,11 +143,18 @@ void AlignmentDialog::intialSetup()
     samplesBox->clear();
     samplesBox->addItem("Select Randomly",
 						QVariant(QVariant::fromValue(static_cast<void*>(nullptr))));
+    vector<mzSample*> selectedSample;
     for(auto sample: _mw->samples) {
         if(sample->isSelected)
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/alignmentdialog.cpp
```diff
@@ -143,11 +143,18 @@ void AlignmentDialog::intialSetup()
     samplesBox->clear();
     samplesBox->addItem("Select Randomly",
 						QVariant(QVariant::fromValue(static_cast<void*>(nullptr))));
+    vector<mzSample*> selectedSample;
     for(auto sample: _mw->samples) {
         if(sample->isSelected)
-            samplesBox->addItem(sample->sampleName.c_str(),
-								QVariant(QVariant::fromValue(static_cast<void*>(sample))));
+            selectedSample.push_back(sample);
```
saifulbkhan, line 149: There is a method in `MainWindow` called `getVisibleSamples`. We can use that here.

saifulbkhan: Also name the variable `selectedSamples` <- plural.

PR #1258, opened 2020-03-25
[fix] Fix the issue where peak bubbles are floating when alignment is performed after peak detection

sakshikukreja14:
When alignment of sample is performed after detecting the peaks, the value of peak groups do not change accordingly. This leads to the peak bubble left floating in the area.
This has been fixed by updating RT value for peaks.

### DIFF
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -1702,16 +1702,6 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
     emit groupSet(group);
     replot(group);
     _clearEicPoints();
-    // Done to update group in case of alignment performed
-    // after peak detection.
-    for (size_t i = 0; i < group->peaks.size(); i++) {
-        auto peak = group->peaks[i];
-        for (auto eic : eicParameters->eics) {
-            if (eic->sample == peak.getSample()) {
-                group->peaks[i].rt = eic->rt[peak.pos];
-            }
-        }
-    }
 	addPeakPositions(group);
 }
 
--- a/src/gui/mzroll/eicwidget.cpp
+++ b/src/gui/mzroll/eicwidget.cpp
@@ -1702,6 +1702,16 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
     emit groupSet(group);
     replot(group);
     _clearEicPoints();
+    // Done to update group in case of alignment performed
+    // after peak detection.
+    for (size_t i = 0; i < group->peaks.size(); i++) {
+        auto peak = group->peaks[i];
+        for (auto eic : eicParameters->eics) {
+            if (eic->sample == peak.getSample()) {
+                group->peaks[i].rt = eic->rt[peak.pos];
+            }
+        }
+    }
 	addPeakPositions(group);
 }
 
--- a/src/gui/mzroll/mainwindow.h
+++ b/src/gui/mzroll/mainwindow.h
@@ -439,9 +439,12 @@ public Q_SLOTS:
     void promptUpdate(QString version);
 
     /**
-     * @brief updateTablePostAlignment Updating peak tables after alignment.
+     * @brief updateTablePostAlignment Update the attributes of peaks for every
+     * group present in all current peak-tables. This method should be called
+     * after operations such as alignment, where the scan's RT changes from its
+     * original value.
      */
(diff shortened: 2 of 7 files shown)


src/gui/mzroll/eicwidget.cpp
```diff
@@ -1727,6 +1727,16 @@ void EicWidget::setPeakGroup(PeakGroup* group) {
...
     replot(group);
     _clearEicPoints();
+    // Done to update group in case of alignment performed
+    // after peak detection.
+    for (size_t i = 0; i < group->peaks.size(); i++) {
+        auto peak = group->peaks[i];
+        for (auto eic : eicParameters->eics) {
+            if (eic->sample == peak.getSample()) {
+                group->peaks[i].rt = eic->rt[peak.pos];
+            }
```
saifulbkhan, line 1737: This is a nice attempt to get rid of floating bubbles. But the fix is only visual. If I export one of these groups (whose peak RTs were updated after alignment), here's a snapshot of the peak-detailed report:

Screenshot 2020-05-06 at 2 00 31 PM

A peak's `rt` should be a value between its `rtmin` and `rtmax`. I have highlighted a few cases in the screenshot, where you can see that it is not.

My point is that we should aim to actually update every attribute of every peak of every peak-group in every peak-table, just after an alignment operation. I understand all of this can be confusing to someone who has not worked on these data structures for a long time. So, let me know if you need some help with this.

src/gui/mzroll/mainwindow.h
```diff
@@ -438,6 +438,11 @@ public Q_SLOTS:
      */
     void promptUpdate(QString version);
 
+    /**
+     * @brief updateTablePostAlignment Updating peak tables after alignment.
+     */
```
saifulbkhan, line 443: Here's a better docstring: "Update the attributes of peaks for every group present in all current peak-tables. This method should be called after operations such as alignment, where the scan's RT changes from its original value."

Try to keep a balance between being descriptive and being precise, when writing documentation.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -136,6 +138,29 @@ void TableDockWidget::sortBy(int col) {
   treeWidget->sortByColumn(col, Qt::AscendingOrder);
 }
 
+void TableDockWidget::updateTableAfterAlignment()
+{
+    for(size_t i = 0; i < allgroups.size(); i++)
```
saifulbkhan, line 143: Use range based loop with references:
```C++
for (PeakGroup& group : allgroups)
    ...
```

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -136,6 +138,29 @@ void TableDockWidget::sortBy(int col) {
...
+    {
+        auto slice = allgroups[i].getSlice();
+        auto samples = _mainwindow->getVisibleSamples();
+        auto eics  = peakDetector->pullEICs(&slice,
+                                           samples,
+                                           _mainwindow->mavenParameters);
+        for(auto eic : eics)
+        {
+            for(size_t j = 0; j < allgroups[i].peaks.size(); j++)
```
saifulbkhan, line 152: Same - use range-based loop, unless you need the index for other purposes.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -62,6 +63,7 @@ TableDockWidget::TableDockWidget(MainWindow *mw) {
   _targetedGroups = 0;
   pal = palette();
   setAutoFillBackground(true);
+  peakDetector = new PeakDetector(_mainwindow->mavenParameters);
```
saifulbkhan, line 66: I do not think that a `PeakDetector` instance should be owned by `TableDockWidget`. In general avoid using algorithmic logic (or data structures meant for similar purposes), in view-only widgets.

What you can do is, add the update function to `BackgroundPeakUpdate` class and maybe you can make it `static` so that you don't need an instance of this thread subclass:
```C++
static updateGroup(vector<PeakGroup*> groups, vector<mzSample*> samples)
{
    PeakDetector detector;
    for (auto& group : groups) {
        auto eics  = peakDetector->pullEICs(&(group->getSlice()),
                                            samples,
                                            _mainwindow->mavenParameters);
        ... // update peaks
    }
    group->groupStatistic();
}
```

PR #1260, opened 2020-03-26
[fix] Fixed the issue of missing peak detection for isotopologues

sakshikukreja14:
In case of C13 isotopic detection, some EIC were represented on the graph which do not have the corresponding peaks. That gave the user the view of missing peaks.
This has been corrected by adding a check that only those sample's  EIC will be represented if there is a corresponding peak for it.


saifulbkhan:
@sakshikukreja14 I am closing this PR since the correct fix for this issue is covered in PR #1272. Although, we are not sure that this was an issue to begin with, or maybe just a matter of improper setting configuration.

PR #1264, opened 2020-04-01
[ref] Refactored the 'DeleteGroups' for reducing time for deletion in large peak tables

sakshikukreja14:
'DeleteGroups' function of tabledockwidget class has been refactored to take less time as compared to original function especially in case where the peak table are larger. This time complexity is now reduced to less than half of the original time taken.

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1230,6 +1230,8 @@ void TableDockWidget::deleteSelectedItems()
             delete(item);
     }
 
+    showAllGroups();
+
     mzUtils::stopTimer(timerMain, "Deletion ends");
     return;
 }
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1119,8 +1119,6 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
 
 void TableDockWidget::deleteSelectedItems()
 {
-    auto timerMain = mzUtils::startTimer();
-
     // temporarily disconnect selection trigger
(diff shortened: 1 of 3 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1113,82 +1113,123 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
   updateCompoundWidget();
 }
 
-void TableDockWidget::deleteGroups() {
-
-  QList<PeakGroup *> selectedGroups;
-  QTreeWidgetItem *nextItem;
+void TableDockWidget::deleteSelectedItems()
+{
+    auto timerMain = mzUtils::startTimer();
```
saifulbkhan, line 1118: Let's lose the timer logic. This was just for debugging.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1113,82 +1113,123 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
...
+
+    // reconnect selection trigger
+    connect(treeWidget,
+            SIGNAL(itemSelectionChanged()),
+            this,
+            SLOT(showSelectedGroup()));
+
+    if(nextItem){
+        treeWidget->setCurrentItem(nextItem);
+    }
```
saifulbkhan, line 1222: Does this work? I deleted groups in many different ways but never got a "next selection" by default. No items were highlighted and the EIC still showed me the last selected group (which was now deleted).

saifulbkhan:
Oh and I did get a crash too after a deletion, but I was not able to reproduce it doing the same things. The backtrace was also not related to deletion logic. So we can ignore it for now. Just FYI.

Issue #1271, opened 2020-04-11
Link isotopologue peak integration x-ranges


saifulbkhan:
Quoting Sam De Craemer from KU Leuven:
> For working with isotopologue data, there should be an option to link peak integration so that the integration area has the same x-axis range for all isotopologues of a compound.
> ...
> We do believe that the boundaries should be the same. Retention time is wholly based on chemistry, so isotopologues of the same compound who are slightly different in mass but chemically identical do elute during the exact same time window. Usually a significantly less abundant isotope will simple have a narrower peak because of chance. Less molecules elute at the edges of the time window, so an uncommon isotopologue is unlikely to be detected at the edge.

> In fact, we noted some inconsistencies in isotopologue abundances between samples that were replicates of the same experimental conditions in El-Maven but not our manual integration. The reason was that in some replicates the isotopologue did look like one nice continuous peak, while in others due to chance it looked more like two separate peaks, as in the image below. Forcing integration over the same range gives consistent results in El-Maven as well.

We can add a checkbox in the isotope detection dialog that, if checked, will mandate the integration for isotopologues over the exact same RT range as the parent ion.

saifulbkhan:
Shipped as a part of v0.11.0. Closing.

Issue #1255, opened 2020-03-23
Missing peak detection for isotopologues


624284916:
**Describe the bug**
I was working with 13C labeled LC-MS data in El-MAVEN (v0.10.0). To discover all isotopic peaks when the parent peak abundance is 0, we imported one unlabeled sample together with three labeled samples. After performing peak detection according to a defined database, all good peaks were shown in a bookmark table. The problem we saw from the table results were that isotopic peaks were not correctly detected. There were missing (unsolid) circles assigned for some existing peaks. It happens to most metabolites. We have tried changing related parameters: distX Weight, distY Weight (value: 1, 2, 6), match retention time (value: 0.25, 0.5), isotope is within [x] scans of parent (value: 1, 5). The problem couldn't be fixed

**To Reproduce**
1. Go to Align, then peaks
2. Click on 'find peaks'
3. See error
  1) parent peak deetction
![parent peak deetction](https://user-images.githubusercontent.com/60973150/77311628-c6259200-6d00-11ea-9043-acc63ffbcaa3.PNG)

  2) 13C-label-3 peak detection
![13C-label-3 peak detection](https://user-images.githubusercontent.com/60973150/77311629-c6be2880-6d00-11ea-8807-3a6e942ff8f4.PNG)

  3) 13C-label-3 peak detection missing in the second sample
![13C-label-3 peak detection missing in the second sample](https://user-images.githubusercontent.com/60973150/77311631-c756bf00-6d00-11ea-9630-4f5cc295e699.PNG)

**System:**
 - OS: [e.g. Windows 10]
 - Version [v0.10.0]

saifulbkhan:
Hi @624284916. Thanks for reporting. I was able to reproduce this just now. We will definitely look into this and let you know when the fix is available.

rmtfleming:
@saifulbkhan Have you been able to resolve this issue? We are considering now whether we can continue with ElMaven or must switch to another solution so we need to make a decision.

saifulbkhan:
@rmtfleming This issue is part of the v0.11.0 plan. The tentative date for this release is May 15th. We might be able to share an internal build with this issue fixed, sooner, but with such builds there's a high chance of encountering bugs and other stability issues. Please allow us the time to properly resolve this.

chubukov:
@saifulbkhan any chance you could provide a short description of the underlying cause of the bug? it would be great to evaluate whether we expect it to affect our workflow as well

[...]

PR #1272, opened 2020-04-13
[Ftr] Add an option to link isotope-parent RT range #1271 #1255

saifulbkhan:
As described in #1271, a new option has been added that will allow tying the bounds of each non-parent isotopologue peak to that of its parent peak. Additionally, some restructuring and fixes were made to the isotope detection methods that should help with the issue described in #1255.


sakshikukreja14:
I tried to compare this branch with the latest version of El-Maven. So, what's happening is there are some isotopes that are not being detected. And secondly, there are cases when the EIC is plotted but no peak corresponding is shown marked. 
This is happening even after trying different settings. 
(Right image is the latest version and left one is the compiled branch) 
Screenshot 2020-05-11 at 3 02 45 PM

saifulbkhan:
@sakshikukreja14 Pushed a commit fixing the issue mentioned above. MS/MS scans were being counted when calculating distance from peaks and such. Please let me know if you still face this issue.

Also, I noticed that isotope detection is now noticeably (10x to 15x) slower compared to previous releases. But then again, the earlier techniques were not properly implemented, even though they used to give mostly decent results.

sakshikukreja14:
> @sakshikukreja14 Pushed a commit fixing the issue mentioned above. MS/MS scans were being counted when calculating distance from peaks and such. Please let me know if you still face this issue.
> 
> Also, I noticed that isotope detection is now noticeably (10x to 15x) slower compared to previous releases. But then again, the earlier techniques were not properly implemented, even though they used to give mostly decent results.



The issue of detecting peak is solved.. The only concern which is left is that in same image(right) I posted earlier there is C-13 label 18 which is still missing from this branch.. else all other thing is working

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -63,207 +63,182 @@ void IsotopeDetection::pullIsotopes(PeakGroup* parentgroup)
...
-            if (isotopePeakIntensity == 0 || rt == 0) continue;
+            // even though we will do a peak detection on the isotope's slice
+            // later on, this is a good fast check to see if there is anything
+            // worth searching for; maybe also for adduct & fragment detection?
+            Scan* scan = parentPeak->getScan();
+            std::pair<float, float> topMatch = getIntensity(scan, mzmin, mzmax);
+            float isotopePeakIntensity = topMatch.first;
+            float rt = topMatch.second;
 
-            if (filterIsotope(x, isotopePeakIntensity, parentPeakIntensity, sample, parentgroup))
```
shubhra-agrawal, line 108: Would suggest moving the correlation check to this point? Considering you have the isotope's mzmin and mzmax as well as the parent peak's RT range already. The speed might not take as much of a hit. I think the same correlation calculation is happening multiple times right now.

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -63,207 +63,182 @@ void IsotopeDetection::pullIsotopes(PeakGroup* parentgroup)
...
 {
-    float highestIntensity = 0;
-    float rt = 0;
+    float highestIntensity = 0.0f;
+    float rt = 0.0f;
     mzSample* sample = scan->getSample();
 
-    for (int i = scan->scannum - 2; i < scan->scannum + 2; i++) {
+    for (int i = scan->scannum - _mavenParameters->maxIsotopeScanDiff;
+         i < scan->scannum + _mavenParameters->maxIsotopeScanDiff;
```
shubhra-agrawal, line 220: `<=` ? I, as a user would expect +-maxIsotopeScanDiff around the peak, but it's not a big deal. Your call.

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -63,207 +63,182 @@ void IsotopeDetection::pullIsotopes(PeakGroup* parentgroup)
...
+         i++) {
         Scan* s = sample->getScan(i);
 
         // Filter out MS2 scans when obtaining isotope peak intensities.
         auto pos = i;
-        while(s->mslevel > 1 && i < scan->scannum) {
-            auto index = i < scan->scannum ? --pos
-                                           : ++pos;
+        while(s->mslevel > 1) {
```
shubhra-agrawal, line 226: If `++pos` is executed even once, you'll be pulling the same scan multiple times since the value of `pos` is not retained in the next iteration.

I'd suggest going backward and forward from `scan` to find relevant x MS1 scans and then running this whole `for` loop.

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -63,207 +63,182 @@ void IsotopeDetection::pullIsotopes(PeakGroup* parentgroup)
...
+            if (nearestPeak == nullptr) {
+                delete eic;
+                continue;
+            }
+
+            if (_mavenParameters->linkIsotopeRtRange) {
+                eic->adjustPeakBounds(*nearestPeak,
+                                      parentPeak->rtmin,
+                                      parentPeak->rtmax);
+                eic->getPeakDetails(*nearestPeak);
```
shubhra-agrawal, line 157: peakFiltering should happen here instead? since the baseline and area have both changed now.

src/core/libmaven/isotopeDetection.cpp
```diff
@@ -63,207 +63,182 @@ void IsotopeDetection::pullIsotopes(PeakGroup* parentgroup)
...
+            if (isotopePeakIntensity == 0.0f || rt == 0.0f)
                 continue;
 
-            vector<Peak> allPeaks;
-
-            EIC * eic = sample->getEIC(mzmin, mzmax, sample->minRt,sample->maxRt, 1, _mavenParameters->eicType,
-                                        _mavenParameters->filterline);
-            //actually last parameter should probably be deepest MS level?
-            //TODO: decide how isotope children should even work in MS mode
+            float rtmin = parentGroup->getSlice().rtmin;
```
shubhra-agrawal, line 97: I would expect the rtmin and max to be the parent peak's rtmin and max, instead of the group range to prevent any issues from minor RT drift across samples. It might also reduce the number of peaks we get per EIC since the range will be narrower. Let me know what you think.

src/core/libmaven/isotopeDetection.cpp
saifulbkhan: Good point. Will shift.

src/core/libmaven/isotopeDetection.cpp
saifulbkhan: Yea, agree. Will change.

src/core/libmaven/isotopeDetection.cpp
saifulbkhan: Right. But what if the nearest peak is filtered and there was another (second-nearest) peak that was worthy of being the isotope's peak but it was earlier filtered due to distance? Very unlikely, but possible. Instead, while finding nearest peak, we change each peak's boundaries, apply peak filter on it, and only when it passes will its distance be considered.

src/core/libmaven/isotopeDetection.cpp
saifulbkhan: No, `<=` makes more sense.

src/core/libmaven/isotopeDetection.cpp
saifulbkhan: Good catch. It took me a while to understand you were talking about the outer iteration.

shubhra-agrawal: Oh, yeah. Picked the wrong line, sorry.

src/core/libmaven/isotopeDetection.cpp
shubhra-agrawal: that makes sense

saifulbkhan:
@sakshikukreja14 Can you check if any of the missing labels actually have peaks resembling the respective parent peak? I compared this branch's results against that of v0.10.0 and all the missing, were filtered for valid reasons. Especially, the new full-peak correlation check was responsible for the elimination of many noisy signals.

@shubhra-agrawal Made changes as suggested. The "correlation before peak finding" thing worked really well! The runtime is now very comparable to previous releases. I did not even notice a difference.

shubhra-agrawal:
Looks good! Hopefully, a lot of noise in isotope detection will be filtered out after this change. 

During testing, please take special care to match the results in peak table with the isotope widget as well as isotope plot. Old emDB versions should be thoroughly tested as well.

Issue #1273, opened 2020-04-14
Some chromatograms from MRM data are not displaying correctly


saifulbkhan:
For some transitions the EIC shown, contains alternating zero values in the peak signal (first image). The same transition is displayed differently when browsed from the SRM list (second image). And even the second chromatogram's signal is not smooth. These peaks will probably not be quantified correctly either.

**Expected behavior**
All EICs should show the actual signal encoded in the sample file. The true nature of these signals was verified using the SeeMS tool (part of the ProteoWizard package).

**Screenshots**
EIC for a particular transition:
<img width="1392" alt="Screenshot 2020-04-14 at 7 19 46 AM" src="https://user-images.githubusercontent.com/10397162/79177489-898d1800-7e20-11ea-8d71-8ebfb6c07249.png">

EIC for that same transition while browsing through the SRM list widget:
<img width="1392" alt="Screenshot 2020-04-14 at 7 20 47 AM" src="https://user-images.githubusercontent.com/10397162/79177518-9f024200-7e20-11ea-9570-9a5a84c31021.png">

**System:**
 - OS: All operating systems
- Version: v0.10.0

saifulbkhan:
Shipped as part of v0.11.0. Closing.

Issue #358, opened 2017-09-02
Sample number not showing on popped up window for a particular sample 


sunil20dhakad:
In the sample list, I am looking for only particular sample numbers but I am not able to get the information of sample number inside the ELMaven. 

We get other information like m/z range, rt range, scan #, ionization, filename information for each sample present but no information regarding sample number.
We can have an extra column inside sample widget for sample number.
ElMaven version # El_Maven_v0.1.4-18-g654cfb1

saifulbkhan:
Additionally, I see `-1` values for "sample number", when I hover over a peak in the EIC.

saifulbkhan:
There is now a separate column for sample numbers in the samples widget, in v0.10.0.

PR #1274, opened 2020-04-14
[Fix] Filter MRM data scans using productMz and CE #1273

saifulbkhan:
Data for MRM experiments is encoded as "chromatograms" within mzML files. These chromatograms are unique at least for a combination of precursor m/z, product m/z and collision energy. While creating "EIC" objects for these chromatograms, the scans beind iterated over were not correctly checking for product m/z and collision energy (the code was there but commented out). This has been fixed.

Additionally, SRM IDs can share the same string (once stripped of their key-value pairs) which led to two chromatograms being merged into one. This has also been fixed by adding collision energy data beside each string. This should be enough to maintain uniqueness amond SRM ID strings.

### DIFF
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -337,7 +337,6 @@ void mzSample::parseMzMLChromatogramList(const xml_node& chromatogramList)
         string chromatogramId = chromatogram.attribute("id").value();
         int sampleNo = getSampleNoChromatogram(chromatogramId);
 
-        // why is this cleaning needed?
         cleanFilterLine(chromatogramId);
 
         vector<float> timeVector;
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -1158,6 +1158,14 @@ EIC* mzSample::getEIC(float precursorMz,
                 if (scan->intensity[k] > eicIntensity) {
                     eicIntensity = scan->intensity[k];
                     eicMz = scan->mz[k];
+
+                    // We set the filterline to be the SRM ID of the first scan
+                    // that matches, so that all subsequent observations also
+                    // originate from the same SRM ID. Since, there could be
+                    // multiple SRM chromatograms that satisfy the m/z checks
+                    // above, but in the absence of CE, their observations may
+                    // not be differentiable.
+                    filterline = scan->filterLine;
                 }
             }
             break;
@@ -1176,6 +1184,7 @@ EIC* mzSample::getEIC(float precursorMz,
             if (sumIntensity != 0.0) {
                 eicMz = static_cast<float>(sumMz / sumIntensity);
                 eicIntensity = static_cast<float>(sumIntensity);
+                filterline = scan->filterLine;
             }
             break;
         }
@@ -1185,6 +1194,7 @@ EIC* mzSample::getEIC(float precursorMz,
                 if (scan->intensity[k] > eicIntensity) {
                     eicIntensity = scan->intensity[k];
                     eicMz = scan->mz[k];
+                    filterline = scan->filterLine;
                 }
             }
             break;
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -1144,6 +1144,7 @@ EIC* mzSample::getEIC(float precursorMz,
         if (productMz != 0.0f && abs(scan->productMz - productMz) > amuQ3)
             continue;
         if (collisionEnergy != 0.0f
+            && scan->collisionEnergy != 0.0f
             && abs(scan->collisionEnergy - collisionEnergy) > 0.5) {
             continue;
         }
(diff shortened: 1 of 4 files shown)


src/core/libmaven/mzSample.cpp
```diff
@@ -337,6 +337,7 @@ void mzSample::parseMzMLChromatogramList(const xml_node& chromatogramList)
         string chromatogramId = chromatogram.attribute("id").value();
         int sampleNo = getSampleNoChromatogram(chromatogramId);
 
+        // why is this cleaning needed?
```
shubhra-agrawal, line 340: sometimes the scan number or RT etc is added to the SRM ID. This makes every ID unique and hence, useless as an identifier for scans emerging from a single source.

cleanFilterLine just removes such extraneous data.

src/core/libmaven/mzSample.cpp
```diff
@@ -388,21 +396,21 @@ void mzSample::parseMzMLChromatogramList(const xml_node& chromatogramList)
...
+            // wasteful - maybe we should directly create EICs and store them
+            // within the sample object for MRM data.
             for (unsigned int i = 0; i < timeVector.size(); i++) {
                 Scan* scan = new Scan(
                     this, scannum++, mslevel, timeVector[i], precursorMz, -1);
                 scan->productMz = productMz;
-                scan->mz.push_back(productMz);
-                scan->filterLine = chromatogramId;
+                scan->collisionEnergy = collisionEnergy;
+                scan->filterLine = chromatogramId
```
shubhra-agrawal, line 409: 👍

src/core/libmaven/mzSample.cpp
```diff
@@ -1128,25 +1136,25 @@ EIC* mzSample::getEIC(float precursorMz,
...
+            continue;
+        }
 
         float eicMz = 0;
         float eicIntensity = 0;
 
         switch ((EIC::EicType)eicType) {
         case EIC::MAX: {
             for (unsigned int k = 0; k < scan->nobs(); k++) {
-                if (abs(productMz - scan->mz[k]) > amuQ3)
```
shubhra-agrawal, line 1148: From what I remember, we added this check for every observed mass in a scan to get a fuller EIC for the selected productMz range. The productMz extracted at #358 for a scan is the nominal mass value, I'm not sure how it's calculated. Even if that value does not fall within our search window, the individual values within the scan often do. Let me know what you think.

shubhra-agrawal:
What about "spectrumList"? From what I remember, MRM experiments can also be encoded as "spectrumList" in mzML files. The parsing code is also written with that in mind. Would be better if the CE search can be implemented for both XML structures at the same time.

If this is an urgent feature, then please create a separate issue for spectrumList.

src/core/libmaven/mzSample.cpp
saifulbkhan: Got it. Thanks. Will remove that comment.

src/core/libmaven/mzSample.cpp
saifulbkhan: I tried out mzML sample files from two different MRM experiments. Both had the same kind of encoding (as `chromatogramList`). But since you say that they can also be encoded as `spectrumList`, maybe then the above might be true.

At least for data read from chromatogram lists, each single RT-intensity point gets a whole scan for itself. So that for loop (line no. 1157) is only iterating on one single observation. And therefore whether we check for productMz before or later in the loop does not matter.

saifulbkhan:
> What about "spectrumList"? From what I remember, MRM experiments can also be encoded as "spectrumList" in mzML files. The parsing code is also written with that in mind. Would be better if the CE search can be implemented for both XML structures at the same time.
Thanks for letting me know. I did not know. I will have to look for MRM files like this.

> If this is an urgent feature, then please create a separate issue for spectrumList.
We have time, if I can find sample files, will make the changes in this PR itself.

src/core/libmaven/mzSample.cpp
shubhra-agrawal: Got it. Run a check once with both structures. If the obs->mz is exactly the same as productMz then moving it out of the loop works. I suspect there would be some precision difference in the two values though, even if it's a single observation.

saifulbkhan:
@shubhra-agrawal I was not able to find any datasets (on Metabolights) where mzML samples had data in their `spectrumList` instead of `chromatogramList`. If you know of any, please send them my way. Otherwise, we can take this up later when we have access to such samples. Which is probably when someone runs actually into this issue, and that is fine.

src/core/libmaven/mzSample.cpp
shubhra-agrawal: @saifulbkhan hey, has the precision been verified between productMz and individual obs->mz ?

saifulbkhan: @shubhra-agrawal If you see line no. 406 and line no. 412, the same `productMz` value is being assigned to `scan->productMz` and the scan's (one and only) observation. So our check comparing this method's `productMz` argument against the current scan's `productMz` would be the same as comparing argument with individual observation. Am I making sense? Or did I misunderstand your question?

shubhra-agrawal: @saifulbkhan Oh I had missed the part where it was getting assigned to the scan's observation. Concern resolved 👍

PR #1278, opened 2020-04-15
[fix] Fixed checking match retention time window if corresponding option is checked

sakshikukreja14:
Earlier, match retention time window remained unchecked even if its corresponding option was checked. This was due to setting its check state to false in constructor of Peak detection dialog, irrespective of the state of its option.


src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -203,7 +203,12 @@ PeakDetectionDialog::PeakDetectionDialog(MainWindow* parent) :
         featureOptions->setChecked(false);
         connect(featureOptions, SIGNAL(toggled(bool)), SLOT(featureOptionsClicked()));
 
-        compoundRTWindow->setEnabled(false); //TODO: Sahil - Kiran, Added while merging mainwindow
+        //TODO: Sahil - Kiran, Added while merging mainwindow
```
saifulbkhan, line 206: Let's remove these "TODO: ... added while ..." comments wherever we make changes.

PR #1280, opened 2020-04-15
[fix] Handle carriage-return character as newlines while reading csv file

sakshikukreja14:
While reading csv files, the file which contains carriage return character as new lines must be read successfully. Earlier ElMaven, did not read such files and show failure error. This has now been successfully handled.

### DIFF
--- a/src/gui/mzroll/database.cpp
+++ b/src/gui/mzroll/database.cpp
@@ -895,7 +895,7 @@ int Database::loadCompoundCSVFile(string filename)
             if (!name.empty())
                 id = name;
             invalidRows.push_back(id);
-        }   
+        }
     };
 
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/database.cpp
```diff
@@ -907,9 +896,36 @@ int Database::loadCompoundCSVFile(string filename)
                 id = name;
             invalidRows.push_back(id);
         }
+        sort(compoundsDB.begin(),compoundsDB.end(), Compound::compMass);
```
saifulbkhan, line 899: This call to `sort` should happen once at the end, once all compounds have been read. Right now it will be called everytime this lambda is invoked. So, it gets worse with increasing number of entries in CSV database.

Issue #1259, opened 2020-03-26
nan m/z after untargeted peak picking


chubukov:
After untargeted peak picking, some m/z values are displayed as NaN or NA in the peak table.

If I click on the row, I get some reasonable-looking EIC and m/z value displayed, which leads me to think it's a problem only in the peak table display, or maybe string conversion. I'll note that if I export to CSV, I still get the nan values for medMz, etc.

Version 0.10.0, QExactive positive mode LC-MS.

![image](https://user-images.githubusercontent.com/13080968/77603081-853aa280-6ee5-11ea-9ace-6b46d58ccfe6.png)

chubukov:
Two notes

1. This seems to happen only when EIC type is set to "Sum", not "Max"
2. If I save to emDB and re-load the file in a new session, the m/z values seem to show up normally.

saifulbkhan:
@chubukov Was finally able to reproduce this with the "Sum" EIC type on Windows release builds _only_. This does not affect compiled builds on Windows (older Qt version) or Mac OS builds, at all. This leads me to think that this indeed is some display-level bug in the Qt framework being used by our Windows deployment pipeline. We will check if this gets fixed with one of the newer versions available.

saifulbkhan:
Should be fixed with v0.11.0. Closing.

PR #1288, opened 2020-04-25
[Fix] Zero division error for sum type EIC #1259

saifulbkhan:
When users opt for summed intensity values over a slice's m/z range, for each scan, the m/z assigned per time observation is calculated by weighted-averaging of all m/z values in that range. The weights here are intensities and their sum can be zero for a given time point. However, no check for a zero sum was made which often led to the weighted m/z being evaluated to a "nan" (zero division error). All subsequent calculations using this m/z also resulted in "nan" values. This has been fixed by simply adding the required check.


saifulbkhan:
@shubhra-agrawal Thanks for reviewing.

Issue #1225, opened 2020-02-10
Values in Isotope Widget should depend upon the peak table selected and not the global settings


avijitzutshi:
Currently, the values in the Isotope Widget depend upon the global settings and not individual peak tables. This leads to confusion wherein people shifting from one peak tale to another do not see a change in the values in the isotope widget.

**Expected Outcome**
The values for different parameters within the Isotope Widget should change with the selection of the peak table to avoid this confusion.

saifulbkhan:
Should be available with v0.11.0. Closing.

PR #1293, opened 2020-05-04
[Fix] Consistent isotope distribution for integrated groups #1225

saifulbkhan:
Peak-groups that have already been integrated and added to a table are meant to be frozen in terms of their contents and properties. However, isotope widget always pulls isotopes at current global settings, which the user could have changed after doing detection or manual integration. Selected a group then, would show a different set of isotopes being populated in isotope widget than the ones shown in peak table (under the parent peak-group). This has been fixed and the isotope widget will now show the same set of isotopologues that are actually present in a parent group.

### DIFF
--- a/src/core/libmaven/PeakGroup.cpp
+++ b/src/core/libmaven/PeakGroup.cpp
@@ -77,6 +77,8 @@ PeakGroup::PeakGroup()  {
     _type = GroupType::None;
     _sliceSet = false;
 
+    _tableName = "";
+
     changePValue=0;
     changeFoldRatio=0;
     //children.reserve(0);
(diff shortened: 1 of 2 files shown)


sakshikukreja14:
It is working fine in many cases. But in a few groups, there are less entries of isotopologues in isotope widget than in the peak table. 
Here is one such example, where in peaktable shows 4 isotopes but there are 2 entries in isotope widget. Could you please look into this?
Screenshot 2020-05-07 at 12 22 13 PM

saifulbkhan:
@sakshikukreja14 The isotope widget show's peaks for one sample at a time. It is possible that the peak-group you have selected has peaks in the currently selected sample for only two isotopologues. Can you check if that's the case? Otherwise the issue needs attention.

sakshikukreja14:
yupp.. got it sorry my bad.

PR #1294, opened 2020-05-04
[Fix] Read MoNA libraries correctly

saifulbkhan:
The spectral libraries available at the official "Massbank of North America" [website](https://mona.fiehnlab.ucdavis.edu/downloads), have some differences in attribute names. This made El-MAVEN unable to read them correctly - especially the ionization mode. To deal with this, a few additional string checks have been added.


sakshikukreja14:
It is working correctly. The concern I have is that MoNa spectral libraries have "smile string" as a part of "comments". This smile string is not getting set according to the code as no line in the library is starting with smile. Rather, can we not add a check in "else if" block on line 579 to set smile string according to these libraries.

saifulbkhan:
@sakshikukreja14 Thanks for pointing out. I'll make this change.

Issue #1290, opened 2020-05-02
Allow manually reviewing and editing of peak-group integration


saifulbkhan:
Every once in a while, the inspection and curation of individual peaks is unavoidable (not everything has a nice peak, and isotopologues might be abundant enough for importance but not for a nice peak shape). Therefore, there is a need to allow customization of boundaries of specific compounds in specific samples after bookmarking.

Preferably, it would be possible to adapt them for a selection of samples ranging between one sample to all at the same time.

saifulbkhan:
Shipped as part of v0.11.0. Closing.

PR #1297, opened 2020-05-05
[Ftr] Modify integrated peaks #1290

saifulbkhan:
A new peak-editor widget has been will allow users to edit the peak boundaries of individual or a set of peaks from an already integrated peak-group (manual or automated). This editor will be available from the peak table.

*To-do*:
- [x] Add button and icon for the editor.


sakshikukreja14:
I observed a segment fault while lowering Rt range but could not reproduce it.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1710,22 +1717,6 @@ void TableDockWidget::writeMascotGeneric(QString filename) {
   file.close();
 }
 
-void TableDockWidget::align() {
```
sakshikukreja14, line 1713: Could you please tell me why these functions are removed? Were these not being used or these were the part of that galleryWidget that is not being used now

saifulbkhan: No this was called from and within TableDockWidget itself. Maybe this was some old code that used to work, but it no longer makes sense to me. we are not going to do selective alignment based on just a few groups.

saifulbkhan:
> ... observed a segment fault while lowering Rt range but could not reproduce it.

Was it around the absolute RT bounds? If you have a backtrace, that would help.

sakshikukreja14:
> > ... observed a segment fault while lowering Rt range but could not reproduce it.
> 
> Was it around the absolute RT bounds? If you have a backtrace, that would help.
No, unfortunately it was not on debug mode so, I don't have. I tried reproducing it but couldn't.
 I was trying to make Rt Bound to 0 so, that's where it crashed.

src/gui/mzroll/tabledockwidget.cpp
sakshikukreja14: okay. got it

saifulbkhan:
> No, unfortunately it was not on debug mode so, I don't have. I tried reproducing it but couldn't.
> I was trying to make Rt Bound to 0 so, that's where it crashed.

Hm. I made the range widgets so that their minimum/maximum would be set dynamically based on whichever set of peaks was being displayed. So a zero may not be possible - since each sample has a minimum RT reading, which will be very close to zero but unlikely that it would actually be a zero. Keep an eye out and let me know if this happens again.

src/gui/mzroll/spectrawidget.cpp
```diff
@@ -885,7 +885,7 @@ void SpectraWidget::addAxes()
     	Axes* y = new Axes(1, _minY, _maxY, 10);
     	scene()->addItem(y);
         y->setZValue(999);
-        y->showTicLines(false);
+        y->showTickLines(false);
```
shubhra-agrawal, line 888: possible typo

src/gui/mzroll/gallerywidget.cpp
```diff
@@ -1,477 +1,449 @@
...
+        if (eic == nullptr)
+            continue;
+
+        QColor color = QColor::fromRgbF(eic->sample->color[0],
+                                        eic->sample->color[1],
+                                        eic->sample->color[2],
+                                        0.5);
+
+        Peak* samplePeak = nullptr;
+        for (auto& peak : group->peaks) {
```
shubhra-agrawal, line 123: `Peak* getPeak(mzSample* sample)` method exists for finding the peak by sample name.

src/gui/mzroll/gallerywidget.cpp
```diff
@@ -1,477 +1,449 @@
...
+            if (peak.getSample() == eic->sample) {
```
shubhra-agrawal, line 124: `Peak* getPeak(mzSample* sample)` method exists for fetching peak by sample name. Won't make a difference though.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1600,28 +1603,52 @@ void TableDockWidget::showHeatMap() {
   }
 }
 
+void TableDockWidget::editSelectedPeakGroup()
+{
+  if (treeWidget->selectedItems().size() != 1)
```
shubhra-agrawal, line 1608: Does it count a parent+children selection as 1 selected item or not? Asking here since I'm unable to test atm.

src/gui/mzroll/spectrawidget.cpp
saifulbkhan: I renamed the `Axes` method appropriately, so this is intentional.

src/gui/mzroll/gallerywidget.cpp
saifulbkhan: Oh okay. We should reuse existing methods.

src/gui/mzroll/tabledockwidget.cpp
saifulbkhan: This is a tree-view level check, so I am guessing it has to be at most one item selection. It will be true if one parent item is selected (with or without the children expanded). But I double check just to be sure.

PR #1300, opened 2020-05-11
[fix] Fix segment fault in deletion of group from bookmark table

sakshikukreja14:
Peakgroup when expanded with their isotopes and deleted after isotopes were explored lead to segment fault.


saifulbkhan:
@sakshikukreja14 Can you rebase this on current develop?

PR #1301, opened 2020-05-12
[ref] Add "adductName" to peak file

sakshikukreja14:
PeakFile being generated did not have adduct names in the csv file.


src/core/libmaven/csvreports.cpp
```diff
@@ -399,6 +404,8 @@ void CSVReports::_writePeakInfo(PeakGroup* group)
     // different systems.
     std::sort(group->peaks.begin(), group->peaks.end(), Peak::compSampleName);
 
+
+
```
saifulbkhan, line 408: You really like newlines, hm?

src/core/libmaven/csvreports.cpp
```diff
@@ -417,7 +424,7 @@ void CSVReports::_writePeakInfo(PeakGroup* group)
 
         _reportStream << fixed << setprecision(6) << group->groupId << SEP
                       << compoundName << SEP << compoundID << SEP << formula
-                      << SEP << sampleName << SEP << peak.peakMz << SEP
+                      << SEP << sampleName << SEP <<adductName<< SEP << peak.peakMz << SEP
```
saifulbkhan, line 427: Add spaces before and after operators.

PR #1302, opened 2020-05-14
Alpha fixes

sakshikukreja14:


src/core/libmaven/mzUtils.cpp
```diff
@@ -55,7 +55,13 @@ namespace mzUtils {
...
     {
         vector<string> result;
-        boost::split(result, str, boost::is_any_of(sep));
+        int start = 0;
+        int end = 0;
+        while(end != string::npos){
+            end = str.find(sep, start);
+            result.push_back(str.substr(start, end - start));
+            start = end + sep.length();
+        }
```
sakshikukreja14, line 64: Is this fix for reading MoNa library properly?

src/gui/mzroll/isotopeplot.cpp
```diff
@@ -63,13 +63,17 @@ void IsotopePlot::replot()
...
+    if (!group->tableName().empty())
+        group->childrenBarPlot = group->children;
     if (group->childCountBarPlot() == 0) return;
 
     if (group->isIsotope() && group->getParent() ) {
         setPeakGroup(group->getParent());
         return;
     }
 
+    emit peakGroupSet(!group->tableName().empty());
```
sakshikukreja14, line 76: Have we disabled the labels after plotting from isotope plot, on purpose?

src/core/libmaven/mzUtils.cpp
saifulbkhan: Yes. The split method was splitting on every character within the string individually - not the string as a whole.

src/gui/mzroll/isotopeplot.cpp
saifulbkhan: Yes. So that the user may not get the idea that they can change the plots for integrated groups (basically any group from that is enlisted in a peak-table).

sakshikukreja14: Okkayy..

src/core/libmaven/mzUtils.cpp
sakshikukreja14: cool

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1261,13 +1261,31 @@ void TableDockWidget::deleteSelectedItems()
...
     }
+    connect(treeWidget,
+            SIGNAL(itemSelectionChanged()),
+            this,
+            SLOT(showSelectedGroup()));
+
     auto groupSelected = getSelectedGroup();
     showAllGroups();
 
+    if (allgroups.size() == 0) {
```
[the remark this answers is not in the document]
sakshikukreja14: Changed

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -452,5 +452,40 @@ void BackgroundPeakUpdate::updateGroups(QList<PeakGroup> &groups,
...
+                    C13Flag,
+                    N15Flag,
+                    S34Flag,
+                    D2Flag);
+                isotopeDetection.pullIsotopes(&group);
+                for (PeakGroup& child : group.children) {
+                    child.setTableName(group.tableName());
+                }
+            }
+        }
```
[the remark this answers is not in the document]
sakshikukreja14: will make this change

sakshikukreja14:
> Commit `07b943e` should not have been here. Mistake while pulling changes?
> 
> Also the commit that only changes the "test_untargetedPeakReport.csv" can be merged with its parent commit (one where you actually introduce the adductName fix).
> 
> Added some other comments. Please take a look.


As such no mistake while pulling the changes.. but didn't do "--rebase" so may be that's why it made me commit it

src/gui/mzroll/background_peaks_update.cpp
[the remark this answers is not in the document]
saifulbkhan: Right. I tried this myself. Didn't work. It is because of the way isotopic group are created is different enough. Nearest peaks, RT bounds check and all that would have to happen again. Alright then, we can stick with this solution.

One bug: RTs for groups in bookmark table are not recalculated.

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -461,30 +460,28 @@ void BackgroundPeakUpdate::updateGroups(QList<PeakGroup> &groups,
     for (PeakGroup& group : groups) {
         mavenparameters->allgroups.push_back(group);
     }
+    for (PeakGroup& group : groups) {
```
saifulbkhan: I think we talked, that this could be one single loop?

src/gui/mzroll/background_peaks_update.cpp
```diff
@@ -446,7 +448,8 @@ void BackgroundPeakUpdate::updateGroups(QList<PeakGroup> &groups,
         {
             for(Peak& peak :  group.peaks)
             {
-                if (eic->getSample() == peak.getSample()){
+                if (eic->getSample()->getSampleName() ==
+                    peak.getSample()->getSampleName()){
```
saifulbkhan, line 452: Why resort to using the sample's name and not its pointer? The effect should be the same no? In fact, maybe the earlier way of checking was more robust.

src/gui/mzroll/background_peaks_update.cpp
sakshikukreja14: yeaa... but while doing I noticed.. that pull isotopes works with maven parameters.. I did not want to disturb the main window maven parameters group.. so I made a copy in the earlier loop... where in adding the functionality to same group.. maven parameters all group would not have consistent value

src/gui/mzroll/background_peaks_update.cpp
sakshikukreja14: yeaa.. sure I will make this change.

PR #1313, opened 2020-06-25
More beta fixes

saifulbkhan:
The following things were patched:
- RT windows staying disabled with RT matching on (in "Peaks" dialog).
- Non-global `MavenParameters` object would write to the "lastRun.xml" settings file.
- Editor axes mangled when all peaks are zeroed
- Zeroed groups being auto-deleted on table refresh or session restore.


saifulbkhan:
@sakshikukreja14 Please ignore this PR's CI. For some reason Codecov's bash script is exiting with status 1 while uploading coverage reports. I think this is something they will resolve in time. Until then we can manually check if tests are passing on CI or not, before approving PRs.

PR #1315, opened 2020-07-07
[ftr] Specification of color in metadata upload

sakshikukreja14:
User can now specify colors for samples in metadata upload. Color codes can be specified as either hexadecimal codes or color name.


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2458,6 +2465,29 @@ int MainWindow::loadMetaCsvFile(string filename){
...
+        if (header.count("color") && header["color"] < N) {
+
+            string colorName = fields[header["color"]];
+            redF = new qreal(-1);
+            greenF = new qreal(-1);
+            blueF = new qreal(-1);
+            alpha = new qreal(-1);
+            QColor color;
+            
+            if(!colorName.empty()) {
```
saifulbkhan, line 2478: Minor suggestion: instead of checking the truth condition and then proceeding with (more than 2 or 3) lines of code, I would just `return` or `continue` if the condition is opposite of what I want it to be. You will save on one level of indentation:
```c++
if (colorName.empty())
    continue;

// followed by code that should execute if there was a non-empty color name
.
.
.
```

Does that make sense? You need not change it now, but just something to keep in mind for future.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2366,6 +2369,10 @@ void MainWindow::loadMetaInformation() {
 int MainWindow::loadMetaCsvFile(string filename){
 
     ifstream myfile(filename.c_str());
+
+    ifstream xkcd_colors("src/gui/mzroll/databases/xkcd-colors.json");
+    json rootxkcd_colors = json::parse(xkcd_colors);
```
saifulbkhan, line 2374: Use camel-case.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2458,6 +2465,29 @@ int MainWindow::loadMetaCsvFile(string filename){
...
+            alpha = new qreal(-1);
+            QColor color;
+            
+            if(!colorName.empty()) {
+                
+                if (colorName[0] == '#') {
+                    color.setNamedColor(QString::fromStdString(colorName));
+                    color.getRgbF(redF, greenF, blueF, alpha);
+                } else if (rootxkcd_colors[colorName] != nullptr) { 
+                    string hexValue = rootxkcd_colors[colorName];
```
saifulbkhan, line 2484: If the name of a color does not exist, what does the library return? Quoting a warning from the library's README:
> In function `from_json`, use function `at()` to access the object values rather than `operator[]`. In case a key does not exist, `at` throws an exception that you can handle, whereas `operator[]` exhibits undefined behavior.

If by chance it gives undefined behavior or even an unhandled exception, that would be bad. I did try a color name that did not exist in our xkcd json file, but the software did not crash. Which is good. Still, I would like there to be a check for something like this.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2458,6 +2465,29 @@ int MainWindow::loadMetaCsvFile(string filename){
...
         }
 
+        qreal *redF, *greenF, *blueF, *alpha;
+        if (header.count("color") && header["color"] < N) {
+
+            string colorName = fields[header["color"]];
+            redF = new qreal(-1);
+            greenF = new qreal(-1);
+            blueF = new qreal(-1);
+            alpha = new qreal(-1);
```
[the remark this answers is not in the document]
sakshikukreja14: @saifulbkhan  I have changed this pointer implementation. Actually, 'getRgbF' required these values as pointers, I have passed addresses instead of making them pointers

saifulbkhan (approved):

saifulbkhan:
@sakshikukreja14 The quoted-colors issue still exists (e.g. `aqua blue` works but `"aqua blue"` does not).

sakshikukreja14:
@saifulbkhan  made the change where inverted comas are now acceptable. Also, now the color name can be case insensitive.

PR #1316, opened 2020-07-09
[fix] Fix crash when files with unsupported format are uploaded

sakshikukreja14:
Files with unsupported file format leads to the crash. Now the user will be notified when unsupported file is opened. Ideally, El-MAVEN would open files with supported format and notify the user for unsupported files.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -547,7 +547,8 @@ using namespace mzUtils;
             SIGNAL(compoundsLoaded(QString, int)),
             this,
             SLOT(_postCompoundsDBLoadActions(QString,int)));
-
+    
+    connect(fileLoader, SIGNAL(unsupportedFileFormat(QStringList)), this, SLOT(unsupportedFormat(QStringList)));
     // EMDB singals and slots
     connect(fileLoader,
             SIGNAL(sqliteDBLoadStarted(QString)),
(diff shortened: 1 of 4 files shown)


saifulbkhan:
@sakshikukreja14 When you are patching something specific to a single issue, you should add the issue # somewhere in either the commit string or the PR description. I will be able to find them quickly, instead of searching in the list of issues.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -981,6 +982,31 @@ void MainWindow::threadSave(QString filename)
...
+    for (QString filePath : unsupportedFileList) {
+        auto filePathSplit = mzUtils::split(filePath.toStdString(), "/");
+        string fileName = filePathSplit[filePathSplit.size() - 1];
+        fileList += "<li>" + QString::fromStdString(fileName);
+    }
+
+    auto htmlText = QString("<p>El-MAVEN does not support the file format"
+                            " of the following files: </p>"
+                          "<ul>%1</ul>").arg(fileList);
+    htmlText += "<p>Some files are not uploaded.</p>";
```
saifulbkhan, line 998: "Some files _were_ not uploaded.

saifulbkhan:
@sakshikukreja14 Why the extra "Merge branch ..." commit? If there are conflicts, you should resolve them within your commits while rebasing. Unless the feature branch is huge and its difficult to prevent breaks - which is not the case for this one.

saifulbkhan (approved):

Issue #853, opened 2018-08-28
Give a warning when a user is about to delete all Bookmarked peaks


raaisakuk:
**Is your feature request related to a problem? Please describe.**
I right clicked on my bookmarked group to see if it had a delete option and there was an option "Delete all groups". I didn't notice the all and just clicked it. I don't know if there is an option to undo this behaviour.

**Describe the solution you'd like**
Whenever a user is about to delete all bookmarked groups ask them "Do you want to delete all groups" and by default no should be selected.

**Describe alternatives you've considered**

**Additional context**
Because bookmarking takes time and effort, if one is deleting all groups they should be given a fair warning just to make sure it was not a random click. Otherwise someone can lose day worth of work in a second and they can't even undo.

PR #1318, opened 2020-07-13
[ui] Raise warning if the user tries to delete all groups from table

sakshikukreja14:


.gitignore
```diff
@@ -30,6 +30,7 @@
 bin/crashreporter
 bin/El-MAVEN
 bin/test-libmaven
+bin/node_bin
```
saifulbkhan, line 33: Hah! Thank you :)

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1158,6 +1158,32 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
   updateCompoundWidget();
 }
 
+bool TableDockWidget::deleteAllgroupsWarning() 
+{
+    QMessageBox *warning = new QMessageBox(this);
+    bool selectedOption;
+
+    auto htmlText = QString("<p><b>Are you sure you want to permanently erase all the "
+                            "groups from table?</b></p>");
```
saifulbkhan, line 1167: "... from _this_ table?"

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1158,6 +1158,32 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
...
+    htmlText += "<p>You can't undo this action.</p>";
```
saifulbkhan, line 1168: "You _cannot_ undo ..."

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1158,6 +1158,32 @@ void TableDockWidget::deleteGroup(PeakGroup *groupX) {
...
+    auto yesButton = warning->addButton(tr("Yes"),
+                                  QMessageBox::AcceptRole);
+    warning->exec();
+
+    if(warning->clickedButton() == yesButton)
+        selectedOption = true;
+    else
+        selectedOption = false;
+
+    QCoreApplication::processEvents();
```
saifulbkhan, line 1183: Why was this needed? If not for this, there would be no need for the `selectedOption` boolean and lines 1178 to 1184 can be reduced to:
```C++
if (warning->clickedButton() == yesButton)
    return true;

return false;
```

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1184,6 +1212,14 @@ void TableDockWidget::deleteSelectedItems()
...
         return;
 
+    // checks if the selected item count is same as the no. of 
+    // groups in the table.
+    if (topLevelItemsCount == topLevelGroupCount()) {
+        auto continueDeletion = deleteAllgroupsWarning();
+        if (!continueDeletion) {
+          return;
+        }
+    }
```
saifulbkhan, line 1222: Issue #853 was about warning the user when the select "Delete All Groups" from the context menu as well. This check can still be useful. If the `topLevelItemsCount == topLevelGroupCount()` is true then you simply call `TableDockWidget::deleteAll` method, instead of the complex logic that follows which is totally unnecessary since the user has asked to indiscriminately wipe out each item the table.

saifulbkhan (approved):
Good work. Please rebase.

sakshikukreja14:
Done :)

PR #1319, opened 2020-07-13
[ui] Raises a prompt when no peaks are found instead of empty peak table

sakshikukreja14:


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -560,6 +573,10 @@ void TableDockWidget::showAllGroups() {
     if (viewType == groupView)
       setIntensityColName();
     setVisible(false);
+    _mainwindow->ligandWidget->resetColor();
+    _mainwindow->removePeaksTable(this);
+    noPeakFound();
+    this->~TableDockWidget();
```
saifulbkhan, line 579: Don't do this. Since table deletion makes multiple other calls, simply calling the destructor might block the event loop. When deleting widget (or any `QObject`), always call the `deleteLater` method on the instance. It handles the deallocation asynchronously.

Side note: whenever you do have to call destructors, just use `delete object` and not `this->~ObjectClass()`.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -552,6 +552,19 @@ void TableDockWidget::deleteAll()
   }
 }
 
+void TableDockWidget::noPeakFound()
+{
+  QMessageBox *warning = new QMessageBox(this);
+
+    auto htmlText = QString("<p>No peaks found.</p>");
```
saifulbkhan, line 559: "No peaks we detected. Please check if you settings are not too strict for your data or whether you are able to see any peaks while manually browsing through a compound database."

Try to include a suggestion for the user, unless it is totally obvious what went wrong. This dialog could potentially come up for a user who just started using the software and made a simple yet non-obvious mistake.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -556,7 +556,10 @@ void TableDockWidget::noPeakFound()
 {
   QMessageBox *warning = new QMessageBox(this);
 
-    auto htmlText = QString("<p>No peaks found.</p>");
+    auto htmlText = QString("<p>No peaks we detected. \
```
saifulbkhan, line 559: "No peaks _were_ detected...". My mistake in last comment. You pasted the mistake too :)

sakshikukreja14: oh.. ;p

saifulbkhan (approved):
Approved. Please rebase.

sakshikukreja14:
Rebased :)

PR #1320, opened 2020-07-14
Replot of eic widget as samples in the list are filtered out

sakshikukreja14:


src/gui/mzroll/projectdockwidget.h
```diff
@@ -31,8 +31,25 @@ class ProjectDockWidget : public QDockWidget
...
+     */ 
+    void restoreSampleState();
+    
+
 Q_SIGNALS:
     void samplesDeleted();
+    /**
+     * @brief Signal emitted when the item in treeWidget is clicked.
+     */ 
+    void clicked();
```
saifulbkhan, line 52: Is this really emitted when the user clicks on anything or when you manually emit it when filters are applied to the tree widget item? If so, then the name should reflect that instead.

src/gui/mzroll/projectdockwidget.h
```diff
@@ -31,8 +31,25 @@ class ProjectDockWidget : public QDockWidget
     QColor getSampleColor(mzSample* sample);
     QIcon getSampleIcon(mzSample* sample);
 
+    /**
+     * @brief Map stores the item along with its 
+     * checked and unchecked state.
+     */ 
+    map<QTreeWidgetItem*, bool> saveSampleState;
```
saifulbkhan, line 38: This should be a private member. Also give this a better name like `originalCheckStates` or something. I cannot tell what state it stores when reading the code without looking at the docstring.

src/gui/mzroll/projectdockwidget.cpp
```diff
@@ -137,6 +137,8 @@ ProjectDockWidget::ProjectDockWidget(QMainWindow *parent):
     filterEditor->setPlaceholderText("Sample name filter"); 
     connect(filterEditor, SIGNAL(textEdited(QString)), this, SLOT(filterTreeItems(QString)));
 
+    connect(this, SIGNAL(clicked()), this, SLOT(replotWidget()));
```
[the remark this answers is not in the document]
sakshikukreja14: @saifulbkhan  I have removed this connection as it was not needed. Implementation is working fine without this too. May be for some reasons I had added it earlier.

saifulbkhan (approved):

PR #1321, opened 2020-07-14
[fix] Open options dialog from option in file menu

sakshikukreja14:
This PR is for two followings reasons:
1. "Option" did not appear in "File" menu in MacOS.
2. Clicking "Option" did not trigger Options dialog.


saifulbkhan:
@sakshikukreja14 I have removed the "Options" menu item in another branch since it seems mostly useless to me as we already have a huge icon in the main toolbar which people should use for setting options. So if you don't mind, please close this PR.

sakshikukreja14:
@saifulbkhan sure. I will close it.

PR #1323, opened 2020-07-28
[ftr] Provide user separate options to upload sample and project files

sakshikukreja14:
Now two separate options are provided to user to upload sample and project files. Moreover, user would be notified if he tries to open project file in upload sample file option.
secondly, user would be notified if he tries to open more than one project files at a time.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -1666,57 +1666,59 @@ void MainWindow::analyticsAverageSpectra(){
     analytics->hitEvent("Average Spectra", "Clicked");
 }
 
-void MainWindow::open(QStringList filelist, QFileInfo fileInfo)
+void MainWindow::open()
 {
-    
-    // Changing the title of the main window after selecting the samples
-    setWindowTitle(programName
-                   + " "
-                   + STR(EL_MAVEN_VERSION)
-                   + " "
-                   + fileInfo.fileName());
-
-    QString emdbProjectBeingLoaded = "";
-    Q_FOREACH (QString filename, filelist) {
-        if (fileLoader->isEmdbProject(filename)) {
-            emdbProjectBeingLoaded = filename;
-            analytics->hitEvent("Project Load", "emDB");
-        } else if (fileLoader->isMzrollDbProject(filename)) {
-            emdbProjectBeingLoaded = fileLoader->swapFilenameExtension(filename,
-                                                                       "emDB");
-            analytics->hitEvent("Project Load", "mzrollDB");
-            quantType->setCurrentText("AreaTopNotCorrected");
-        } else if (fileLoader->isMzRollProject(filename)) {
-            analytics->hitEvent("Project Load", "mzroll");
-        }
-        fileLoader->addFileToQueue(filename);
+    // TODO: temporarily added for informing user, remove after a few releases
+    if (!_versionRecordExists()) {
+        QMessageBox msgBox;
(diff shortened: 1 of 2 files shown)


saifulbkhan:
@sakshikukreja14 Before I review this, can you tell me why we had to separate the "Open" action into two sub-actions? I dislike the fact that of all the buttons in the main toolbar, we have the "Open" button with a menu - breaks consistency.

This could have been a much simpler fix:
1. Allow the user to select any number and type of files which are now stored in `filelist`.
2. Run through `filelist` to check if there are any conflicting cases, i.e.,
    - if samples and emDB files both were selected
    - if two emDB files we selected
3. If so, then either allow the user to re-chose which files they really want or make a decision to load the first type of file (if sample then only load samples, if emDB then only load the first emDB) in the list.

I have a feeling that the product/design folks will have an issue with your approach as well. But you may still discuss this with Surbhi and chose which implementation we should go with.

sakshikukreja14:
@saifulbkhan I have changed the implementation according to what was required. Apologies for misunderstanding requirements.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1731,6 +1731,56 @@ void MainWindow::open()
...
+    } else if (!sampleCount && projectCount) {
+        if (projectCount > 1) {
+            filelist << projectList[0];
+        } else {
+            filelist = projectList;
+        }
+    } else {
+        QMessageBox msgBox;
+        msgBox.setText("It seems you are trying to load sample files and "
+                       "project files on one instance.\n\n"
```
saifulbkhan, line 1761: "It seems you are trying to load sample files and project files at once. Please select either a set of samples or a single project file to load into this session."

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1731,6 +1731,56 @@ void MainWindow::open()
...
+                       "Instead, you can choose to upload either of the two.");
+        QPushButton* sampleChosen = msgBox.addButton("Sample Files",
+                                                       QMessageBox::ActionRole);
+        QPushButton* projectChosen = msgBox.addButton("ProjectFiles",
+                                                     QMessageBox::RejectRole);
+
+        msgBox.setDefaultButton(sampleChosen);
```
saifulbkhan, line 1768: I would have made the project button as default. The first project in the file list.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -3023,6 +3073,8 @@ void MainWindow::createToolBars() {
 	btnOpen->setStyleSheet("QToolTip {color: #000000; background-color: #fbfbd5; border: 1px solid black; padding: 1px;}");
 	btnOpen->setToolTip(tr("Sample uploads"));
 
+    connect(btnOpen, SIGNAL(clicked()), SLOT(open()));
```
saifulbkhan, line 3076: Use the type-strict syntax for connecting signals-slots.

src/gui/mzroll/mainwindow.h
```diff
@@ -273,6 +273,8 @@ Q_OBJECT
      */
     TableDockWidget* activeTable();
 
+	void showLargeFilesNotification(QStringList, QFileInfo);
```
saifulbkhan, line 276: This is no longer required, right?

saifulbkhan (changes requested):
@sakshikukreja14 Thanks for making the change. Looks much better now. If you resolve the comments and rebase, this will ready to merge.

PR #1324, opened 2020-07-29
[fix] Fix crash caused by "/" in text search

sakshikukreja14:
El-MAVEN caused crash when typed "/" or for any other search. This was due to no memory allocation for "masscutoff" in peptide fregmentation class.


saifulbkhan (changes requested):
@sakshikukreja14 Even though the crash has been fixed, I am not able to search for precursor-product EIC using the `precursorMz/productMz` format in search box. If you load some MRM samples and search for `massA/massB` then the EIC for that transition should be visible to the user. Right now it shows a value close 1.0 whenever I search for anything of that sort. Ping me directly if you this does not make much sense to you.

PR #1327, opened 2020-08-12
[Ref] Isotope and adduct detection

saifulbkhan:
Isotope detection has been rewritten to:
* not depend on detection of parent ions for any non-parent ions to be detected,
* optionally - filter labelled forms based on the deviation or correlation w.r.t. the unlabelled form (_if_ it was found at all)
* allow a way to look at all possible isotopes of a given compound as well as individually bookmark them,
* keep only one of parent peak-group and get rid of the child group with "C12 PARENT" tag (the tag now resides in the parent itself).

Adduct detection has similarly been rewritten to not rely on existence of default adduct ions ([M-H] or [M+H]). Filtering may also be applied if parent ion is found.

**Note to reviewer**: There is a significant amount of change in peak-detection logic as well as the UI logic for peak-tables, isotope list and plot, EIC widget and compound (ligand) widget. Please do your best to try to break the software on any edge cases that you can think of. Since there is a large amount of code change, I would suggest reviewing the around the major rewrites only (i.e., classes - `MassSlicer`, `PeakDetector`, `GroupFiltering`, `PeakGroup`, `BackgroundOpsThread` and the UI classes for the widgets that I named earlier). The modules `IsotopeDetection` and `AdductDetection` have been removed since their functionality has now been split under `MassSlicer`, `PeakDetector` and `GroupFiltering`. You should also ensure that exports and imports of all kinds are working as they should (reach out to me if you are confused about some previously unseen changes to exports especially). We will perform scientific validation of the whole workflow separately from this PR review - so you need not worry too much about the correctness of the algorithms just yet.


saifulbkhan:
@sakshikukreja14 Also the tests will fail for now, but since I am anticipating changes during the course of the review, I would like to update the tests at the end of it. You are free to withhold approval until then. :)

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -586,6 +590,11 @@ void LigandWidget::resetColor()
 
 void LigandWidget::updateIsotopesAndAdducts()
 {
+    QApplication::processEvents();
+    _busyMessage->setText("Populating isotopologues and adducts…");
```
sakshikukreja14, line 594: @saifulbkhan  This notification is also received when we uncheck the isotopes or adducts option. Is this the expected behaviour? As message 'populating isotopologues and adducts' didn't make sense to me, when we uncheck either of the two.

saifulbkhan: Oh good point. I will substitute an appropriate message for such cases.

src/gui/mzroll/eicwidget.cpp
```diff
@@ -1081,6 +1081,8 @@ void EicWidget::replot(shared_ptr<PeakGroup> group)
         float rtMin = group->minRt;
         float rtMax = group->maxRt;
         addEICLines(_showSpline, _showEIC, true, rtMin, rtMax);
+        _clearEicPoints();
+        addPeakPositions(group);
```
sakshikukreja14, line 1085: 'PeakPositions' must also be added in 'renderPdf' function. Peak-group bubbles are not visible when exported as pdf.

saifulbkhan: This was intentional. They look ugly in the exported PDF. Peak bubbles are meant to convey the quality of each peak which is usually helpful in making decisions when you are within the software itself. The exported PDF is a report for sharing EICs quickly with people who do not want to open the software or have no access to it.

src/cli/peakdetector/peakdetectorcli.cpp
```diff
@@ -703,6 +705,50 @@ void PeakDetectorCLI::loadSamples(vector<string>& filenames)
...
+                "cor", false, 2.0, 1.0, 0.20, 3.40, 0.0, 20.0, false, 0.60);
+            Aligner mzAligner;
+            mzAligner.alignWithObiWarp(
+                mavenParameters->samples, &params, mavenParameters);
+        }
+        break;
+
+        case 2: {
+            mavenParameters->writeCSVFlag = false;
+            peakDetector->processFeatures();
```
shubhra-agrawal, line 727: PolyFit had the option to align groups wrt compound search or mass slices. Has that option been removed now?

saifulbkhan: No, in the GUI, you can still chose the method for detecting your reference groups. This is CLI. We never offered that kind of control here.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
-                    && groupWithTarget.ms2EventCount > 0
-                    && groupFiltering.filterByMS2(groupWithTarget)) {
-                    continue;
-                }
+            if (!orphans.empty()) {
+                // for orphans, create a ghost, that will act as an empty parent
+                PeakGroup parentGroup(
+                    make_shared<MavenParameters>(*_mavenParameters),
+                    PeakGroup::IntegrationType::Ghost);
+                container.push_back(parentGroup);
```
shubhra-agrawal, line 814: Note for self:
is there a new ghost parent for every orphan isotope or adduct? 
if there are M+1, M+2 for the same compound at multiple RTs, how is that handled? do they show up under different parents or the same?

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
+    int groupId = 1;
+    for (auto& group : _mavenParameters->allgroups)
+        group.setGroupId(groupId++);
+
+    if (nonParentIsotopologues.empty() && nonParentAdducts.empty())
+        return;
+
+    // lambda: given a compound and its child indexes, clubs them with their
+    // most likely parent-group if possible, otherwise adds them to a ghost
+    auto makeMeta =
```
shubhra-agrawal, line 779: Could you move this out in a function instead? a 60-line lambda defined in a function hurts readability imo. 
Let me know if there's some advantage to this approach that I'm missing.

src/core/libmaven/PeakDetector.cpp
saifulbkhan: Same parent. Every labelled form for which there is no parent, is put under the one ghost parent. The ghost is only a container (for programming ease), and shows up as such in the UI (no data for any columns only a name and a group ID).

src/core/libmaven/PeakDetector.cpp
saifulbkhan: It started out small. I do not think it hurts readability. In a good C++ editor, you can minimize these code-blocks. But most people will agree with you. I will move this out.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
+                                         adductIndexes,
+                                         parentCompounds,
+                                         [](PeakGroup* group) {
+                                             return group->adduct()->getName();
+                                         });
+        if (metaGroups.count(compound) > 0) {
+            auto& existingMetaGroups = metaGroups[compound];
+            for (auto& elem : metaAdductGroups) {
+                size_t parentIndex = elem.first;
+                vector<size_t>& childIndexes = elem.second;
```
shubhra-agrawal, line 870: Would suggest renaming `childIndexes` to `adductIndices` or `childAdductIndices` for clarity. Hard to read this block with so many similar variable names.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
+        if (_mavenParameters->stop) {
+            _mavenParameters->allgroups.clear();
+            return;
         }
 
-       sendBoostSignal("Identifying features using the given compound set…",
-                       iter - mavenParameters->allgroups.begin(),
-                       mavenParameters->allgroups.size());
+        auto& compoundMetaGroups = elem.second;
+        for (auto& metaGroup : compoundMetaGroups) {
```
shubhra-agrawal, line 895: `metaGroups` is already a vector here. Naming something as `metaGroup` just 2 lines down is asking for trouble, even if its a looping variable.

saifulbkhan: I do this all the time. Give the container a plural name of its children. But in this case it should have been `compoundMetaGroup` instead of just `metaGroup` (which, I guess, you would still object too). I think I must have given it a shorter name because long names within multiple-indented often exceed 80 characters, which I do not like. I will see if I can come up with something more concise.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
         }
+    }
 
-        if (matchFound) {
-            iter = mavenParameters->allgroups.erase(iter);
-        } else {
-            ++iter;
+    // perform final meta-grouping and queue children to be erased
+    vector<size_t> indexesToErase;
+    for (auto& elem : metaGroups) {
```
shubhra-agrawal, line 888: so metaGroups is "compound_name" : [ parentIndex1:[childIndices], parentIndex2:[childIndices],...parentIndexN:[childIndices] ]

Changing `elem` to `compound_name` might make this easier to comprehend.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -376,103 +527,435 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
+            continue;
+
+        if (group.isotope().isParent() && group.adduct()->isParent()) {
+            if (parentCompounds.count(compound) == 0)
+                parentCompounds[compound] = {};
+            parentCompounds[compound].push_back(i);
+        }
+    }
+    if (applyGroupFilters) {
+        _keepNBestRanked(parentCompounds,
```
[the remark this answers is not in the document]
saifulbkhan: You will get the N best real parent groups. Ghost groups do not exist at this point. They are constructed when the `makeMeta` function is called.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -316,41 +461,47 @@ void PeakDetector::processSlices(vector<mzSlice*> &slices, string setName)
...
-        std::sort(peakgroups.begin(), peakgroups.end(),
-                  PeakGroup::compRank);
 
-        for (unsigned int j = 0; j < peakgroups.size(); j++) {
-            // check for duplicates	and append group
-            if (j >= mavenParameters->eicMaxGroups)
-                break;
-            mavenParameters->allgroups.push_back(peakgroups[j]);
+        // we do not filter non-parent adducts or non-parent isotopologues
+        bool isParentGroup = slice->adduct == nullptr
```
shubhra-agrawal, line 466: To self:
What is the `slice->adduct` value when it's an isotope slice?

saifulbkhan: What is this "to self" thing?

All slices that have a target compound have a non-null adduct. Isotope slices have a target compound, so their adduct is by default [M+H] or [M-H] depending on global polarity.

src/gui/mzroll/mzfileio.cpp
```diff
@@ -1322,9 +1342,7 @@ PeakGroup* mzFileIO::readGroupXML(QXmlStreamReader& xml, PeakGroup* parent)
         make_shared<MavenParameters>(*_mainwindow->mavenParameters),
```
shubhra-agrawal, line 1342: Any plans to drop mzroll support completely? just curious

src/gui/mzroll/ligandwidget.cpp
```diff
@@ -240,281 +225,531 @@ void LigandWidget::readCompoundXML(QXmlStreamReader& xml, string dbname) {
...
-
+    treeWidget->setColumnWidth(0, 250);
+    treeWidget->resizeColumnToContents(2);
+    if (numCompoundsWithFormula == 0) {
+        treeWidget->hideColumn(1);
+    } else {
+        treeWidget->showColumn(1);
+        treeWidget->resizeColumnToContents(1);
+    }
+    if (numCompoundsWithRt == 0) {
```
shubhra-agrawal, line 493: oh nice idea.

src/core/libmaven/groupFiltering.cpp
```diff
@@ -124,6 +122,76 @@ bool GroupFiltering::filterByMS2(PeakGroup& peakgroup)
...
+        for (auto sample : _mavenParameters->samples) {
+            auto childPeak = child->getPeak(sample);
+            auto parentPeak = parent.getPeak(sample);
+            if (parentPeak == nullptr || childPeak == nullptr)
+                continue;
+
+            ++numSamplesShared;
+            auto childApexRt = childPeak->rt;
+            auto parentApexRt = parentPeak->rt;
+            auto deviation = abs(childApexRt - parentApexRt) * 60.0f;
```
shubhra-agrawal, line 151: Could you add a comment here to explain that the multiplication with 60 is being done to convert the time unit? it's pretty obvious but still.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -254,13 +238,20 @@ void PeakGroup::setCompound(Compound* compound)
...
 
 void PeakGroup::setSlice(const mzSlice& slice)
 {
     _slice = slice;
     _sliceSet = true;
+    if (!_slice.isotope.isNone()) {
+        tagIsotope(_slice.isotope.name,
+                   _slice.isotope.mass,
+                   _slice.isotope.abundance);
+        isotopeC13count = _slice.isotope.C13;
```
shubhra-agrawal, line 252: I doubt `isotopeC13count` is used anywhere. You can probably remove it.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1006,6 +1034,35 @@ string PeakGroup::tableName() const
 void PeakGroup::setTableName(string tableName)
 {
     _tableName = tableName;
-    for (auto child : children)
+    for (auto child : _childIsotopes)
+        child->setTableName(tableName);
+    for (auto child : _childAdducts)
```
shubhra-agrawal, line 1039: I think I saw a check for tableName in barplot somewhere. Can you check and update if that's the case?

src/core/libmaven/PeakDetector.cpp
shubhra-agrawal: I was writing a note for myself to go back and check. Forgot I wasn't in review mode so it got published.

src/core/libmaven/PeakDetector.cpp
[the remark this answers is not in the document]
shubhra-agrawal: Yeah, I meant in the final output. So I'll have a ghost group for different compounds with all orphan children, both isotopes and adducts. Got it.

src/core/libmaven/PeakDetector.cpp
```diff
@@ -109,205 +126,333 @@ vector<EIC*> PeakDetector::pullEICs(const mzSlice* slice,
...
+        processSlices(massSlicer.slices, setName, applyGroupFilters, true);
+        delete_all(massSlicer.slices);
+    }
+
+    performMetaGrouping(applyGroupFilters, findBarplotIsotopes);
+
+    // filter-out ghost parent groups if parent is required
+    vector<size_t> ghostIndexes;
+    for (size_t i = 0; i < _mavenParameters->allgroups.size(); ++i) {
+        auto& group = _mavenParameters->allgroups[i];
```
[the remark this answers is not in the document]
saifulbkhan: There are different ghost containers for adducts and isotopes (when doing automated detection).

src/gui/mzroll/mzfileio.cpp
saifulbkhan: I had planned on dropping it in this release itself - too many open issues around mzroll import itself. But I am short on time. So maybe later.

src/core/libmaven/groupFiltering.cpp
saifulbkhan: Okay.

src/core/libmaven/PeakGroup.cpp
saifulbkhan: Right. Removing.

src/core/libmaven/PeakGroup.cpp
saifulbkhan: This is safe. The one in barplot is meant to check if the _parent_ peak-group is from a peak table or not. If so, then just display the already known child isotopes instead of finding and populating the barplot-specific isotopes (which can be using tracers). Barplot-specific isotopic groups are never presented to the user in any other way.

PR #1328, opened 2020-08-13
Add '.xml' to the list of filetype of load settings dialog. 

sakshikukreja14:
This PR is for following two tasks: 

1. Add '.xml' filetype while saving El-MAVEN settings.
2. Only '.xml' files would be loaded in settings dialog. Any other uploaded file would give the user warning of unsupported file format.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -2994,14 +2994,16 @@ void MainWindow::loadSettings()
     };
 
     QFile file(fileName);
-    if(!(fileName.contains(".xml") || fileName.contains(".XML"))) {
+    if(fileName.size() 
+        && !(fileName.contains(".xml") 
+            || fileName.contains(".XML"))) {
         auto fileNameSplit = mzUtils::split(fileName.toStdString(), "/");
         string fileNameString = fileNameSplit[fileNameSplit.size() - 1];
         fileName = "<li>" + QString::fromStdString(fileNameString);
         auto htmlText = QString("<p>El-MAVEN settings does not support the file format"
                             " of the following file: </p>"
                           "<ul>%1</ul>").arg(fileName);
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2984,10 +3020,7 @@ void MainWindow::loadSettings()
 
     else {
         // display an error message
-        QMessageBox msgBox;
-        msgBox.setWindowTitle("Error");
-        msgBox.setText("Loading the file failed");
-        msgBox.exec();
+        warning("Unable to load settings. File name not specified.");
```
saifulbkhan, line 3023: "Unable to load settings. Specified file is not readable."

src/gui/mzroll/mainwindow.cpp
```diff
@@ -2965,10 +2968,43 @@ void MainWindow::loadPollySettings(QString fileName)
...
+
+    QFile file(fileName);
+    if(!(fileName.contains(".xml") || fileName.contains(".XML"))) {
+        auto fileNameSplit = mzUtils::split(fileName.toStdString(), "/");
+        string fileNameString = fileNameSplit[fileNameSplit.size() - 1];
+        fileName = "<li>" + QString::fromStdString(fileNameString);
+        auto htmlText = QString("<p>El-MAVEN settings does not support the file format"
+                            " of the following file: </p>"
+                          "<ul>%1</ul>").arg(fileName);
+        htmlText += "<p>Setting file is not uploaded.</p>";
```
saifulbkhan, line 3004: "Setting**s** file _was_ not uploaded".

saifulbkhan (changes requested):
@sakshikukreja14 Works as it should. Added minor string change requests.

saifulbkhan (approved):

PR #1329, opened 2020-08-25
[Ref] UI clean-up and consistent styling across platforms

saifulbkhan:
This branch contains the following changes:
* a consistent theme for all three desktop environments (helps deal with many platform-specific UI issues)
* several minor changes to the UI based on existing issues
* removal of many classes and widgets that were no longer used or were non-functional
* removal of "libpillow" as a dependency (we do not need an HTTP server, at least yet)
* removal of input formats no longer acceptable (peptide databases, RAW files, mzData, etc.)

PR #1330, opened 2020-08-25
[fix] Fix 'deleteALl' warning for bookmark table

sakshikukreja14:


saifulbkhan (approved):
@sakshikukreja14 Seems to be working. Do you mind if we merge this after #1329? There should not be rebase issues either way, but just to be safe.

sakshikukreja14:
For sure :). we can wait.

PR #1331, opened 2020-08-26
Alpha fixes

sakshikukreja14:


src/gui/mzroll/projectsaveworker.cpp
```diff
@@ -89,7 +89,9 @@ void TempProjectSaveWorker::saveProject(const bool saveRawData)
     if (_currentProjectName.isEmpty()) {
         auto now = QDateTime::currentDateTime();
         auto tempFileName = now.toString("dd_MM_yyyy_hh_mm_ss") + ".emDB";
-        auto firstSampleFile = _mw->getSamples()[0]->fileName;
+        string firstSampleFile = "";
+        if (!_mw->getSamples().empty())
+            auto firstSampleFile = _mw->getSamples()[0]->fileName;
```
saifulbkhan, line 94: The good/bad button should not even trigger an "autosave" kind of event, when there is no peak-group selected in a table.

sakshikukreja14: Okkay.. will make the change.

saifulbkhan: You declared a new variable by the name of `firstSampleFile`, therefore the one in outer scope was not overwritten. This is also why autosave now fails because it tries to write a temporary file in root (`/`) directory. Check your terminal logs.

sakshikukreja14: Made the correction.

saifulbkhan: Oh your last commit overwrote this change. Ignore.

saifulbkhan (approved):

PR #1332, opened 2020-09-15
Beta fixes

sakshikukreja14:


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2485,35 +2488,34 @@ int MainWindow::loadMetaCsvFile(string filename){
...
             QColor color;
             
-            if(colorName.empty()) 
-                continue;    
-            if (colorName[0] == '#') {
-                color.setNamedColor(QString::fromStdString(colorName));
-                color.getRgbF(&redF, &greenF, &blueF, &alpha);
-            } else {
-                try {
+            if(!colorName.empty()) {
```
saifulbkhan, line 2500: @sakshikukreja14 Any reason why the code from here on was modified in this patch? Once you've read the JSON from its correct path, I would assume there is no reason to make any further changes.

sakshikukreja14: Earlier, we were using "continue" instead. But it was skipping the implementation of assigning cohort values if color name was empty.

src/core/libmaven/peakdetector.cpp
```diff
@@ -156,15 +156,30 @@ void PeakDetector::identifyFeatures(const vector<Compound*>& identificationSet)
...
         massSlicer.generateAdductSlices(identificationSet, true, false);
+        massSlicer.generateCompoundSlices(compoundsWithNeutralMassOnly, false);
+        massSlicer.generateCompoundSlices(compoundsWithRawMzOnly, false);
     } else if (_mavenParameters->pullIsotopesFlag) {
         massSlicer.generateIsotopeSlices(identificationSet);
+        massSlicer.generateCompoundSlices(compoundsWithNeutralMassOnly, false);
+        massSlicer.generateCompoundSlices(compoundsWithRawMzOnly, false);
     } else if (_mavenParameters->searchAdducts) {
         massSlicer.generateAdductSlices(identificationSet);
+        massSlicer.generateCompoundSlices(compoundsWithRawMzOnly, false);
```
sakshikukreja14, line 182: Why not compounds with neutral mass added here?

src/core/libmaven/peakdetector.cpp
```diff
@@ -992,6 +992,7 @@ void PeakDetector::linkParentIsotopeRange(PeakGroup& parentGroup,
                 child->deletePeak(sample);
         }
 
+        delete_all(eics);
```
sakshikukreja14, line 995: Was this the reason for crash while detecting peaks, reported by testing team?

src/core/libmaven/peakdetector.cpp
saifulbkhan: Right, that was intentional. For isotopes, at least formula is required, but for adducts, either formula or neutral mass is sufficient for calculating adduct mass. So compounds that have formula or neutral mass do not need any special treatment here.

src/core/libmaven/peakdetector.cpp
saifulbkhan: Yes. Entire EICs for each peak-group (and its children) were being created but not being deallocated during this step. And each EIC is a large object - this was a very bad memory leak.

saifulbkhan (approved):

PR #1333, opened 2020-09-18
[Fix] Export empty parent-groups to Polly

saifulbkhan:
This is a Polly-only fix that inserts ghost groups with empty cells in order to make sure that Phi does not break.


sakshikukreja14 (approved):
Works. Approving :)

PR #1336, opened 2020-10-14
Fix VIB issues

saifulbkhan:
Minor UX changes along with refactors for algorithmic consistency in peak-detection (specifically manual vs. automatically integrated peaks).


sakshikukreja14:
While reviewing, I came across a weird bug. When a peak-group is bookmarked, and later a user tries to merge it to a table. A warning appears if he is sure he wants to merge.  Even if clicked "No" to the warning, an alert is generated which says that group is successfully merged. If we can fix it in this branch, it would be great.

src/gui/mzroll/settingsform.cpp
```diff
@@ -221,6 +221,11 @@ SettingsForm::SettingsForm(QSettings* s, MainWindow *w): QDialog(w) {
             &SettingsForm::loadModel);
     _updateModelPath();
 
+    connect(autosaveCheckBox,
```
sakshikukreja14, line 224: Should we connect _"use default parameters"_ in settings dialog to uncheck auto-save checkbox by default??

saifulbkhan: Hm... no. I would want the user to make this decision and then it should stay that way until they chose to change it. "Use default parameters" is as I understand only for algorithm parameters.

saifulbkhan:
@sakshikukreja14 
> A warning appears if he is sure he wants to merge.
This is actually our "delete all peak-groups" warning from the parent peak-table class. When merging, groups are replicated in the other table and deleted from bookmark table - rather than actually moving groups from one to another.

I will make the change such that the warning message does not come up for these merge operations.

sakshikukreja14 (approved):

PR #1339, opened 2020-10-27
[fix] Fix reading csv database file for missing columns

sakshikukreja14:


saifulbkhan (approved):
Thanks @sakshikukreja14. I will rebase `ftr_pd_ddx` on top of this. You should do the same for your branch(es).

PR #1340, opened 2020-10-28
Fixes related to storing correct RT bounds

saifulbkhan:
Two changes for consistency in what RT bounds for slices are stored in peak-groups (and therefore in emDBs as well).
1. For isotopes which are "linked" to their parent (unlabelled) peak-groups, the slice has been made to have the exact same RT bound.
2. For manually bookmarked peak-groups, the RT bounds will always be the entire RT range (inclusive over all currently selected samples).

### DIFF
--- a/src/gui/mzroll/eicwidget.h
+++ b/src/gui/mzroll/eicwidget.h
@@ -6,6 +6,7 @@
 #include "stable.h"
 
 #include "datastructures/isotope.h"
+#include "datastructures/mzSlice.h"
 
 class EIC;
 class EICLogic;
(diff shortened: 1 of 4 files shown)


sakshikukreja14 (approved):

src/gui/mzroll/eicwidget.h
```diff
@@ -257,7 +260,6 @@ public Q_SLOTS:
 	void cleanup();		//deallocate eics, fragments, peaks, peakgroups
 	void clearPlot();	//removes non permenent graphics objects
 	void findPlotBounds(); //find _minX, _maxX...etc
-	mzSlice visibleSamplesBounds();
```
sakshikukreja14, line 258: @saiful-elucidata  _visibleSampleBounds_ is not used as a slot. I think we should move this function to public and not in slots.

saif-el: Good catch. Moving to `public` methods.

PR #1341, opened 2020-10-28
[Fix] Incorrect m/z calculation for EI type

saifulbkhan:
The calculation of m/z differs only for EI ionization type. This
was not taken into account when calculating the m/z value from
parent adduct ions (which should still obey the EI/non-EI rules).

Additinally, the logic for setting ionization type in the
`MavenParameters` class was constantly assigning ESI even when
unintended.

### DIFF
--- a/src/core/libmaven/mavenparameters.cpp
+++ b/src/core/libmaven/mavenparameters.cpp
@@ -524,10 +524,12 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
         setIonizationMode((Polarity)polarity);
     }
 
-    if (strcmp(key, "ionizationType") == 0 && stoi(value) == 1) {
-        MassCalculator::ionizationType = MassCalculator::EI;
-    } else {
-        MassCalculator::ionizationType = MassCalculator::ESI;
+    if (strcmp(key, "ionizationType") == 0) {
+        if (stoi(value) == 1) {
+            MassCalculator::ionizationType = MassCalculator::EI;
+        } else {
+            MassCalculator::ionizationType = MassCalculator::ESI;
+        }
     }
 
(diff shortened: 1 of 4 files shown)


src/core/libmaven/mavenparameters.cpp
```diff
@@ -524,10 +524,12 @@ void MavenParameters::setOptionsDialogSettings(const char* key, const char* valu
         setIonizationMode((Polarity)polarity);
     }
 
-    if (strcmp(key, "ionizationType") == 0 && stoi(value) == 1) {
-        MassCalculator::ionizationType = MassCalculator::EI;
-    } else {
-        MassCalculator::ionizationType = MassCalculator::ESI;
+    if (strcmp(key, "ionizationType") == 0) {
```
sakshikukreja14, line 527: Isn't the new logic same as the earlier used?

saifulbkhan: Nope. Earlier it would execute the `else` block even if the key was not "**ionizationType**". Now it will only execute the else block only if key _is_ "**ionizationType**" but value is other than **1**.

This is important since this method is called every time when a `MavenParameters` settings needs to be changed. Suppose I had set my "**ionizationType**" to be `EI`. And after that I try to set the "**minPeakQuality**" key:

```C++
mp->setOptionsDialogSettings("ionizationType", "1");
mp->setOptionsDialogSettings("minPeakQuality", "0.5");
```

In the second call, while setting "**minPeakQuality**", the else block for "**ionizationType**" would also be executed (unintentionally) and set the "**ionizationType**" back to `ESI`.

sakshikukreja14: Great! understood. Approving.

sakshikukreja14 (approved):

Issue #1347, opened 2020-11-12
extracting MS1 EIC from PRM data leads to huge slowdown or crash 


chubukov:
Mostly PRM data seems to works ok. We're able to extract parent/daughter m/z pairs, as well as EICs for MS1 masses.

However, if I try to extract the MS1 EIC for for the PRM parent m/z, I get some awful behaviour (with a large number of files loaded, it crashes, with a small number of files, it slows down a lot, and I get some very odd grass in the EIC plot). My suspicion is that what's happening is that the fragmentation events widget is trying to process every single scan and this ends up being a huge cpu/memory load

Sample files and source file uploaded to our egnyte folder in the 11_11_2020_prm_bug folder.

saifulbkhan:
@chubukov I was unable to find a reliable way of automatically deducing MS/MS data-type encoded in samples. So in #1349, I have added an input prompt which appears after loading PRM or DDA sample types to confirm this from the user. If you select "PRM" there, you should not experience any issues that would have been happening due to the fragmentation spectra/events widget.

Once built, the Windows installer should be available [here](https://elmaven-installers.s3-us-west-2.amazonaws.com/Windows/El-MAVEN-fix_diff_prm_dda.exe). Hopefully, your slowdown (and crashes) should go away.

chubukov:
@saifulbkhan ok, thanks. I agree it might be tricky to decide exactly what kind of method you have only from the mzML (although possible, I think). But maybe instead of the dialog you could simply gate on the relevant parameter, which seems to be the number of scans in the function?

saifulbkhan:
@chubukov Sorry, I did not understand. Do you mean the number of MS/MS scans? And in which function?

Implementation aside, did the separate PRM treatment help with your slowdown issues with large number of samples?

PR #1349, opened 2020-11-12
[Fix] Add prompt to confirm MS/MS type #1347

saifulbkhan:
When loading a sample having MS1 and MS2 scans El-MAVEN has no way of being able to tell whether these samples were from a DDA or a PRM experiment. From now on, an input dialog will ask the user to confirm the MS/MS data type, whenever they load such samples. For now, this prevents MS2 events logic from interfering with PRM data analysis. It could potentially help with other things in the future.

**Note to reviewer**: please ask me for sample files that you can use to test this branch.

### DIFF
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -605,6 +605,10 @@ using namespace mzUtils;
             SIGNAL(sampleLoaded()),
             this,
             SLOT(_warnIfNISTPolarityMismatch()));
+    connect(fileLoader,
+            &mzFileIO::sampleLoaded,
+            this,
+            &MainWindow::_confirmMsMsType);
 
 	connect(fileLoader,SIGNAL(sampleLoaded()), this, SLOT(setInjectionOrderFromTimeStamp()));
     connect(fileLoader,SIGNAL(sampleLoaded()),projectDockWidget, SLOT(updateSampleList()));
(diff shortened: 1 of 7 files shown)


src/gui/mzroll/mainwindow.cpp
```diff
@@ -2143,6 +2145,38 @@ void MainWindow::_warnIfNISTPolarityMismatch()
     }
 }
 
+void MainWindow::_confirmMsMsType()
+{
+    QList<mzSample*> msMsSamples;
+    for (auto sample : samples) {
```
sakshikukreja14, line 2151: @saiful-elucidata, this function makes use of all the samples present in the session and not the newly loaded. 

So, in case we load the PRM dataset in the El-MAVEN session, we get the prompt which is expected behavior. After loading the PRM dataset, if we try to load the LCMS dataset in the same session we again get the prompt, which in my viewpoint must not appear. 

But it is up to you if we would like to handle it or not.

saif-el: To be able to remember which type of dataset(s) have been loaded into the software we will need an additional settings/MavenParameters storage. I would like to avoid that for something that users are very unlikely to do. The only time you would load samples in batches in El-MAVEN would be when it's too large to be loaded at once (because of memory issues) and even then the batches of sample files would be from the same dataset. Loading different types of datasets together has never been handled or allowed within the software. This is fine in my opinion.

sakshikukreja14: This makes sense. I will approve it. :)

sakshikukreja14 (approved):

PR #1360, opened 2020-12-08
[Fix] Adjusted m/z ignoring global charge

saifulbkhan:
When we moved to calculating masses purely using adducts attached to compounds (which is always true for non-isotopes), the global charge stopped affecting the expected m/z values for compounds with formulas or neutral masses. This issue has been resolved.


sakshikukreja14 (approved):
works

PR #1361, opened 2020-12-10
Ftr peak ml

sakshikukreja14:


.travis.yml
```diff
@@ -55,7 +55,7 @@ install:
       sudo apt-get update -qq;
       sudo apt-get install -y -qq --allow-unauthenticated qt5-qmake qtbase5-dev qtscript5-dev qtdeclarative5-dev libqt5webkit5-dev;
       sudo apt-get install -y -qq --allow-unauthenticated --allow-downgrades libpulse-mainloop-glib0=1:8.0-0ubuntu3.10 libpulse0=1:8.0-0ubuntu3.10 libpulse-dev
-      sudo apt-get install -y -qq --allow-unauthenticated libqt5opengl5 libqt5multimedia5 libqt5multimedia5-plugins qtmultimedia5-dev;
+      sudo apt-get install -y -qq --allow-unauthenticated libqt5opengl5 libqt5multimedia5 libqt5multimedia5-plugins qtmultimedia5-dev libqt5charts5-dev;
```
saifulbkhan: Are we still using Qt Charts library? I thought we moved to a custom (QGraphicsView-based) plot for the force plot.

If we still are using it, please make sure to add this dependency in our README as well (for all three platforms).

src/core/libmaven/PeakGroup.cpp
```diff
@@ -44,6 +44,7 @@ PeakGroup::PeakGroup(shared_ptr<MavenParameters> parameters,
     sampleMean=0;
 
     deletedFlag = false;
+    isHiddenFromTable = false;
```
saifulbkhan: I do not like the fact that a UI-specific flag is leaking into the core library. I know there are cases like this in the existing codebase, but we should try to minimize this with any new code.

One solution would be for each peak-table to keep a simple map of peak-group rows that are hidden and act accordingly. Feel free to ignore this if I have misunderstood the use case here.

If you choose to keep it, make it private and expose using a getter.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1056,93 @@ void PeakGroup::setGroupId(int groupId)
...
+        return;
+    }
+    _predictionProbability = 0.0f;
+    _predictionInference.clear();
+    _userLabel = label;
+
+    if (parent != nullptr && tagString == "C12 PARENT") {
+        parent->setUserLabel(label);
+        return;
+    }
```
saifulbkhan: This will never be true anymore (from v0.12.0 and above). The `tagString` of all child isotopes are always "C13-label-x", "N15-label-x", etc.

Instead, you could change the second conditional to: `parent->tagString == C12_PARENT_LABEL`, and get the same effect.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1056,93 @@ void PeakGroup::setGroupId(int groupId)
...
+
+    for (auto& child : _childIsotopes) {
+        if (child->tagString == "C12 PARENT" && child->_userLabel != label)
+            child->setUserLabel(label);
+    }
```
saifulbkhan: This is no longer needed. The child isotope that used to have "C12 PARENT" as its `tagString` is gone.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1056,93 @@ void PeakGroup::setGroupId(int groupId)
...
+        this->_userLabel = 'b';
+    } else {
+        this->_userLabel = 'g';
+    }
+    _predictionProbability = probability;
+
+    if (parent != nullptr && tagString == "C12 PARENT") {
+        parent->setPredictedLabel(label, probability);
+        return;
+    }
```
saifulbkhan: Change the second conditional as mentioned in the comment above for `PeakGroup::setUserLabel` method.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1056,93 @@ void PeakGroup::setGroupId(int groupId)
...
+    for (auto& child : _childIsotopes) {
+        if (child->tagString == "C12 PARENT"
+            && child->_predictedLabel != label
+            && child->_predictionProbability != probability)
+            child->setPredictedLabel(label, probability);
+    }
```
saifulbkhan: Again, not needed anymore. Parent peak-group is the only "C12 PARENT".

src/core/libmaven/PeakGroup.h
```diff
@@ -134,19 +181,22 @@ class PeakGroup{
...
         float maxSignalBaseRatio;
         float maxSignalBaselineRatio;
         int goodPeakCount;
         float groupRank;
 
+        //Model used in peakML.
+        string model;
+
+        float noiseRange;
+        float mayBeRange;
```
saifulbkhan: Do these three details need to be stored per-group? Even so, I would recommend saving them in the `MavenParameters` object associated to each group. The number of things that you need to change might seem a little excessive and error prone, if you have never added a new parameter/setting to El-MAVEN before. Feel free to ask me, if you need more details around this.

src/core/libmaven/csvreports.cpp
```diff
@@ -558,6 +569,72 @@ void CSVReports::_writePeakInfo(PeakGroup* group)
...
+        for (auto& group : groups) {
+            if (!group.isGhost())
+                writePeakInfo(group, group.groupId());
+        }
+        for (auto& group : groups) {
+            for (auto&child : group.childIsotopes()) {
+                if (!child->isGhost())
+                    writePeakInfo(*child, child.get()->groupId());
+            }
+        }
```
saifulbkhan: What if the group also has child adducts in it? They should also be exported just like child isotopes are.

src/core/libmaven/csvreports.cpp
```diff
@@ -604,103 +681,103 @@ void CSVReports::writeDataForPolly(const std::string& file,
...
+    //                     }
+    //                     break;
+    //                 }
+    //             }
+    //         }
+    //     }
+    //     inputGroupFile.close();
+    //     savedGroupFile.close();
+    //     remove("groupReport.csv");
+    // }
```
saifulbkhan, line 881: Commented out tests until the CSV report format is final, yes? Be sure to uncomment it and make any changes required, before this PR gets merged.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -125,6 +125,9 @@ MavenParameters::MavenParameters(string settingsPath):lastUsedSettingsPath(setti
         quantileSignalBlankRatio = 0.0;
         quantilePeakWidth = 0.0;
 
+        peakMl = false;
```
saifulbkhan: We can name it something more descriptive like `usePeakMl` or `classifyUsingPeakMl`, etc. Makes it more readable when this boolean is used in an `if` conditional.

src/core/libmaven/mavenparameters.h
```diff
@@ -271,6 +271,12 @@ class MavenParameters
         double overlapWeight;
         bool useOverlap;
 
+        //for enabling peak Ml curation
+        bool peakMl;
+        string peakMlModelType;
+        float badGroupLimit;
+        float maybeGoodGroupLimit;
```
saifulbkhan: I saw some values like this in `PeakGroup` class too. Maybe you should remove those attributes in the `PeakGroup` class and use values from each peak-group's `parameters` object.

Also why not have limits like `badGroupUpperLimit` and `goodGroupLowerLimit` and then anything in between those will be the "maybe" category? Somehow this seems like a better idea to me. But it is up to you, not necessary to change.

saifulbkhan: Also I do not see any initialization for these values in "mavenparameters.cpp" (constructor). They are also not being set anywhere. Did I miss them?

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -140,6 +152,24 @@ void BackgroundOpsThread::qtSignalSlot(const string& progressText,
...
+    //     for (auto& child : group.childIsotopes())
+    //         child->setUniqueId(++lastUniqueId);
+
+    if (mainwindow->mavenParameters->peakMl) 
+        classifyGroups(mavenParameters->allgroups);
+
+    emitGroups();
+
+    Q_EMIT(updateProgressBar("Done", 1, 1));
+}
```
saifulbkhan: I think I removed this function in v0.12.0. It had been there for way too long and for no reason. This is probably a rebase error. What you should do is remove this method and replace the calls to it (on lines 301 and 325) with:
```C++
    if (mainwindow->mavenParameters->peakMl) 
        classifyGroups(mavenParameters->allgroups);

    emitGroups();
```
That should do it.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -24,6 +24,11 @@
 #include "peakdetector.h"
 #include "samplertwidget.h"
 #include "EIC.h"
+#include "csvreports.h"
```
saifulbkhan, line 27: Why do we need to import CSV reporting functions here?

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+    if (groups.empty())
+        return;
+    
+    auto tempDir = QStandardPaths::writableLocation(
+                       QStandardPaths::GenericConfigLocation)
+                   + QDir::separator()
+                   + "ElMaven";
+
+    // TODO: binary name will keep changing and should not be hardcoded
+    // TODO: model should not exist anywhere on the filesystem
```
saifulbkhan: I think I added these. Are they two no longer true? If so, please remove them.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+#endif
+    QString mlModel;
+    // Following code is commented out because for now we have only one model in
+    // aws bucket. Once, we have more comments would be removed.  
+    // if(mainwindow->mavenParameters->peakMlModelType == "Global Model Elucidata"){
+        mlModel = tempDir + QDir::separator() + "model.pickle.dat";
+    // }
+    // else{
+    //     return;
+    // }
```
saifulbkhan: I am not aware of the complexities here, but do keep it in mind (and inform anybody who should know) that once this build is out there (in the form of a release), the behaviour cannot be changed on the fly. It can only go out as part of a new release.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+    
+    QString peakAttributesFile = tempDir
+                                 + QDir::separator()
+                                 + "peak_ml_input.csv";
+    QString classificationOutputFile = tempDir
+                                       + QDir::separator()
+                                       + "peak_ml_output.csv";
+
+    // // have to enumerated and assign each group with an ID, because multiple
+    // // groups at this point have the same ID
```
saifulbkhan: commented comments

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+    int startId = 1;
+    for (auto& group : groups) {
+        group.setGroupId(startId++);
+        for (auto& child : group.childIsotopes()) {
+            child->setGroupId(startId++);
+        }
```
saifulbkhan, line 466: This for-loop should come before the line (no. 452) where you assign a group ID to the parent group. Starting v0.12.0, the `PeakGroup::setGroupId` method also updates its `metaGroupId` as well as the `metaGroupId` of its child items. But since you are setting the group ID of children after you have set it for the parent, the metaGroupId will not match. Definitely ignore this advice if `metaGroupId` has no significance here.

But do add groupID to child adducts as well. Maybe that could be important since its happening for child isotopes.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+    QString badGroupLimit = QString::fromStdString(
+                                mzUtils::float2string(mavenParameters->badGroupLimit, 1));
+    QString maybeGoodGroupLimit = QString::fromStdString(
+                                    mzUtils::float2string(mavenParameters->maybeGoodGroupLimit, 1));
+    QStringList mlArguments;
+    mlArguments << "--input_attributes_file" << peakAttributesFile
+                << "--output_moi_file" << classificationOutputFile
+                << "--model_path" << mlModel
+                << "--bad_group_limit" << badGroupLimit
+                << "--maybeGood_group_limit" << maybeGoodGroupLimit;
```
saifulbkhan: Please do not mix camel-case and underscore-based naming. Choose one and stick to it. In this case please prefer underscores since that is already being used by other args.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+    int groupSize = groups.size();
+    int countGroups = 0;
+    for (auto& group : groups) {
+         if (mavenParameters->stop) {
+            mavenParameters->allgroups.clear();
+            terminate();
+            removeFiles();
+            return;
+        }
+        emit updateProgressBar("Classifying Peakgroups", countGroups++, groupSize);
```
saifulbkhan: Add ellipsis for indicating an ongoing process. So something like: "Classifying peak-groups…"

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+                                    group->addCorrelatedGroup(elem.first, elem.second);
+                                
+                            };
+    
+    int groupSize = groups.size();
+    int countGroups = 0;
+    for (auto& group : groups) {
+         if (mavenParameters->stop) {
+            mavenParameters->allgroups.clear();
+            terminate();
```
saifulbkhan: Why do you think the thread needs to be terminated here? What if the thread terminates before `removeFiles()` is invoked? Anyway, once you `return` from this function, the thread will stop properly, right?

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+            return;
+        }
+        emit updateProgressBar("Classifying Peakgroups", countGroups++, groupSize);
+        assignPrediction(&group, 
+                         predictions, 
+                         inferences, 
+                         correlations);
+    }
+    for (auto& group : groups) {
+        for (auto& child : group.childIsotopes()) {
```
saifulbkhan, line 660: Okay, one final time, most things that are happening for child isotopes also need to happen for child adducts. I know we wrote a lot of this code before v0.12.0. So, we should revisit such instances and make any relevant modifications.

I will try to keep an eye out for cases like this henceforth and mark them for your reference.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+        tempDir = tempDir + QDir::separator() + "model.pickle.dat";
+    }
+
+    auto cookieFile = QStandardPaths::writableLocation(
+                    QStandardPaths::GenericConfigLocation)
+                    + QDir::separator() 
+                    + "El-MAVEN_cookie.json" ;
+    
+    ifstream readCookie(cookieFile.toStdString());
+    json cookieInput = json::parse(readCookie);
```
saifulbkhan: This might be where you need to make a change to fix that "missing cookie file" crash. Check for `QFile::exists(cookieFile)` and do something if false.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -356,3 +388,393 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
...
+
+    auto res = _pollyIntegration->runQtProcess("downloadFilesForPeakML", args);
+    QString str(res[0].constData());
+    auto splitString = mzUtils::split(str.toStdString(), "\n");
+    auto data = splitString[splitString.size() - 2];
+    auto dataQstring = QString::fromStdString(data);
+    QJsonObject dataObj;
+
+    QJsonDocument doc = QJsonDocument::fromJson(dataQstring.toUtf8());
+    dataObj = doc.object();
```
saifulbkhan: So we are also using Qt JSON parsing along with the `nlohmann` library? I would have sticked to one (Qt is preferable if available and the files are not large).

src/gui/mzroll/backgroundopsthread.h
```diff
@@ -93,8 +99,18 @@ class BackgroundOpsThread : public QThread
      */
     void restoreAlignment();
 
+    /**
+     * @brief Some functions of classification runs outside
+     * El-MAVEN and cancellation at that point can not be 
+     * handled and hence signal to disable/enable cancel button
+     * in peak detector dialog.
+     */
+    void toggleCancel();
```
saifulbkhan, line 108: Nice.

src/gui/mzroll/QHistogramSlider.h
```diff
@@ -0,0 +1,84 @@
...
+    double maxVal;
+
+	double maxY;
+	double minY;
+
+    double maxBound;
+    double minBound;
+    QColor colBar;
+
+};
```
saifulbkhan: I am curious what is this widget? Where are we using it?

saifulbkhan (changes requested):
### Review part I

Added comments where I think some of the library-level code could be improved. Have yet to take a look at the front-end. Will do it in another sitting.

I did compile and try to test out basic functionality. Some comments here:
* The slider is not in the same line as the signal/noise range labels. It is also very glitchy when I drag any of the two slider controls. Clicking along the bar works well.
* The peak-detection dialog goes outside the bounds on my laptop screen (height-wise). Maybe we can have a separate PeakML tab in the detection dialog, even if it contains a handful elements. Anyhow, I feel some UI improvement can be made here.
* The application crashed for me when I set cohorts for 8 example files. I had already run one detection (with PeakML) once without setting the cohorts. Tried again, crashed on the first run - with samples cohorts set. Attached the crash log [here](https://hastebin.com/onudolagaj.yaml). Seems to be the same “missing cookie file” thing we talked about before.

Could not test further because of the crash happening every time now. Once you fix it, I will try out a few more things. Quite a bit has changed since I last worked on this branch.

src/gui/mzroll/QHistogramSlider.h
sakshikukreja14: Sorry about this one.. I just checked it out. And it seems like some silly addition maybe because of some rebasing. I will check it further to be sure.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Yup, I agree with you. will make the change.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Yup, it was on my to-do too. Adding a few commits tomorrow with this issue resolved.

sakshikukreja14:
Thank you  @saifulbkhan for your reviews. Sorry, for the repeated crash. I will make the required changes soon.

src/gui/mzroll/forms/infodialog.ui
```diff
@@ -25,6 +25,27 @@
...
+	background-color: rgb(117, 110, 210);
+	color: white;
+	border: 1px solid rgb(117, 110, 210);
+}
+
+QPushButton:focus {
+	outline: none;
+	border: 1px solid black;
+}</string>
+  </property>
```
saifulbkhan: I think we lost this style in v0.12.0. Please remove this style hunk. Probably a rebase error as well.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -431,6 +423,13 @@ using namespace mzUtils;
 
 	setWindowIcon(QIcon(":/images/icon.png"));
 
+    //seeting correlation table
```
saifulbkhan: * seeting -> setting

src/gui/mzroll/loginform.cpp
```diff
@@ -10,14 +10,16 @@
...
+LoginForm::LoginForm(PollyElmavenInterfaceDialog* pollyelmaveninterfacedialog, bool showPollyApp) :
     QDialog(),
     _aboutPolly(NULL),
     ui(new Ui::LoginForm)
     
 {
     _pollyelmaveninterfacedialog = pollyelmaveninterfacedialog;
     _pollyintegration = _pollyelmaveninterfacedialog->getMainWindow()->getController()->iPolly;
+
+    _showPollyApp = showPollyApp;
```
saifulbkhan: Suggestion: maybe a more descriptive name for this flag would be `showEPI`.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -1038,6 +1037,20 @@ void MainWindow::threadSave(const QString filename, const bool saveRawData)
...
 }
 
+void MainWindow::noActiveInternet() {
+    QMessageBox *warning = new QMessageBox(this);
+            
+    auto htmlText = QString("<p><b>Failed to download required files. Check for internet "
+                            " connection and try again later.</b></p>");
+    htmlText += "<p>Please contact tech support at elmaven@elucidata.io if the problem persists.</p>";
+    warning->setText(htmlText);
+    warning->setIcon(QMessageBox::Icon::Critical);
```
saifulbkhan: If this is a warning, then we should be showing a "Warning" icon. Critical messages indicate that something really bad happened (e.g. major system failure or crash).

src/gui/mzroll/mainwindow.h
```diff
@@ -349,14 +346,16 @@ public Q_SLOTS:
     void markGroup(shared_ptr<PeakGroup> group, char label);
     int getIonizationMode();
 	void setTotalCharge();
+    void noActiveInternet();
```
saifulbkhan: Are you sure the main-window is the best place for a slot like this? Reading this would give me the impression that this is an application-wide utility. But it is not - since it is only used for notifying about how PeakML files were not downloaded. I would suggest you either move this to peak-detection dialog or at least make  the `MainWindow::noActiveInternet` slot generic enough to be able to show any configurable message - since slots can take any number of inputs.

src/gui/mzroll/mzroll.pro
```diff
@@ -79,6 +79,7 @@ CONFIG(release, debug|release) {
 INCLUDEPATH +=  /usr/include/x86_64-linux-gnu/qt5/QtXml/ /usr/include/x86_64-linux-gnu/qt5/QtSql
 
 INCLUDEPATH +=  $$top_srcdir/src/core/libmaven  \
+                $$top_srcdir/src/common         \
```
saifulbkhan: We had not added this include path earlier because all headers from the "common" namespace were included using `#include common/headername.h`. It was more indicative of the nature of library. But since we are now adding this new path, let's make sure that the include patterns are not a mix of both (i.e. - check if some places have "`common/headername.h`" and some just "`headername.h`").

src/gui/mzroll/numeric_treewidgetitem.cpp
```diff
@@ -1,15 +1,37 @@
...
+}
 
 bool NumericTreeWidgetItem::operator<( const QTreeWidgetItem & other ) const
 {
     int sortCol = treeWidget()->sortColumn();
 
-    // hack to prevent sorting of child tree-widget items (such as isotopes that
-    // need to have a fixed sort order
-    if (parent() != nullptr && sortCol != 1)
-        return false;
```
saifulbkhan: Seems like another rebase issue. Please include this red diff too, it is important for sorting isotope/adduct child rows. It should be just below line no. 7 and above the newly added logic for sorting based on PeakML labels.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -82,8 +93,8 @@ void PeakDetectionSettings::updatePeakSettings(string key, string value)
         if(QString(v.typeName()).contains("QGroupBox"))
             v.value<QGroupBox*>()->setChecked(std::stod(value));
 
-        if(QString(v.typeName()).contains("QCheckBox"))
-            v.value<QCheckBox*>()->setChecked(std::stod(value));
+/*        if(QString(v.typeName()).contains("QCheckBox"))
+            v.value<QCheckBox*>()->setChecked(std::stod(value));*/
```
saifulbkhan: Why did we comment this out? With this, checkboxes in peak-detection dialog will not be synced anymore from default settings or when loading settings from emDB files.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -183,6 +196,21 @@ PeakDetectionDialog::PeakDetectionDialog(MainWindow* parent) :
...
+                [this](const bool checked)
+                {
+                    if(checked){
+                        getLoginForPeakMl();
+                    }
+                    else{
+                        peakMlSet = false;
+                        mainwindow->mavenParameters->peakMl = false;
+                        modelTypes->setEnabled(false);
+                    }
```
saifulbkhan: Let's add analytics here too? Mixpanel would be preferable over GA. Since it gives us a better idea of who was interested in PeakML. But as I mentioned before, the current label for this group-box does not really hint at some cool ML tech.

src/gui/mzroll/peakdetectiondialog.cpp
```diff
@@ -417,6 +506,62 @@ void PeakDetectionDialog::inputInitialValuesPeakDetectionDialog() {
...
+    QString str(res[0].constData());
+    if (str.isEmpty())
+        return;
+    auto splitString = mzUtils::split(str.toStdString(), "\n");
+    auto data = splitString[splitString.size() - 2];
+    auto dataQstring = QString::fromStdString(data);
+    QJsonObject dataObj;
+
+    QJsonDocument doc = QJsonDocument::fromJson(dataQstring.toUtf8());
+    dataObj = doc.object();
```
saifulbkhan: I remember seeing some very similar logic in `BackgroundOpsThread` class. Is it possible to share it? Anyhow, you should consider moving this whole work to the background thread. Reading and parsing data is a potentially blocking operation and should not be part of the main thread. Instead you could show a status string here (which I think happens post-detection when peaks are about to be classified). Or you could totally ignore all of this if you are sure that the files will never be large enough to cause an issue.

src/gui/mzroll/peakdetectiondialog.h
```diff
@@ -67,6 +80,8 @@ class PeakDetectionDialog : public QDialog, public Ui_PeakDetectionDialog
                     void findPeaksClicked();
         public:
                 QString massCutoffType;
+                bool peakMlSet;
+                RangeSlider* slider;
```
saifulbkhan: Is there a reason to keep these two public?

src/gui/mzroll/superSlider.h
```diff
@@ -0,0 +1,86 @@
+#ifndef SUPERSLIDER_H
+#define SUPERSLIDER_H
+#pragma once
+
+#include <QWidget>
+#include <QPainter>
+#include <QMouseEvent>
+
+class RangeSlider : public QWidget
```
saifulbkhan: SuperSlider is a nice name. But then if the name of the class within this file is `RangeSlider`, then the filename and include guards should have the same names. Please chose one or the other. I like `RangeSlider` better - conveys exactly what it is.

saifulbkhan (changes requested):
Review part II:
* I was able to get it to not crash when I logged out of EPI and then logged back in when selecting “Peak curation” before doing peak detection. Also is “Peak curation” the right thing to call it? Should it not be something like “Use PeakML to classify peaks”? Maybe get some input from Richa and Surbhi on this.
* Even when I was able to get it to work without crashing I do not get any classification. The “Label” column was blank for all peak-groups.
* When PeakML is not able to classify anything I get “nan%” values in the label-filtering dropdown in peak tables. It should instead display “NA” or “0%”.
* Another thing that came to my mind: are we taking care of saving enough state in the emDB sessions? Such that reloading the file restores all of the data and state as it was when saved? UI state positioning and visibility excluded, of course.

Also added some more comments on the code. Please make changes wherever required.

Still need to review 15 or so file. Will do in another sitting, next week. Hopefully you will have made changes to fix some of the blocking issues.

src/core/libmaven/PeakGroup.cpp
sakshikukreja14: Basically, we have changed the export options in CSV. One of them is exporting 'displayed groups', (which is one of the functions in csvreports.cpp, the core library) hence this flag is used there to get information if is displayed or not. 
I will make it private and use it accordingly.

src/gui/mzroll/peakdetectiondialog.cpp
sakshikukreja14: The reason to use the same logic here is that we need to change the content of the dropdown (select models) on the basis of the Polly credentials. This is done so that the user can choose one model out of many to run PeakML.

src/core/libmaven/csvreports.cpp
sakshikukreja14: Sure, Once all the comments are resolved I will uncomment these tests.

src/core/libmaven/mavenparameters.h
sakshikukreja14: Sure will remove these from the peak-group class. Also, these values are set through peak detection dialog using slider

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Yes, it was for testing purpose and would be removed.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Because the input file for peakML exported in this class is created using csv reports

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Done :)

.travis.yml
```diff
@@ -87,7 +87,7 @@ script:
 
         # export encoded update repository URL
         export UPDATE_REPO_URL_BASE64=`echo $UPDATE_REPO_URL | base64`
-
+
```
saifulbkhan: This is the only change in this file :P

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1059,91 @@ void PeakGroup::setGroupId(int groupId)
...
+    _userLabel = label;
+
+    if (parent != nullptr && parent->tagString == C12_PARENT_LABEL) {
+        parent->setUserLabel(label);
+        return;
+    }
+
+    for (auto& child : _childIsotopes) {
+            child->setUserLabel(label);
+    }
```
saifulbkhan: As it is now, here is what I think will happen:
* If you mark an child isotope/adduct as good/bad or unmark it, then its parent will also be marked as good/bad/unmarked, right?
* When the parent's label is being set, the second loop will run and set all its child items to have the same label as the parent.

Are those things what should be happening? I think not. In isotopic labelled studies, you can very easily have a non-parent child group that has higher quality while the parent is a non-existent/bad peak-group.

In any case, you can discuss this with Richa on what should be the expected behavior. In my opinion, only mark the peak-group that is currently selected and leave its parent/siblings as they are.

src/core/libmaven/PeakGroup.cpp
```diff
@@ -1069,3 +1059,91 @@ void PeakGroup::setGroupId(int groupId)
...
+
+    if (parent != nullptr && parent->tagString == C12_PARENT_LABEL) {
+        parent->setPredictedLabel(label, probability);
+        return;
+    }
+    for (auto& child : _childIsotopes) {
+        if (child->_predictedLabel != label
+            && child->_predictionProbability != probability)
+            child->setPredictedLabel(label, probability);
+    }
```
saifulbkhan: Same thing here as above. Only difference is, the model is now marking peak-groups instead of a user. The behavior should be consistent in my opinion.

src/core/libmaven/csvreports.cpp
```diff
@@ -558,6 +569,76 @@ void CSVReports::_writePeakInfo(PeakGroup* group)
...
+             << "groupOverlap" << "\n";
+        file << fixed << setprecision(6);
+
+        for (auto& group : groups) {
+            if (!group.isGhost())
+                writePeakInfo(group, group.groupId());
+        }
+        for (auto& group : groups) {
+            for (auto&child : group.childIsotopes()) {
+                if (!child->isGhost())
```
saifulbkhan, line 725: Child groups are never ghosts. But okay, no harm here.

src/core/libmaven/mavenparameters.cpp
```diff
@@ -309,6 +314,10 @@ void MavenParameters::copyFrom(const MavenParameters& mp)
     // vectors are meant to be used only by the global `MavenParameters` object,
     // which will not be created by a copy operation like this.
     samples = mp.samples;
+
+    peakMlModelType = mp.peakMlModelType;
+    badGroupUpperLimit = mp.badGroupUpperLimit;
+    goodGroupLowerLimit = mp.goodGroupLowerLimit;
```
saifulbkhan: Why not also copy the `MavenParameters::classifyUsingPeakMl` attribute? Makes sense to do it and you can use it to selectively show or hide PeakML attributes in "setting log" widget.

src/gui/mzroll/backgroundopsthread.cpp
saifulbkhan: This is still there.

src/gui/mzroll/peakdetectiondialog.cpp
saifulbkhan: Hm, yes and the main "fetch model list" logic can still be moved to `BackgroundOpsThread` to keep all such methods in one place. But it's up to you. I do not mind repetitions in two places (I would if it was repeated thrice).

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -23,6 +24,8 @@ PollyElmavenInterfaceDialog::PollyElmavenInterfaceDialog(MainWindow* mw)
     qRegisterMetaType<PollyApp>("PollyApp");
     qRegisterMetaType<QMap<PollyApp, bool>>();
 
+    showPollyApps = true;
```
saifulbkhan: Is this ever set to `false`? If not, I would suggest making the default value of the second argument to the constructor of `LoginForm` to be `true`. This would make its default behavior to be what it used to and also prevent you from calling it with a flag everytime, throughout this file. A data member (`showPollyApps`) that never changes is not needed.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -195,8 +198,9 @@ PollyElmavenInterfaceDialog::PollyElmavenInterfaceDialog(MainWindow* mw)
             &QTabWidget::currentChanged,
             this,
             &PollyElmavenInterfaceDialog::_changeMode);
+    connect(this, SIGNAL(loginResponse()), _mainwindow->peakDetectionDialog, SLOT(loginSuccessful()));
+    connect(this, SIGNAL(loginUnsuccessful()), _mainwindow->peakDetectionDialog, SLOT(unsuccessfulLogin()));
```
saifulbkhan: Please use the new signal-slot syntax for all newly written code. It is type-safe and fails fast.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -419,6 +423,27 @@ void PollyElmavenInterfaceDialog::_enableExistingProjectUi()
...
+void PollyElmavenInterfaceDialog::loginForPeakMl()
+{
+    try {
+        _callLoginForm(false);
+    } catch (...) {
+        QMessageBox msgBox(this);
+        msgBox.setWindowModality(Qt::NonModal);
+        msgBox.setWindowTitle("Error in loading login form");
+        msgBox.exec();
+    }
```
saifulbkhan, line 239: Why did we need a try catch here? The login form has it's own error reporting capabilities, right?

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -1087,9 +1113,9 @@ QStringList PollyElmavenInterfaceDialog::_prepareFilesToUpload(QDir qdir,
                             + QDir::separator()
                             + datetimestamp
                             + "_Peaks_information_json_Elmaven_Polly.json";
+    peakTable->exportJsonToPolly(_writeableTempDir, jsonFilename, true);
 
     if (_selectedApp == PollyApp::PollyPhi) {
-        peakTable->exportJsonToPolly(_writeableTempDir, jsonFilename, true);
```
saifulbkhan: Why did this move outside the `if` block?

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -419,6 +423,27 @@ void PollyElmavenInterfaceDialog::_enableExistingProjectUi()
...
+}
+
+void PollyElmavenInterfaceDialog::emitLoginReady()
+{
+    emit loginResponse();
+}
+void PollyElmavenInterfaceDialog::loginFormClosed()
+{
+    emit loginUnsuccessful();
+}
```
saifulbkhan: Hmm... I see these being called as a slot by `LoginForm`. And then these signals are themselves used to trigger slots in `PeakDetectionDialog`. Why not directly connect the `LoginForm` signals to `PeakDetectionDialog` slots? Eliminate this seeming unnecessary layer?

src/gui/mzroll/rangeSlider.h
```diff
@@ -0,0 +1,86 @@
+#ifndef RANGESLIDER_H
+#define RANGESLIDER_H
+#pragma once
```
saifulbkhan: I know you borrowed this code from an external source, but let's stick to include guards for consistency and drop the pragma.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -211,8 +327,15 @@ shared_ptr<PeakGroup> TableDockWidget::groupForItem(QTreeWidgetItem *item)
 {
   if (item == nullptr)
     return nullptr;
+  
+  int index;
+  if (hasClassifiedGroups)
+    index = 1;
+  else
+    index = 0;
```
saifulbkhan: Set `index` with a default value of `0` and then if `hasClassifiedGroups` is true, then set it to `1`. Ensures default init value, conveys the same intent and lesser code.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -296,89 +517,140 @@ void TableDockWidget::updateItem(QTreeWidgetItem *item, bool updateChildren) {
...
+    item->setText(index++, QString::number(group->fragMatchScore.mergedScore, 'f', 2));
+    item->setText(index++, QString::number(group->ms2EventCount));
+    item->setText(index++, QString::number(group->predictionProbability(), 'f', 2));
+    item->setText(index++, QString::number(group->groupRank, 'e', 6));
 
     if (fabs(group->changeFoldRatio) >= 0) {
-      item->setText(15, QString::number(group->changeFoldRatio, 'f', 3));
-      item->setText(16, QString::number(group->changePValue, 'f', 6));
+      item->setText(index++, QString::number(group->changeFoldRatio, 'f', 3));
+      item->setText(index++, QString::number(group->changePValue, 'f', 6));
```
saifulbkhan: Really good stuff! Using an incrementing index to set items instead of absolute values. Why did I never thing of this...

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -678,51 +959,171 @@ void TableDockWidget::showAllGroups() {
...
   QMap<int, QTreeWidgetItem *> parents;
   for (size_t i = 0; i < _topLevelGroups.size(); ++i) {
     auto group = _topLevelGroups[i];
+
+    // Setting curation parameters;
+    if (hasClassifiedGroups) {
+      group->isClassified = true;
+      for (auto childIsotope : group->childIsotopes()) {
+        childIsotope->isClassified = true;
+      }
```
saifulbkhan: Child adducts?

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -855,12 +1256,22 @@ void TableDockWidget::exportGroupsToSpreadsheet() {
                         ms2GroupExists,
                         includeSetNamesLines,
                         _mainwindow->mavenParameters);
-  QList<shared_ptr<PeakGroup>> selectedGroups = getSelectedGroups();
+  QList<shared_ptr<PeakGroup>> selectedGroups;
+  
+  if (static_cast<int>(peakTableSelection) == 10)
+    selectedGroups = getDisplayedGroups();
+  else
+    selectedGroups = getSelectedGroups();
```
saifulbkhan: Does that `10` mean something special here? Please use a constant or at least add a comment.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1471,6 +2102,30 @@ void TableDockWidget::unmarkGroup() {
...
+//       }
+//       totalCount++;
+//     }
+
+//     if (totalCount == goodCount + badCount)
+//       return true;
+//   }
+
+//   return false;
+// }
```
saifulbkhan: No longer used? Why not remove it entirely?

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1799,7 +2482,17 @@ void TableDockWidget::contextMenuEvent(QContextMenuEvent *event) {
...
 
-  menu.exec(event->globalPos());
+  if(hasClassifiedGroups)
+  {
+      QAction *z8 = menu.addAction("Explain classification");
+      classificationWidget = new ClassificationWidget(this);
+      connect(z8,
+              SIGNAL(triggered()),
+              classificationWidget,
+          SLOT(showClassification()));
```
saifulbkhan: Use type-safe syntax for signal-slot.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1865,6 +2560,84 @@ void TableDockWidget::writeMascotGeneric(QString filename) {
...
+              this,
+              &TableDockWidget::filterForSelectedLabels);
+  
+  if (changedBadLimit == badGroupLimit 
+      && changedMaybeGoodLimit == maybeGoodGroupLimit) {
+    QMessageBox::warning(this,
+                          "No Change!",
+                          "The curation parameters you entered are same as" 
+                          " in classified table.\nYou may change the range"
+                          " and try again.\n");
```
saifulbkhan: "The curation parameters you entered are the same as they were in the classified peak-table."

"You may change the range and try again."

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1865,6 +2560,84 @@ void TableDockWidget::writeMascotGeneric(QString filename) {
...
+                                      }
+                                      
+                                }
+                          };
+
+  auto groups = _topLevelGroups;
+  int totalSteps = groups.size();
+  int steps = 1;
+  for (auto &group : groups) {
+    Q_EMIT(updateRelabelStatusBar("Relabeling", steps, totalSteps));
```
saifulbkhan: "Relabeling…"

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -2827,6 +3658,7 @@ void ScatterplotTableDockWidget::setupPeakTable() {
     colNames << "Max quality";
     colNames << "MS2 score";
     colNames << "#MS2 events";
+    colNames << "Probability"; // TODO: add this column conditionally
```
saifulbkhan: Don't you think we should name this "Classification probability" instead? Ignore if this decision was made after discussion already.

src/pollyCLI/pollyintegration.cpp
```diff
@@ -35,7 +36,7 @@ PollyIntegration::PollyIntegration(DownloadManager* dlManager):
       nodeModulesPath = binDir + "node_modules" + QDir::separator();
     #endif
 
-    indexFileURL = "https://raw.githubusercontent.com/ElucidataInc/polly-cli/master/prod/index.js";
+    indexFileURL = "https://raw.githubusercontent.com/sakshikukreja14/polly-cli/ftr_moi_api/prod/index.js";
```
saifulbkhan, line 43: I am going to keep this comment up so that the branch does not get merged without this change.

saifulbkhan (changes requested):
Review part III - leaving some more comments here. Will do user-based testing now based on the feature document that you shared earlier. Might do this today itself.

saifulbkhan:
@sakshikukreja14 A few comments from previous reviews have not been resolved because I do not see the requested changes. Do take a look.

sakshikukreja14:
@saifulbkhan sure will resolve the comments soon. Sorry, for missing out on some previously.

src/core/libmaven/PeakGroup.cpp
sakshikukreja14: Yes, The child peak groups and child adducts are now classified separately. So, we would be requiring this piece of code. Removing this.

src/core/libmaven/PeakGroup.cpp
sakshikukreja14: Exactly, I feel the same that child peak-group behaviour must be independent of parent/siblings. Setting the user label is outside peakML. Removing this piece of code too. So, that it gives the same functionality as it was in v0.12

src/gui/mzroll/tabledockwidget.cpp
sakshikukreja14: Added.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Removed.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Made the required changes.

src/gui/mzroll/tabledockwidget.cpp
sakshikukreja14: Sure..Will discuss it with Richa and change it accordingly.

src/gui/mzroll/pollyelmaveninterface.cpp
sakshikukreja14: Yes, but those include if the user is not authorized or email Id not given. This try-catch block reports if for some reason the login form fails to appear (the probability of which is very low).

src/gui/mzroll/pollyelmaveninterface.cpp
sakshikukreja14: Actually, this was being used while calling the log-in form and it was set to false whenever a log-in form is called through peak detection dialog. 
To get the same effect, I have removed it from the constructor and added it as a default argument in the function it was needed.

src/gui/mzroll/peakdetectiondialog.cpp
sakshikukreja14: Haven't added it till now, was facing an issue. It will be resolved asap.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -267,8 +279,14 @@ void BackgroundOpsThread::computePeaks()
...
         peakDetector->mavenParameters()->searchAdducts = hadSearchAdducts;
     }
+    
+    if (mainwindow->mavenParameters->classifyUsingPeakMl) 
+        classifyGroups(mavenParameters->allgroups);
+
     emitGroups();
 
+    Q_EMIT(updateProgressBar("Done", 1, 1));
+
```
saifulbkhan: This should not be here. The emit will be followed by another emit in the next line, effectively making it useless.

src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -290,8 +308,14 @@ void BackgroundOpsThread::findFeatures()
...
```
saifulbkhan: Same as above.

src/gui/mzroll/backgroundopsthread.cpp
saifulbkhan: Was also referring to `--maybeGood_group_limit` arg.

src/gui/mzroll/pollyelmaveninterface.cpp
```diff
@@ -13,6 +13,7 @@
 #include "pollywaitdialog.h"
 #include "projectdockwidget.h"
 #include "tabledockwidget.h"
+#include "peakdetectiondialog.h"
```
saifulbkhan, line 16: This is no longer needed here now that we have moved the slot connection to login-form, right?

src/gui/mzroll/pollyelmaveninterface.h
```diff
@@ -191,6 +204,10 @@ public Q_SLOTS:
      */
     void uploadFinished(bool success);
 
+    void loginResponse();
+
+    void loginUnsuccessful();
+
```
saifulbkhan: These two should be removed now.

src/gui/mzroll/backgroundopsthread.cpp
sakshikukreja14: Ohhh! Okay. To change this argument, I would need to make a change in the MOI file as the current moi takes this as an argument. 
Can we please let it be for this iteration??? As it is uploaded to Aws and folks are using it already

saifulbkhan: Oh no problem. This is just a minor thing anyway.

saifulbkhan:
@sakshikukreja14 Adding another usability review for PeakML:
* Peak detection dialog comes up with “Polly-PeakML” tab selected by default. Should be “Detection method”.
* The “Preparing inputs for classification…” step is very slow for large peak-count. Is it the write-to-CSV part? I am on an SSD. Normal hard-disks can frustrate user. Eventually I force quit the application - not sure if this was just stuck or slow.
* Tried the same dataset again in a fresh session and it worked very fast this time.
* Sorting by label column does not work. Is that intentional?
* Cycling through the “correlated peak-groups” worked well. But the currently selected peak-group in the peak-table was always the first entry in the correlation table. So if I cycle between masses 99, 100, 101 and 102 and the current mass selected is 101, then the order in the correlation table is 101, 99, 100, 102. Would be nice to have them always sorted by mass. The cycling itself is according to mass (which is the right behavior). So there is some inconsistency.
* I edited one of the maybe peak-groups to include some of the missed peak areas and then marked it as “good” manually. Then pressed Cmd+z to see if undo is working. It worked the first time but was really slow (UI froze for about 5 seconds). Then I marked it as good again and then again tried undoing the marking this time it froze for 5 seconds again but did not undo my manually set “good” mark. Tried undo (pressed Cmd+z) one more time and then it worked. This was a table with less-than 200 peak-groups.
* The force plot can do a little better when it comes to drawing labels for contributing attributes. In the example below, the markers are overlapping with the text (on the blue side) but there’s ample space to the right that can be used to position them without any overlaps.
Screenshot 2020-12-23 at 8 25 19 AM

* I am able to select multiple peak-groups in the table and then explain their classification (but the force plot only comes up for the last selected group). We should disable the “Explain classification” menu item during multi-selection.
* After relabeling the peak-table once or twice using new ranges, the context-menu (that comes up on right-click) for peak-groups no longer came up again for that table. It worked fine on any newly created table. Was not able to reproduce this bug again. Might be a one off case.

Everything else seems to be working as it should. The QA team might have some more feedback since they will probably play around with it more than I could.

PR #1369, opened 2021-01-07
Fix - untargeted memory leak

saifulbkhan:
Various small fixes (with decent saves) to memory consumption during untargeted (and somewhat DB search) detections.

@sunil20dhakad This should resolve most of your concerns. @sakshikukreja14 will share the `ftr_peak_ml` branch rebased on this, once we merge it.

### DIFF
--- a/src/core/libmaven/peakdetector.cpp
+++ b/src/core/libmaven/peakdetector.cpp
@@ -1082,6 +1082,7 @@ void PeakDetector::detectIsotopesForParent(PeakGroup& parentGroup,
                                                    _mavenParameters,
                                                    _mavenParameters->clsf,
                                                    true);
+            delete_all(eics);
             if (isotopeGroup->peakCount() == 0)
                 continue;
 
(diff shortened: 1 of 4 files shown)


src/core/libmaven/peakdetector.cpp
```diff
@@ -571,7 +571,7 @@ void PeakDetector::processSlices(vector<mzSlice*>& slices,
             return;
 
         _mavenParameters->allgroups.insert(
-            _mavenParameters->allgroups.begin(),
+            _mavenParameters->allgroups.end(),
```
sakshikukreja14, line 574: Does this show any affect on memory leak??

saifulbkhan: Oh yes! Surprisingly this is the one that gave us 85% of the save. Also decreased peak-detection time by some amount but I did not measure it.

The 'insert at begin' thing was meant for when rank-based filtering happened in this method. But since v0.12.0, I moved it out to the meta-grouping stage and this insertion remained there. It is no longer required to do it this way and appending a vector at the end of a vector is much more efficient (both in terms of memory and CPU time).

sakshikukreja14 (approved):

Issue #1351, opened 2020-11-16
does "maximum retention time difference between peaks in a group" do what it sounds like?


chubukov:
I was imagining that this setting would be a hard filter on whether these were called one group, but that doesn't seem to be happening. Do I not understand the parameter well, or is it a bug?

This happens with automated peak detection but not with manual EIC selection.

![image](https://user-images.githubusercontent.com/13080968/99204575-1a93eb00-2784-11eb-8ce2-c34a9dc059a1.png)

![image](https://user-images.githubusercontent.com/13080968/99204618-3dbe9a80-2784-11eb-908e-4bbcdcdaa6b8.png)

chubukov:
Looking at this again I wonder if the settings are being used at all, because even the EIC smoothing window should prevent these being grouped I would think. (And there are other examples that look similar)

saifulbkhan:
@chubukov Definitely a bug. And I have noticed this myself (once during a recent demo, embarrassingly enough).

I think this filter is not properly used when the "Consider overlap" setting is checked (in "Grouping score" settings). Here's the annotated code for these checks:
```C++
if (mp->useOverlap) {
    // 'a' and 'b' are two peaks from an EIC, being considered for grouping
    // 'overlap' is the fraction of x-region shared by both peaks (so, 0.0 < overlap < 1.0)
    if (overlap == 0 and a.rtmax < b.rtmin)
        continue;
    if (overlap == 0 and a.rtmin > b.rtmax)
        break;

    // 'distx' is difference between the RT at apexes of the two peaks
    // The bug is maybe here? If either condition is false, grouping is allowed.
    if (distx > mp->grouping_maxRtWindow && overlap < 0.2)
        continue;

    // This score being assigned means 'a' and 'b' are valid candidates for grouping.
    score = 1.0
            / (mp->distXWeight * distx + 0.01)
            / (mp->distYWeight * disty + 0.01)
            * (mp->overlapWeight * overlap);
} else {
    // This check here seems to be working as it should - when "Consider overlap" is not checked.
    if (distx > mp->grouping_maxRtWindow)
        continue;

    score = 1.0
            / (mp->distXWeight * distx + 0.01)
            / (mp->distYWeight * disty + 0.01);
}

```

Not sure how long this has been the case. If it was this severe, somebody should have noticed it sooner. So far, the fix looks simple. Changing that `&&` to an `||` should give expected results, i.e., do not group two peaks if:
1. distance between two peaks is too high
or
[...]

PR #1370, opened 2021-01-08
[Fix] RT deviation abused when checking overlap #1351

saifulbkhan:
When the user opts to "Consider overlap" in peak grouping, the condition should make the grouping criteria more strict. However, the logic for group scoring was written such that peaks were allowed to be grouped if either RT deviation was not too bad or group overlap was not too bad (which made it more lenient). This has been fixed.


sakshikukreja14:
@saiful-elucidata, I think we would need to alter the test cases. Doctest fails in all the environments.

saif-el:
@sakshikukreja14 I have fixed both the failing tests. You can now review this one.

sakshikukreja14 (approved):

PR #1371, opened 2021-01-11
Fix VIB issues

saifulbkhan:
The follow changes were made:
* Allow RT range sync for siblings-only meta-groups - isotopes under ghost parents were not being synced when instructed.
* Allow auto-marking of duplicates as bad - allows the user to review duplicates of a given compound before exporting.
* Allow zeroed peak groups to be added to table - helpful when loading an emDB that contains peak-groups with no peaks.
* Show and allow setting explicit RT bounds in peak-editor - two editables will now show the current lower and upper range.

Note: this branch may get some minor changes as per feedback. Not ready for review yet.

### DIFF
--- a/src/gui/mzroll/gallerywidget.cpp
+++ b/src/gui/mzroll/gallerywidget.cpp
@@ -32,6 +32,12 @@ GalleryWidget::GalleryWidget(QWidget* parent)
     recursionCheck = false;
     _plotItems.clear();
 
+    _leftEdit = nullptr;
+    _leftProxyEdit = nullptr;
+    _rightEdit = nullptr;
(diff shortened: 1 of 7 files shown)


src/gui/mzroll/gallerywidget.cpp
```diff
@@ -217,9 +227,57 @@ void GalleryWidget::replot()
     recursionCheck = false;
 }
 
+void GalleryWidget::_drawBoundaryEditables(float rt1,
```
sakshikukreja14, line 230: This is really cool to have.

sakshikukreja14 (approved):

PR #1379, opened 2021-02-16
[FIX] fixed the crashing of app when ms2 files are loaded and peak detected

ghost:
We were accessing the isParent function of an adduct even when adduct is not created so it was giving the segmentation fault.
Solution was to check if the group is adduct then call the function.


saif-el:
@MdAzmatAli Thanks for working on this. But I think there was miscommunication from my end and I forgot to update here. This same bug was fixed in another PR that was merged recently. You can see [this commit](https://github.com/ElucidataInc/ElMaven/commit/3fda620809ab115a7bb6f7c7461397fe54687a39) which does nearly the exact same thing as you did here.

I will go ahead and close this PR.

Issue #1375, opened 2021-02-03
Elmaven crashes when file path contains non-English characters


LiChenPU:
**Describe the bug**
Elmaven cannot recognize a path if it contains characters from other languages (e.g. put into a folder whose name contains other language’s characters). It crashed and threw out error message (see screenshots).
![crash](https://user-images.githubusercontent.com/35385956/106758370-6449ee80-666c-11eb-9e3e-3dc1cafa750c.png)

**To Reproduce**
1. Create a new folder, with name "新建文件夹", meaning "New folder" in Chinese
2. Put a mzXML file into the folder
3. Use Elmaven to open the mzXML file
4. Elmaven crashes

**Expected behavior**
It should not crash.

**Screenshots**

**System (please complete the following information):**
 - Windows 10
 - 0.12.0

saif-el:
Thanks for reporting @LiChenPU. We will try to include this in our next release.

PR #1381, opened 2021-02-18
[FIX] fixed the crashing of app when folder or file contains non-English characters

ghost:
issue can be found at #1375


sakshikukreja14 (approved):
Nice Fix, Works fine.

saif-el (approved):

saif-el:
@MdAzmatAli Thanks for fixing. Didn't realize it would be this simple. Unicode magic. Can you rebase this branch on the latest `develop`? I will merge it then.

Issue #1286, opened 2020-04-20
El-MAVEN crashes with few mzXML files in the datasets while works for other 


Shreya-crypto:
[http://genomics-pubs.princeton.edu/mzroll/index.php?show=download](url)
The KFP 20 MIN TIME COURSE EXAMPLE DATASET : Download this and load the files to El-MAVEN. It would instantly crash.
While if 4 from one cohort and 4 from other are added, It would work fine. Size of the files isnt seem to be an issue as well

PR #1383, opened 2021-02-23
[FIX] fix the crashing of app when corrupted files are loaded

ghost:
The respective issue can be found at #1286 
Crashing of app is fixed and added the proper pop-up message.

### DIFF
--- a/src/core/libmaven/mzSample.cpp
+++ b/src/core/libmaven/mzSample.cpp
@@ -810,8 +810,7 @@ void mzSample::parseMzXML(const char* filename)
         setInstrumentSettigs(doc, spectrumstore);
         // parse mzXML information from the scan
         parseMzXMLData(spectrumstore);
-    } else
-        throw MavenException(ErrorMsg::ParsemzXml);
+    }
 }
 
 /**
--- a/src/gui/mzroll/mainwindow.cpp
+++ b/src/gui/mzroll/mainwindow.cpp
@@ -683,7 +683,11 @@ using namespace mzUtils;
                     msgBox.setInformativeText("Please make sure your file "
                                               "import settings do not conflict "
                                               "with the type of samples being "
-                                              "loaded.");
+                                              "loaded.\n\nIf your settings are "
+                                              "correct, then it could be that "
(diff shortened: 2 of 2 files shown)


src/core/libmaven/mzSample.cpp
```diff
@@ -810,8 +810,7 @@ void mzSample::parseMzXML(const char* filename)
         setInstrumentSettigs(doc, spectrumstore);
         // parse mzXML information from the scan
         parseMzXMLData(spectrumstore);
-    } else
-        throw MavenException(ErrorMsg::ParsemzXml);
```
sakshikukreja14, line 814: @MdAzmatAli  why is this exception removed??? Wasn't it getting handled somewhere? I see this exception is thrown in other such functions too. This means those would also lead to a crash if the conditions match.

src/gui/mzroll/mainwindow.cpp
```diff
@@ -683,7 +683,11 @@ using namespace mzUtils;
                     msgBox.setInformativeText("Please make sure your file "
                                               "import settings do not conflict "
                                               "with the type of samples being "
-                                              "loaded.");
+                                              "loaded.\n\nIf your settings are "
```
sakshikukreja14: your file import* settings

src/core/libmaven/mzSample.cpp
saif-el: @MdAzmatAli Sakshi is right. These exceptions are thrown in a few other places in mzSample.cpp. They should ideally be caught and informed about the user on what we were expecting and what went wrong.

For example, in mzfileio.cpp, we [catch memory allocation exception](https://github.com/ElucidataInc/ElMaven/blob/develop/src/gui/mzroll/mzfileio.cpp#L82-L88) when user runs out of RAM while loading samples. Something similar should happen in this case as well - catch the specific type of exception (`ErrorMsg::ParsemzXml`) and then tell user that something is wrong with the file.

saif-el:
And why is CI complaining about a C++17 exception. There were no changes related to anything like that.

PR #1385, opened 2021-03-07
[Upgrade] Upgrade llvm version from 6 to 7

sakshikukreja14:


saif-el (approved):
I will continue to use llvm v6 and trust that you are making sure that it works for v7 and above.

PR #1386, opened 2021-03-07
[Fix] Fix post-detection alignment causing feature-detected peaks to …

sakshikukreja14:
…disappear.

### DIFF
--- a/src/gui/mzroll/backgroundopsthread.cpp
+++ b/src/gui/mzroll/backgroundopsthread.cpp
@@ -331,13 +331,16 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
     auto updateGroup = [samples](PeakGroup* group) {
         MavenParameters* mp = group->parameters().get();
         auto slice = group->getSlice();
+        slice.rtmin = samples[0]->minRt;
+        slice.rtmax = samples[0]->maxRt;
 
         auto eics  = PeakDetector::pullEICs(&slice, samples, mp);
         for(auto eic : eics) {
             for(Peak& peak :  group->peaks) {
-                if (eic->getSample() == peak.getSample())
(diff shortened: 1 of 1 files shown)


src/gui/mzroll/backgroundopsthread.cpp
```diff
@@ -331,8 +331,6 @@ void BackgroundOpsThread::updateGroups(QList<shared_ptr<PeakGroup>>& groups,
     auto updateGroup = [samples](PeakGroup* group) {
         MavenParameters* mp = group->parameters().get();
         auto slice = group->getSlice();
-        slice.rtmin = samples[0]->minRt;
-        slice.rtmax = samples[0]->maxRt;
```
saif-el: Okay so that fixed it? Can I get some theories on why setting slice bounds was the issue?

This thing is important. Pulling the whole EIC makes sure that all the peaks within that EIC have been assigned quality w.r.t. other peaks in the EIC. Then when later our `group`'s peaks are scored again, they get assigned the correct quality. But sadly, this proper way of re-assigning peak quality was not happening here even if the EIC being pulled was whole. See peak-detector's [isotope-range linking logic](https://github.com/ElucidataInc/ElMaven/blob/develop/src/core/libmaven/peakdetector.cpp#L1225-L1255) for a similar use case. Let's add the relevant checks here as well.

And I still cannot think of how the slice bounds being set to min/max of the sample was the issue.

sakshikukreja14: @saiful-elucidata setting slice bounds for untargeted pulls EIC(s) (some things go wrong here I didn't get what exactly it is) and when we try to get peak details for the peak in later steps, wrong values for rt(s) are set. 

Sure, I got your point will have a look at it again and will make changes accordingly.

saif-el (approved):
Works well. Makes sense too. Rebase this and I will merge.

PR #1388, opened 2021-03-09
[FTR] Move selected Group to different Table

ghost:
Added a feature to move the selected peak groups to different peak tables and added option for the same.

### DIFF
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1893,6 +1893,7 @@ void TableDockWidget::moveSelectedRows(QString destinationTableName) {
         }
     }
 
+    if (!destinationPeakTable) return;
     for(auto group : selected) {
       destinationPeakTable->addPeakGroup(group.get());
     }
@@ -1915,6 +1916,7 @@ void TableDockWidget::moveSelectedRows(QString destinationTableName) {
         }
     }
 
+    if (!sourcePeakTable) return;
     for(auto group : selected) {
       sourcePeakTable->deleteGroup(group.get());
     }
@@ -2731,10 +2733,6 @@ void BookmarkTableDockWidget::showMsgBox(bool check, int tableNo) {
 
   msgBox->open();
 }
-
-
-
-
 
 void BookmarkTableDockWidget::mergeGroupsIntoPeakTable(QAction *action)
 {
--- a/src/gui/mzroll/tabledockwidget.cpp
+++ b/src/gui/mzroll/tabledockwidget.cpp
@@ -1868,6 +1868,59 @@ void TableDockWidget::showIntegrationSettings()
   log.exec();
 }
 
+void TableDockWidget::moveSelectedRows(QString destinationTableName) {
+
+  QString sourceTableName = this->titlePeakTable->text();
+  QList<QPointer<TableDockWidget>> peaksTableList = _mainwindow->getPeakTableList();
+  QList<shared_ptr<PeakGroup>> selected = getSelectedGroups();
+  QString bookmarkTableName = _mainwindow->bookmarkedPeaks->titlePeakTable->text();
+
+  if(bookmarkTableName == destinationTableName) {
+    BookmarkTableDockWidget* destinationPeakTable = _mainwindow->bookmarkedPeaks;
+    for(auto group : selected) {
+      destinationPeakTable->addPeakGroup(group.get());
+    }
+    destinationPeakTable->showAllGroups();
+    destinationPeakTable->setVisible(true);
+  } else {
+    TableDockWidget* destinationPeakTable = nullptr;
+    for (auto table : peaksTableList) {
+        if (table->titlePeakTable->text() == destinationTableName) {
+            destinationPeakTable = table;
+        }
+        if(destinationPeakTable != nullptr) {
+          break;
+        }
+    }
+
+    for(auto group : selected) {
+      destinationPeakTable->addPeakGroup(group.get());
+    }
+    destinationPeakTable->showAllGroups();
+  }
+
+  if(bookmarkTableName == sourceTableName) {
+    BookmarkTableDockWidget* sourcePeakTable = _mainwindow->bookmarkedPeaks;
+    for(auto group : selected) {
+      sourcePeakTable->deleteGroup(group.get());
+    }
+  } else {
+    TableDockWidget* sourcePeakTable = nullptr;
+    for (auto table : peaksTableList) {
+        if (table->titlePeakTable->text() == sourceTableName) {
+            sourcePeakTable = table;
+        }
+        if(sourcePeakTable != nullptr) {
+          break;
+        }
+    }
+
+    for(auto group : selected) {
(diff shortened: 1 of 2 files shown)


src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -2533,6 +2623,10 @@ void BookmarkTableDockWidget::showMsgBox(bool check, int tableNo) {
   msgBox->open();
 }
 
+
+
+
+
```
sakshikukreja14: @MdAzmatAli  Too many blank lines.. Could you please remove them? Ideally, we have just one blank line between the functions.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1765,59 @@ void TableDockWidget::showIntegrationSettings()
...
+        if (table->titlePeakTable->text() == destinationTableName) {
+            destinationPeakTable = table;
+        }
+        if(destinationPeakTable != nullptr) {
+          break;
+        }
+    }
+
+    for(auto group : selected) {
+      destinationPeakTable->addPeakGroup(group.get());
```
sakshikukreja14, line 1897: @MdAzmatAli, I think this might lead to a segment fault. In the case where the _destinationPeakTable_ is 'nullptr' even after for loop. 
So, for the safer side, we can add a check on line 1792 -
`if (!destinationPeakTable)
         return;
`

src/gui/mzroll/tabledockwidget.cpp
ghost: @sakshikukreja14 Fixed

src/gui/mzroll/tabledockwidget.cpp
ghost: I think it will not happen but I have added it for safer side @sakshikukreja14

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1765,61 @@ void TableDockWidget::showIntegrationSettings()
...
+    destinationPeakTable->setVisible(true);
+  } else {
+    TableDockWidget* destinationPeakTable = nullptr;
+    for (auto table : peaksTableList) {
+        if (table->titlePeakTable->text() == destinationTableName) {
+            destinationPeakTable = table;
+        }
+        if(destinationPeakTable != nullptr) {
+          break;
+        }
```
saif-el: The `if` followed by a `break` can simply be a part of the `if` above it. That is, if `destinationPeakTable` is set to a table, then we break out of the loop:
```C++
if (table->titlePeakTable->text() == destinationTableName) {
  destinationPeakTable = table;
  break;
}
```

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1765,61 @@ void TableDockWidget::showIntegrationSettings()
...
+      destinationPeakTable->addPeakGroup(group.get());
+    }
+    destinationPeakTable->showAllGroups();
+  }
+
+  if(bookmarkTableName == sourceTableName) {
+    BookmarkTableDockWidget* sourcePeakTable = _mainwindow->bookmarkedPeaks;
+    for(auto group : selected) {
+      sourcePeakTable->deleteGroup(group.get());
+    }
```
saif-el: Since the items in this table are still "selected", I think it would be better if we use the `deleteSelectedItems` method for both Bookmark table as well as other Peak tables. The method takes care of many other things in addition to deleting groups.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1765,61 @@ void TableDockWidget::showIntegrationSettings()
...
+        }
+        if(sourcePeakTable != nullptr) {
+          break;
+        }
+    }
+
+    if (!sourcePeakTable) return;
+    for(auto group : selected) {
+      sourcePeakTable->deleteGroup(group.get());
+    }
```
saif-el: Same as above.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1765,6 +1765,61 @@ void TableDockWidget::showIntegrationSettings()
...
+  } else {
+    TableDockWidget* sourcePeakTable = nullptr;
+    for (auto table : peaksTableList) {
+        if (table->titlePeakTable->text() == sourceTableName) {
+            sourcePeakTable = table;
+        }
+        if(sourcePeakTable != nullptr) {
+          break;
+        }
```
saif-el: Once more, the `break` can just go inside the first `if` and the second `if` can be safely removed.

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1788,6 +1843,43 @@ void TableDockWidget::contextMenuEvent(QContextMenuEvent *event) {
   QAction *z4 = menu.addAction("Delete all groups from this table");
   connect(z4, SIGNAL(triggered()), SLOT(deleteAll()));
 
+  QList<QPointer<TableDockWidget>> peaksTableList = _mainwindow->getPeakTableList();
+  int n = peaksTableList.count();
+
+  QMenu* subMenu = menu.addMenu(tr("Move to Table"));
```
saif-el: Let's change the label to "Move to another table…".

src/gui/mzroll/tabledockwidget.cpp
```diff
@@ -1788,6 +1843,43 @@ void TableDockWidget::contextMenuEvent(QContextMenuEvent *event) {
...
+    QAction* action = iter.key();
+    QString tableName = iter.value();
+    connect(action,
+            &QAction::triggered,
+            this,
+            [this, tableName]() { this->moveSelectedRows(tableName); });
+    if (selected.count() == 0) {
+      // disable action not relevant when nothing is selected
+      action->setEnabled(false);
+    }
```
saif-el: I know we went over this before you opened a PR, but I think it makes more sense to simply disable the parent menu item instead of disabling each child action item. So "Move to another table…" will itself be grayed when nothing is selected in that table.

saif-el (changes requested):
@MdAzmatAli Good work! I tried to make it crash but it didn't. Behaves well even for child-groups (isotopes and adducts). Feature-wise it is good to go. Suggested some code-wise improvements. Once you make those changes (and rebase this branch to latest `develop`), it can be merged.

Also, you did not have to actually name the branch `ftr_cool_feature`. I was just giving an example. The `cool_feature` part was supposed to be replaced with the actual feature you were implementing. But it's okay, this dull project can use some coolness :)

src/gui/mzroll/tabledockwidget.cpp
ghost: oh yes, will do

src/gui/mzroll/tabledockwidget.cpp
ghost: yeah right it makes more sense

saif-el (approved):
Everything works. Approving.

sakshikukreja14 (approved):

PR #1392, opened 2021-03-25
[Fix] Fix timeout while fetching user projects

sakshikukreja14:
Made the call to v2 API to use pagination while retrieving user's project names.

### DIFF
--- a/src/pollyCLI/pollyintegration.cpp
+++ b/src/pollyCLI/pollyintegration.cpp
@@ -2,6 +2,10 @@
 #include <common/downloadmanager.h>
 #include <common/logger.h>
 #include <QTemporaryFile>
+#include "core/libmaven/mzUtils.h"
+#include "json.hpp"
+
+using json = nlohmann::json;
 
 PollyIntegration::PollyIntegration(DownloadManager* dlManager):
     _dlManager(dlManager),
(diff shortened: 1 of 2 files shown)


src/pollyCLI/pollyintegration.cpp
```diff
@@ -2,6 +2,10 @@
 #include <common/downloadmanager.h>
 #include <common/logger.h>
 #include <QTemporaryFile>
+#include "core/libmaven/mzUtils.h"
+#include "json.hpp"
+
+using json = nlohmann::json;
```
saif-el, line 8: We could have used Qt's own JSON parsing library. But this is not that big of a deal.

src/pollyCLI/pollyintegration.cpp
```diff
@@ -541,21 +545,55 @@ void PollyIntegration::logout() {
...
+    QTextStream refreshTokenStream(&refreshTokenFile);
+    QString refreshToken;    
+    refreshToken = refreshTokenStream.readAll();
+    refreshTokenFile.close();
+
+    cookies += "refreshToken=";
+    cookies += refreshToken;
+    cookies += ";";
+    cookies += "idToken=";
+    cookies += idToken;
```
saif-el: I would have used `QString`'s formatting to do this in one line instead:
```C++
cookies += QString("refreshToken=%1;idToken=%2").arg(refreshToken, idToken)
```

saif-el (approved):
@sakshikukreja14 EPI perfectly now! Many thanks for fixing this.

Is there no way that older versions can be fixed by making certain changes in the index.js file? If not, we should inform the product team about this once.

PR #1394, opened 2021-03-31
Remove polly app integration

sakshikukreja14:


src/gui/mzroll/mainwindow.cpp
```diff
@@ -917,13 +917,6 @@ void MainWindow::showNotification(TableDockWidget* table) {
...
-    connect(fluxomicsPrompt,
-            &Notificator::promptClicked,
-            this,
-            [=] {
-                analytics->hitEvent("Prompt", "Clicked", "PollyPhi");
-                pollyElmavenInterfaceDialog->switchToApp(PollyApp::PollyPhi);
-            });
     connect(fluxomicsPrompt,
             SIGNAL(promptClicked(TableDockWidget*)),
             pollyElmavenInterfaceDialog,
```
saif-el: Let's remove the entire `MainWindow::showNotification` method itself. It only ever gives a notification for PollyPhi. We do not need it anymore.

saif-el (commented):
@sakshikukreja14 Thanks for cleaning this up! Some suggestions here:
* Please remove the "flux.png", "fluxSelected.png", "quant.png" and "quantSelected.png" images from "src/gui/mzroll/images" folder.
* Please also remove the videos folder entirely as well as remove the code where they were being referenced (you might have already done the latter, not sure).
* Can we get rid of the top-level tab widget in the EPI entirely? And set the "Send files to workspace" string as the main title of the dialog? See screenshot.
* Also I don't think a '-' is needed in the "Go to Polly-Workspace" button. See screenshot.
Screenshot 2021-03-31 at 12 01 37 PM

Issue #1393, opened 2021-03-30
Edits are not saved in Peak Editor for certain compounds


kehj215:
**Describe the bug**
When I'm working through a batch, certain compounds that definitely, absolutely have a peak are not integrated. When I manually select the peak in the peak editor, change the thresholding, alter the integration window, then click apply edits, the edits don't save. This only happens for certain compounds- others in my batch are totally unaffected.

**To Reproduce**
1. Bookmark a compound
2. Click peak editor
3. press "C" to create a peak
4. change threshold and RT window to encompass peak
5. click apply edits
6. Click peak editor again, see that no edits have been applied.

**Expected behavior**
Edits to my peak I created are saved after clicking "apply edits"

**Screenshots**
attached.

**System:**
 - MacOS Big Sur 11.2.3
 - I've been using the developers build for Big Sur: https://elmaven-installers.s3-us-west-2.amazonaws.com/Mac/El-MAVEN-develop.zip

**Additional context**
I am totally new to El Maven and trying to learn so it could very well be that I am the one causing the issue, however because I am able to apply edits for other peaks I'm less inclined to think so.
<img width="779" alt="Screen Shot 2021-03-30 at 11 35 38 AM" src="https://user-images.githubusercontent.com/80782500/113016574-9ea3a980-914c-11eb-8710-136fe4268f87.png">
<img width="776" alt="Screen Shot 2021-03-30 at 11 35 57 AM" src="https://user-images.githubusercontent.com/80782500/113016575-9ea3a980-914c-11eb-8ba0-0bec036c0854.png">
<img width="826" alt="Screen Shot 2021-03-30 at 11 36 20 AM" src="https://user-images.githubusercontent.com/80782500/113016577-9ea3a980-914c-11eb-8f50-8c67b8eef1ce.png">

saif-el:
@kehj215 I suspect this may not be a bug. Looking at this peak, it seems to me that it is likely being classified as a "bad" peak and being discarded even if you manually integrate it and "Apply edits".

To check whether this is the actual issue, you can turn down your quality threshold in "Peak filtering" settings tab:
![Screenshot 2021-03-31 at 11 12 16 AM](https://user-images.githubusercontent.com/77447361/113096174-5cdd3680-9212-11eb-8e1e-c7fe7f45bd3f.png)

After doing that, if you are still unable to get these to show up after editing integration, then this is something we should look into.

saif-el:
[...]

PR #1395, opened 2021-04-02
[Fix] Peak re-integration

saif-el:
Made the following fixes:
* **Peaks with `pos = 0` not being re-integrated**: each peak gets a `peak.pos` value, that tells us how high it ranks in the intensity-ordered list of all peaks from the chromatogram. The highest peak in an EIC gets the `pos` value of 0. Unfortunately, while integration this "best" peak was always getting deleted.
* **First observation in EIC always being ignored**: the rt-intensity point at the 0th position of any chromatogram was always being disregarded even if the user chose to keep it as a part of their peak of interest, while integrating through the editor.

Relevant issue: #1393


sakshikukreja14 (approved):
Works as expected :)

PR #1396, opened 2021-04-02
[UI] Show error message when MS2 samples are aligned

sakshikukreja14:
When a user tried to align MS2 sample, El-MAVEN crashed. To avoid El-MAVEN to crash, added an error message when MS2 samples are aligned.


saif-el (approved):
Good work. Merging.

PR #1402, opened 2021-04-16
[FIX] Same compounds loaded twice while opening a saved project

sakshikukreja14:
The compounds were compared on the basis of masses at different charges. Hence, these were being treated as different.
This has been fixed. The compounds will now be compared by adjusting the Mz values on the same charge


saif-el (approved):
Thanks for fixing this. Hopefully we won't hear about this issue anymore.

Issue #1415, opened 2022-07-08
misplaced string header causes failure during compilation


MetabolomicsAustralia-Bioinformatics:
When compiling from the source code within the Linux Ubuntu environment by following the guideline on the front page, the fatal error has been triggered. 

After a quick code review, the issue (and the solution) has been identified, where the root of the issue is because a string header has been misplaced. 

Harry Wang 
on behalf of Metabolomics Australia (VIC node), University of Melbourne

![Error Capture](https://user-images.githubusercontent.com/32947836/177895938-1f9b6484-c181-40cc-b425-b5bc0f31d3d5.PNG)

MetabolomicsAustralia-Bioinformatics:
The fix is really simple:
the string header within the 3rdparty/errorcodes.cpp needs to be moved into the 3rdparty/errorcodes.h

A pull request has been created to fix this issue

PR #1416, opened 2022-07-08
Fix: misplaced string header causes failure during compilation issue #1415

MetabolomicsAustralia-Bioinformatics:
This commit is a simple fix on issue #1415 

As mentioned in issue #1415, the misplaced string header causes a fatal error during compilation.

To Fix this error, the string header needs to be included in the errorcodes.h instead of errorcodes.cpp

Issue #1415


shubhra-agrawal (approved):
Thanks for the contribution!

